PR #11, opened 2018-05-18
Add a fast GetHeight that gets an approximate water height, calculate…

dizzy2003:
…s a list of most significant amplitudes, and implements a spatial cache for duplicate requests. Our Boat physics does 1000's of getHeight calls a frame, with a 90% cache hit rate with a spatial size of 10cm.  went from 230ms to 1.5 ms spent in GetHeight

### DIFF
--- a/src/unity/Assets/Crest/Scripts/OceanDepthCache.cs
+++ b/src/unity/Assets/Crest/Scripts/OceanDepthCache.cs
@@ -11,10 +11,13 @@ namespace Crest
         public bool _populateOnStartup = true;
         public LayerMask _mask;
         public int _resolution = 512;
+        public bool m_ForceAlwaysUpdateDebug = true;
 
         RenderTexture _cache;
         GameObject _drawCacheQuad;
         Camera _camDepthCache;
+
+        public float m_CameraMaxTerrainHeight = 100; //a big hill will start want to write its height into the depth texture
 
         void Start()
         {
@@ -23,7 +26,13 @@ namespace Crest
                 PopulateCache();
             }
         }
-
+#if UNITY_EDITOR
+        void Update()
+        {
+            if(m_ForceAlwaysUpdateDebug)
+                PopulateCache();
+        }
+#endif
         public void PopulateCache()
         {
             if (_cache == null)
@@ -40,6 +49,10 @@ namespace Crest
                 _drawCacheQuad = GameObject.CreatePrimitive(PrimitiveType.Quad);
                 Destroy(_drawCacheQuad.GetComponent<Collider>());
                 _drawCacheQuad.name = "Draw_" + _cache.name;
+                Vector3 pos = _drawCacheQuad.transform.position;
+                pos.y = 0; //force it to be at 0 height so the quad can be seen by the water cameras
+                _drawCacheQuad.transform.position = pos;
+
                 _drawCacheQuad.transform.SetParent(transform, false);
                 _drawCacheQuad.transform.localEulerAngles = 90f * Vector3.right;
                 _drawCacheQuad.AddComponent<RenderOceanDepth>();
@@ -52,7 +65,7 @@ namespace Crest
             if (_camDepthCache == null)
             {
                 _camDepthCache = new GameObject("DepthCacheCam").AddComponent<Camera>();
-                _camDepthCache.transform.position = transform.position + Vector3.up * 10f;
+                _camDepthCache.transform.position = transform.position + Vector3.up * m_CameraMaxTerrainHeight;
                 _camDepthCache.transform.parent = transform;
                 _camDepthCache.transform.localEulerAngles = 90f * Vector3.right;
                 _camDepthCache.orthographic = true;
--- a/src/unity/Assets/Crest/Scripts/Shapes/ShapeGerstnerBatched.cs
+++ b/src/unity/Assets/Crest/Scripts/Shapes/ShapeGerstnerBatched.cs
@@ -1,6 +1,8 @@
 ﻿// This file is subject to the MIT License as seen in the root of this folder structure (LICENSE)
 
+using System.Collections.Generic;
 using UnityEngine;
+using UnityEngine.Profiling;
 using UnityEngine.Rendering;
 
 namespace Crest
@@ -21,9 +23,13 @@ namespace Crest
 
         // data for all components
         float[] _wavelengths;
+        float[] _waveSpeed;
         float[] _amplitudes;
         float[] _angleDegs;
         float[] _phases;
+        List<int> _mostSignificant=new List<int>();
+
+        public float m_SignificantThresholdForCPU = 0.01f; //amplitude needed to use it for CPU water height calc
 
         // useful references
         WaveSpectrum _spectrum;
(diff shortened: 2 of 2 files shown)


src/unity/Assets/Crest/Scripts/Shapes/ShapeGerstnerBatched.cs
```diff
@@ -348,6 +364,92 @@ public Vector3 GetDisplacement(ref Vector3 worldPos, float toff)
...
+
+                        cos = Mathf.Sqrt(1 - Sin * Sin);
+
+                        return Sin;
+            */
+        }
+
+        Dictionary<uint,Vector3> m_Recent=new Dictionary<uint, Vector3>();
+
+        uint CalcHash(Vector3 _wp)
```
huwb: great approach with the hashing :).

the small idea occurred to me, could we do this with int instead of uint? that might allow this to reduce to:
return (int)(_wp.x + _wp.z*65000);

fairly trivial i guess but if calling this many times per frame it could make a small difference.

src/unity/Assets/Crest/Scripts/Shapes/ShapeGerstnerBatched.cs
```diff
@@ -348,6 +364,92 @@ public Vector3 GetDisplacement(ref Vector3 worldPos, float toff)
...
+        {
+            _wp *= 10;
+            return (uint)((_wp.x+32000)+(_wp.z+32000)*65000);
+        }
+        public Vector3 GetDisplacementFast(ref Vector3 worldPos, float toff)
```
huwb: good stuff. im pondering a bit if the cache could be implemented outside this class, to separate the code out and have teh cache as a layer that sits above. it may also help prevent code duplication, although i see you've broken it out because you're using the most significant wavelengths only, and you've optimised the code a bit. but this can all come later, happy with what you've implemented, with minor feedback - do you mind making whitespace consistent and remove any commented code.

src/unity/Assets/Crest/Scripts/Shapes/ShapeGerstnerBatched.cs
```diff
@@ -74,14 +81,23 @@ void Update()
...
+            {
+                _waveSpeed = new float[_wavelengths.Length];
+            }
             for (int i = 0; i < _wavelengths.Length; i++)
             {
+                _waveSpeed[i] = ComputeWaveSpeed(_wavelengths[i]);
                 _amplitudes[i] = _spectrum.GetAmplitude(_wavelengths[i]);
+                if (_amplitudes[i] > m_SignificantThresholdForCPU)
+                {
+                    _mostSignificant.Add(i);
```
huwb: good stuff. not required as part of this review, but an idea - it *might* be fruitful to actually make duplicate versions of _amplitudes, _waveSpeed, _wavelengths that have only the significant components, instead of having an array of indexes. this could be better for cache coherence. however, you might see 0 improvement for it, especially if all the data already fits in a 4k cache line which im sure it does. as i said, not required for this review but thought i'd braindump. i could also give this a stab in the future if useful. (and a random thought that just occurred to me you want to be profiling in fully optimised standalone builds if you're not already - you'll likely see quite different behaviour and bottlenecks).

src/unity/Assets/Crest/Scripts/OceanDepthCache.cs
```diff
@@ -11,19 +11,28 @@ public class OceanDepthCache : MonoBehaviour
         public bool _populateOnStartup = true;
         public LayerMask _mask;
         public int _resolution = 512;
+        public bool m_ForceAlwaysUpdateDebug = true;
 
         RenderTexture _cache;
         GameObject _drawCacheQuad;
         Camera _camDepthCache;
 
+        public float m_CameraMaxTerrainHeight = 100; //a big hill will start want to write its height into the depth texture
```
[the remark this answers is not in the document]
dizzy2003: Yeah I expected you to change this when I noticed it after submission, I added some stuff in your style but some crept through in our house style, especially as I have resharper nagging me to implement our m_ style

src/unity/Assets/Crest/Scripts/OceanDepthCache.cs
```diff
@@ -11,19 +11,28 @@ public class OceanDepthCache : MonoBehaviour
         public bool _populateOnStartup = true;
         public LayerMask _mask;
         public int _resolution = 512;
+        public bool m_ForceAlwaysUpdateDebug = true;
```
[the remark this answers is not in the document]
dizzy2003: it could be wrapped but when you make a stand alone build unity warns about structures not matching between editor and stand alone

src/unity/Assets/Crest/Scripts/Shapes/ShapeGerstnerBatched.cs
```diff
@@ -348,6 +364,92 @@ public Vector3 GetDisplacement(ref Vector3 worldPos, float toff)
             return result;
         }
 
+        public bool m_Approx = true;
+
+        public float SinCos(float x,out float cos)
```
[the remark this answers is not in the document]
dizzy2003: Yeah I actually had an inline approximation function as well as the sincos, but I couldnt get accurate enough timings through the profiler to know if they were a win, and once the spatial system cam online it made these minor optimisations probably irrelevant.

src/unity/Assets/Crest/Scripts/Shapes/ShapeGerstnerBatched.cs
dizzy2003: I did consider this, but in my use case the getheight is called high frequency one after the other so the cache should be good . Also the full arrays are small like 256 bytes small. If we really want to go all in, making the array access unsafe will remove the bounds checks that will be happening 

Yeah profiling in standalone is about 3x faster but not exceptionally different from one function to the next, the unity profiler is pretty bad, and absolutely cant be trusted in deep profile mode, then its overhead outweighs any useful data.

src/unity/Assets/Crest/Scripts/OceanDepthCache.cs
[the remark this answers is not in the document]
huwb: Hm thats pretty unfortunate about resharper style conflicts.. I had a brief google to see if i could potentially put resharper style settings in the crest folder but couldn't find much.

src/unity/Assets/Crest/Scripts/OceanDepthCache.cs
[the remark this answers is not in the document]
huwb: you're right, its better left as is

src/unity/Assets/Crest/Scripts/Shapes/ShapeGerstnerBatched.cs
huwb: ah i didnt know the unsafe array trick! thanks for the notes. happy to go with what you have.

src/unity/Assets/Crest/Scripts/Shapes/ShapeGerstnerBatched.cs
dizzy2003: thinking about it I think you need to make one axis positive. because -1,1 would be the same as 64999,0

huwb:
Closing this as I believe you moved all changes to the other pull request. If that is not correct let me know. Thanks!

PR #13, opened 2018-05-20
CPU collision from shape textures

huwb:
Wave data cameras asynchronously readback the shape render targets to the CPU, where they can then be sample to give water displacement/height information.

There are a number of things that worked out super well:

* The code/implementation is simple and fairly short (but relies on experimental APIs, as detailed below)
* I've not implemented an interpolation function and I'm not sure its required, point sampling gives good results for me
* Collision data is naturally lod'd, as are the shape textures. This gives lower res shape in the distance.
* Collision data will match almost 1:1 with rendered surface, so no disturbing artifacts where objects appear to be floating in air or submerged
* Collision data will naturally include any shape rendered - this includes wave particles, dynamic sims, and any other custom shape. This also includes height attenuation in shallow water.

Two demo scripts are provided which demonstrate sampling displacement at a position and height at a position.

This is using an experimental unity api which only supports DX11/12 currently, but Unity plan to role it out to other platforms. This api added a dependency on unity version 2018.1 so I upgraded the project as well, and no issues were encountered while upgrading.

I've not noticed any bugs so far, but I've only tested on DX11. However, there are currently a couple of limitations:

* There is a small latency between requesting the shape textures and getting them through, typically 2 or 3 frames in my experiments.  If you click Freeze waves on the debug GUI with one of the collision probe scripts active, you can observe a slight shift as the requests catch up. I hope this will not be an issue.
* Shape data is point sampled - interpolation could be added if needed
* There is no blending across shape textures, which can lead to slight pops and slight error in collision vs visual surface. Blending could be added if needed.


dizzy2003:
I implemented the new system, seems the new texture look up performance wise is on par with the GetHeightFast I implemented, so thats good. Will update by branch shortly, I made GetHeightFast check the cache then if that fails use the texture, but if the lod is -1 it falls back to the approximate function

huwb:
Thanks Dizzy! Good to hear feedback about the public API. I initially wanted to force users to call the suggest lod function and give the sample area. This means that for the case that many samples are taken over an area, the lod index only has to be computed once, and a large sampling area will filter out small wavelengths that would be susceptible to aliasing and would not make a big difference to a calculation such as buoyancy.

However this is clearly confusing and I've added functions that simply take a point and return the height/displacement. These are less likely to be efficient this way, but probably much easier to understand. Does this make sense? If you have feedback let me know in the next day or two, otherwise ill get it merged in!

PR #14, opened 2018-05-22
Integrate the cache and the getHeightFast and the texture lookup for …

dizzy2003:
You probably cant take this or at least the shader changes to BIAS and SEA_LEVEL
we use sea level of 10 and BIAS of 0  (for now, until we sort out what we are doing with sea depth) (Our terrain is at 0 y and sea is at 10 Meters.

I have made a generic GetHeightFast that uses the cache, if that fails it uses the texture lookup if that fails it uses the cpu approximation.

I also did the clean ups you asked for, at least most of them anyway..

### DIFF
--- a/src/unity/Assets/Crest/Scripts/OceanDepthCache.cs
+++ b/src/unity/Assets/Crest/Scripts/OceanDepthCache.cs
@@ -11,10 +11,11 @@ namespace Crest
         public bool _populateOnStartup = true;
         public LayerMask _mask;
         public int _resolution = 512;
-
+        public bool _forceAlwaysUpdateDebug = true;
         RenderTexture _cache;
         GameObject _drawCacheQuad;
         Camera _camDepthCache;
+        public float _cameraMaxTerrainHeight = 100; //a big hill will still want to write its height into the depth texture
 
         void Start()
         {
@@ -23,6 +24,14 @@ namespace Crest
                 PopulateCache();
             }
         }
+
+#if UNITY_EDITOR
+        void Update()
+        {
+            if (_forceAlwaysUpdateDebug)
+                PopulateCache();
+        }
+#endif
 
         public void PopulateCache()
         {
@@ -40,6 +49,9 @@ namespace Crest
                 _drawCacheQuad = GameObject.CreatePrimitive(PrimitiveType.Quad);
                 Destroy(_drawCacheQuad.GetComponent<Collider>());
                 _drawCacheQuad.name = "Draw_" + _cache.name;
+                Vector3 pos = _drawCacheQuad.transform.position;
+                pos.y = 0; //force it to be at 0 height so the quad can be seen by the water cameras
+                _drawCacheQuad.transform.position = pos;
                 _drawCacheQuad.transform.SetParent(transform, false);
                 _drawCacheQuad.transform.localEulerAngles = 90f * Vector3.right;
                 _drawCacheQuad.AddComponent<RenderOceanDepth>();
@@ -52,6 +64,7 @@ namespace Crest
             if (_camDepthCache == null)
             {
                 _camDepthCache = new GameObject("DepthCacheCam").AddComponent<Camera>();
+                _camDepthCache.transform.position = transform.position + Vector3.up * _cameraMaxTerrainHeight;
                 _camDepthCache.transform.position = transform.position + Vector3.up * 10f;
                 _camDepthCache.transform.parent = transform;
                 _camDepthCache.transform.localEulerAngles = 90f * Vector3.right;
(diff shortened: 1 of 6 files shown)


huwb:
Thanks! Yeah I'd probably cherry pick some of these changes.

I think there is overlap between this cached sampling work and the texture lookup. It sounds like from initial tests you're finding the perf of the texture lookup stuff is decent. Given this and current lack of ocean depth calculations in the gerstner code I think i'll get the texture lookup stuff merged in first and we can evaluate the need for the cache going forwards? I'm still getting my head straight about the pros/cons of each, i think it would be good to run with the texture lookup stuff first and get clarity on where that is falling short. The texture lookup stuff has some limitations but I'd like to know if these are issues in practice. Let me know if I'm missing important angles here - I'm probably a bit too tired to be doing this right now so I might be overlooking something obvious!

dizzy2003:
With out the cache my 1500 samples that run in the boat buoyancy physics (a
unity asset store package) update took 9 ms on a high end PC, so probably
double that on a console, quadruple on a switch as a very rough estimate..

1.5 ms with the cache

I will need the cache as an easy win at least for now. Allthough it has its
own issues, dictionaries causing mem allocs..

Having said that I havent looked into the new texture system, without
calcing the lod each time it might be much quicker.

And I can keep just merging my local changes in.

dizzy2003:
One thing I forgot to mention is I noticed the Profiler.Begin() and end()
where actually really slow when called a lot, even without the profiler
running so I deleted them from my branch, from the texture sampling code too

src/unity/Assets/Crest/Scripts/OceanDepthCache.cs
```diff
@@ -40,6 +49,9 @@ public void PopulateCache()
                 _drawCacheQuad = GameObject.CreatePrimitive(PrimitiveType.Quad);
                 Destroy(_drawCacheQuad.GetComponent<Collider>());
                 _drawCacheQuad.name = "Draw_" + _cache.name;
+                Vector3 pos = _drawCacheQuad.transform.position;
+                pos.y = 0; //force it to be at 0 height so the quad can be seen by the water cameras
+                _drawCacheQuad.transform.position = pos;
```
huwb: I'm a bit unsure about this one - I give false as a second argument to set transform on the line below which i thought was actually not taking the world position. can you confirm that this code is merged up correctly and fixes an issue?

huwb:
I merged the async texture readback version into mainline, and then i merged this into the coll_from_disp_texs branch afterwards. i want to process your notes above and my thoughts and perhaps do some perf tests and settle on something to take to mainline.

i hope this has not made a mess of things. i dont think it has, but let me know.

currently the result in editor looks.. foamy.. i think its because of the BIAS but i need to dig deeper. i've run out of time now but will check it tomorrow.

dizzy2003:
Yeah the bias and default sea height(0) need to be set in a component
somewhere, then passed through to the shaders, the defaults wont
necessarily work for everyone..

huwb:
Yeah the sea level was not supposed to be hardcoded, good point. It is taken from the Ocean prefab y value and is available through OceanRenderer.SeaLevel. I've not un-hardcoded it in 2c4026a2cbc4dbb0ff92fff8f561eb8838cc1833 , I tried at a couple of different sea levels and it seems to work with both cached and dynamic ocean depths.

The bias is an offset relative to the sea level. It moves the depth values away from 0 which means the buffer can be cleared to 0 which saves doing a init pass somewhere else in the code. It should never have to change, or at least it should not be dependent on sea level (there could be another issue where it interacts badly with terrain heights or something, in which case we can increase it). For now I've made it 100 in the coll_from_disp_texs branch, if this is causing badness let me know.

huwb:
Ok i've done a cleanup of this branch:

I merged some bits to master, like removing the profiling samples and adding the debug option to render ocean depth every frame.
I cleaned out some stuff that I believe should not be in the branch such as some ocean material tweaks and the depth bias - see my comment above.

There's now two main things left in the branch. One is setting the depth cache draw quad y coord to 0, i wasn't sure if this was intentional or had an effect, as commented above.

The other is the rest of the changes to cache the gerstner lookups which currently has two issues. firstly it doesnt take ocean depth into account. secondly it doesnt take chop into account. the chop issue is minor but the depth one is bigger. given that the tex readback stuff accounts for these things, and also supports all shape types, and should be efficient, I'd recommend moving over to that API. adding a caching layer around that API might also help perf (which you have support for in this branch).

in any case I'll let you consider that and leave this branch as is with the current changes. i hope my clean up stuff does not make your life too painful when reconciling with your fork.

dizzy2003:
We had some issues with shore depth not working so I poked a lot of stuff,
some of it was due to the height of the camera that renders the depth cache
being too low, some of it was the shader sea level not being compatible
with ours and I think I also had an issue with the quad being positioned
above the sea lod camera. Its possible not all of these were real issues as
I was debugging multiple issues in multiple areas to get the shore working.
(The depth bias I'm not sure about, our terrain is at 0 and sea level is at
10, I was concerned about getting a step where our terrain stops, the
terrain only extends 20 or so meters under the sea (we have individual
unity terrains dotted around a large multi km map)..  so at the moment we
only have a max sea depth of 10 meters if we pretend the terrain extends
out in all directions at 0 altitude)

I will just return to your head master branch and let you know if theres
any issues.

PR #23, opened 2018-06-05
Collision providers, and collision cache

huwb:
I've been thinking about a way to get @dizzy2003 's nice caching work (#14) integrated into master, while supporting different collision types.

Here i've experimented with creating an API for a collision provider (disp texs, gerstner waves, etc) ICollProvider which the other systems use. The Gerstner wave component implements this interface.

On top of that I added a cache that wraps around this: CollProviderCache.

The usage for accelerating Gerstner wave queries would be something like:

var _myCache = new CollProviderCache( myGerstnerWavesComponent );
...
_myCache.SampleHeight( ... );

This paves the way for giving options to the user for where cpu coll samples should come from and if a cache should be used.

@dizzy2003 if you have thoughts for/against moving in this direction, let me know!


huwb:
I just added a bit more code - there is an option on ocean render to use the cache (has to be set before startup, not at runtime).

Did a quick test in editor for 1000 height queries at a position the cache dropped the time from 4.3ms to 0.46ms . This was in dev not release but still thats really promising.

I added cache hits/misses to the debug gui.

@dizzy2003 - there is one change that might affect existing ocean uses - the OceanRenderer update always needed to happen first. I've now put it before the default update time, because it is now responsible for clearing the height cache. If this is not called first up, any updates that query the height will get stale data. If you see any glitchiness from the update order change let me know.

PR #29, opened 2018-06-25
Add option to display checkerboard in ocean surface alpha shader

moosichu:
Before:
image
After:
image

### DIFF
--- a/src/unity/Assets/Crest-Examples/Shaders/OceanSurfaceAlpha.shader
+++ b/src/unity/Assets/Crest-Examples/Shaders/OceanSurfaceAlpha.shader
@@ -4,6 +4,7 @@
 	{
 		_MainTex ("Texture", 2D) = "white" {}
 		_Alpha("Alpha Multiplier", Range(0.0, 1.0)) = 1.0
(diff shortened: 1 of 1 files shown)


huwb:
Thanks for posting this. I'm a bit on the fence about adding code especially for this. This shader aims to give a simple(ish!) example of how alpha can be drawn on the ocean surface, and adding procedural checkerboard adds a lot of code relative to the original. However i havent actually made use of the checkerboard pattern in this context. Did you find it useful enough to be worth the added code size/complexity? If so maybe extracting it out into a function helps maintain some of the readability/simplicity.

Another idea might be to copy the shader and make a checkerboard version. Or there is always the option of adding a checkerboard texture to give the pattern and not adding any code at all :). I usually prefer the no-code option, as nonexistent code is the easiest kind of code to maintain and ship :).

Sorry if i'm overthinking this or making something simple complicated! Let me know what you think.

huwb:
I guess I'll close this for now with the above comments, but do reopen or chat to me directly if you think we should merge it in. Cheers!

PR #32, opened 2018-07-31
Prepared a simple test scene with planar reflections

Rbn3D:
I've prepared a sample scene showing what I have done for planar reflections so far. The boat of the boat.unity scene it's probably too small to showcase the squares/seams bug, but it's more noticeable if you place a bigger object on the water.

In summary:

- There are seams/squares, mostly noticeable with big objects (somewhat related to geometry/normals)
- There are a 'low update rate' of planar reflections over the water (looks like the reflected texture is not updated every frame)
- There is also a Clip Plane Offset problem, but I think it's normal in this case (the ocean surface it's not really planar)

Scene is at CrestExamples/BoatDev/Scenes/boat-planarReflections.unity

previous discussion at https://github.com/huwb/crest-oceanrender/issues/31


Rbn3D:
Oh, forgot to mention that my local Unity version is 2018.2.1f1, so the project got updated on this branch.

Sorry 😢

Rbn3D:
Just updated the branch with latest stuff form master, also merged Ocean.shader to OceanPlanarReflections.shader, so the planar reflection shader also benefits from latest stuff.

PR #39, opened 2018-08-06
This is a wip PR for factoring out the foam texture from the displacement texture

moosichu:


huwb:
looks good to me, at least from first glance. i sent a query about the calculate UVs function you added, i think there is already a function for this.

PR #48, opened 2018-08-19
Unify all LOD data into a single coherent structure

huwb:
BIG refactor together with @moosichu.

System has been creaking since sims were bolted onto the side. This change unifies the 'sims' and the existing wave/depth rendering, among other changes:

* Animated (gerstner) waves, ocean depth, dynamic waves, foam, all unified into a single LODData system
* Sims now float with LODs just like the animated (gerstner) waves - resolutions dont need to be set manually
* Terrible bug fixed where foam was being multiplied more and more the closer to the camera (foam was being accumulated down the lod chain erroneously). Besides being really confusing and hard to tweak, the bug also caused a subtle pop in the foam appearance.
* Simplified system for advecting sim results across frames, made robust. Validation of update order in a few places.
* Rebuilt shader params and samplers - new Bind() functions more intuitive and explicit
* No longer manually clearing first/last lods on scale changes - just give blackTexture as input!
* Simplified foam tweak vars - only a single strength var on the foam sim, and a feather var on the ocean mat. If we can get by with just these then it will be much simpler.
* Added optional min/max grid size setting for dynamic wave sims to specify range of resolutions that this sim should run at
* Water interaction forces are aware of the number of wave sims so amplitude should stay roughly constant independent of sim count
* Debug view shows different buffer types (thanks @moosichu)

Impact:

* A lot of type names have changed, but a lot of the scripts were assigned automatically at startup, so it should not affect existing data too much
* Layer names have changed -  this will make existing data go stale. Fixup is required: SimFoam -> LodDataFoam, SimShape -> LodDataAnimatedWaves, SimWave -> LodDataDynamicWaves.
* Foam vars are simplified (both on sim and ocean mat), will likely need retweaking. Bubble coverage tweak var inverted to give expected behaviour.
* By default 7 detail levels of dynamic wave sims will run. If performance is an issue this can be limited in the dynamic wave sim settings using the grid size options.


huwb:
@moosichu !

PR #57, opened 2018-08-28
Create a flow sim system for advecting the water surface

moosichu:
This commit adds a new sim which can be used to store and manipulate flow on the Ocean's surface.

Currently, only normals can be advected. An example of this has been added, featuring a whirlpool-like structure.

This feature is currently experimental.

image

Special thanks to @huwb for all the guiders, pointers and help!


huwb:
Super excited about this! 

Some requests - please let me know if i misunderstand something

* Could you please name New Material.mat if its being used, or remove it otherwise
* Did you intend to check in the change to threeboats.unity? If the flow is being used there we can check it in, otherwise i'd be inclined to leave it off
* Do you have a plan for WhirlPoolGenerator? If so then ignore this and lets check it in, otherwise it could be a candidate for removal and/or bringing back later.
* I see what you mean about variable names. Could w1 be sample_offset1 or something, and ww1 be sample_weight1, perhaps?

Thoughts (dont necessarily need actioning for this PR merge):

* It just occurred to me we should think about if we want persistent flow state across frames (such as something fading or simulating over time). Its currently stateless and if it stays like that we could move it inherit from LodData instead of LodDataPersistent. This would be slightly simpler and would not create the pingpong render targets etc.
* Related to the above - if we're not actually 'sim-ing' anything substantial we could remove ShapeSimFlow and set the override CamClearFlags in LodDataFlow to clear to solid color and it will be cheaper than running this shader. But we can defer this decision or if you have plans along these lines, ignore this point!

Thats it - nice work with this!

moosichu:
Yeah, currently I can't think of any reason why flow should be stateful. I was thinking about that earlier as well. However, it might be worth deferring changing the system until later, because we may think of something as we use the system.

However, I do think that a stateless flow system should at least be an option down the road, even if we find uses for a stateful one..

moosichu:
Feedback should all be implemented now.

Issue #65, opened 2018-09-04
UpdateCmdBufOceanFloorDepth FindObjectsOfType is very very slow


dizzy2003:
If you ever want to update the command buffer at runtime (we do) this FindObjectsOfType is a huge bottleneck 42 ms.

Maybe instead RenderOceanDepth objects should register themselves with the LodDataSeaFloorDepth object.

Issue #66, opened 2018-09-04
UpdateCmdBufOceanFloorDepth checking objects for the existance of OceanDepthCache is a bit clunky


dizzy2003:
Having rolled my own type of ocean depth cache with my own specific needs I found it didnt work for a while because LodDataSeaFloor is replacing the material with its own global one.

Having to put an OceanDepthCache on to stop that behaviour isnt really a good solution. (Locally I just deleted the check as I never want to use the global depth cache material)

If you go over to a register option as mentioned in the other issue, it could also flag if the renderer should use its own material.

PR #67, opened 2018-09-05
Refactor ocean depth registration

huwb:
Hi @dizzy2003 ,

Would these changes resolve issues #65 and #66 ?

Thanks for the suggestion!


huwb:
I decided to add a material override field. I tested in editor and standalone and it seems fine so i'll merge it in.

huwb:
Hi @dizzy2003, in #117 i'm looking at moving all the lod data rendering over to registration instead of using cameras and layers. will be great if it works out!

The material override field i added for specifying the depth render material has become awkward, and i'd like to remove it from the RenderOceanDepth script. It wasn't clear in the above, are you guys using this field?

dizzy2003:
Thats fine, do it, I Look forward to the change to all command buffers,
should be a big win over the large number of cameras we have now.

PR #73, opened 2018-09-18
Change uses of ref to other modifiers in functions

moosichu:
The reason for this change is that ref implies that an input value
can both be read from and written to. However, for some parameters
they were only being read from (so were changed back to pass-by-value,
although the 'in' parameter modifier would be ideal in future versions
of C#).

Furthermore, some parameters were only being written to, so were changed
to `out` parameters.

For further reading see: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/method-parameters

### DIFF
--- a/src/unity/Assets/Crest/Scripts/Collision/CollProviderDispTexs.cs
+++ b/src/unity/Assets/Crest/Scripts/Collision/CollProviderDispTexs.cs
@@ -17,7 +17,7 @@ namespace Crest
         {
             int lod = LodDataAnimatedWaves.SuggestDataLOD(new Rect(in__worldPos.x, in__worldPos.z, 0f, 0f), 0f);
             if (lod == -1) {
-                displacement = Vector2.zero;
+                displacement = Vector3.zero;
                 return false;
             }
             return OceanRenderer.Instance._lodDataAnimWaves[lod].SampleDisplacement(ref in__worldPos, out displacement);
@@ -34,7 +34,7 @@ namespace Crest
         {
             int lod = LodDataAnimatedWaves.SuggestDataLOD(new Rect(in__worldPos.x, in__worldPos.z, 0f, 0f), 0f);
             if (lod == -1) {
-                height = float.NaN;
+                height = 0;
                 return false;
             }
             height = OceanRenderer.Instance._lodDataAnimWaves[lod].GetHeight(ref in__worldPos);
--- a/src/unity/Assets/Crest/Scripts/Collision/CollProviderDispTexs.cs
+++ b/src/unity/Assets/Crest/Scripts/Collision/CollProviderDispTexs.cs
@@ -13,31 +13,31 @@ namespace Crest
     {
         int _areaLod = -1;
 
-        public bool SampleDisplacement(Vector3 worldPos, out Vector3 displacement)
+        public bool SampleDisplacement(ref Vector3 in__worldPos, out Vector3 displacement)
         {
-            int lod = LodDataAnimatedWaves.SuggestDataLOD(new Rect(worldPos.x, worldPos.z, 0f, 0f), 0f);
+            int lod = LodDataAnimatedWaves.SuggestDataLOD(new Rect(in__worldPos.x, in__worldPos.z, 0f, 0f), 0f);
             if (lod == -1) {
                 displacement = Vector2.zero;
                 return false;
             }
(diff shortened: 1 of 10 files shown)


huwb:
Hey, thanks for this!

I'm only unsure about changing to pass by values. This is potentially very hot code and my (possibly wrong) impression is that pass by value is a no-no for perf critical code?

HyroVitalyProtago:
@huwb As I know, you're right. Even if like @moosichu , I dislike using permissive things like _ref_, I think it's better in terms of performance because _struct_ are copied when passed by value and as @moosichu said, the _in_ keyword is not there now...

moosichu:
Yeah, I'm going to modify this code to use ref again, but prepend the 'in' variables with `in__` so that they can be changed at a later date.

moosichu:
@huwb The refs are back! Let me know what you think.

src/unity/Assets/Crest/Scripts/Collision/CollProviderDispTexs.cs
```diff
@@ -13,25 +13,31 @@ public class CollProviderDispTexs : ICollProvider
...
 
-        public bool SampleDisplacement(ref Vector3 worldPos, ref Vector3 displacement)
+        public bool SampleDisplacement(ref Vector3 in__worldPos, out Vector3 displacement)
         {
-            int lod = LodDataAnimatedWaves.SuggestDataLOD(new Rect(worldPos.x, worldPos.z, 0f, 0f), 0f);
-            if (lod == -1) return false;
-            return OceanRenderer.Instance._lodDataAnimWaves[lod].SampleDisplacement(ref worldPos, ref displacement);
+            int lod = LodDataAnimatedWaves.SuggestDataLOD(new Rect(in__worldPos.x, in__worldPos.z, 0f, 0f), 0f);
+            if (lod == -1) {
+                displacement = Vector2.zero;
```
huwb: Should this be Vector3.zero instead of Vector2.zero

src/unity/Assets/Crest/Scripts/Collision/CollProviderDispTexs.cs
```diff
@@ -13,25 +13,31 @@ public class CollProviderDispTexs : ICollProvider
...
 
-        public bool SampleHeight(ref Vector3 worldPos, ref float height)
+        public bool SampleHeight(ref Vector3 in__worldPos, out float height)
         {
-            int lod = LodDataAnimatedWaves.SuggestDataLOD(new Rect(worldPos.x, worldPos.z, 0f, 0f), 0f);
-            if (lod == -1) return false;
-            height = OceanRenderer.Instance._lodDataAnimWaves[lod].GetHeight(ref worldPos);
+            int lod = LodDataAnimatedWaves.SuggestDataLOD(new Rect(in__worldPos.x, in__worldPos.z, 0f, 0f), 0f);
+            if (lod == -1) {
+                height = float.NaN;
```
huwb: NaN's terrify me - maybe an irrational fear but in previous projects I've seen them spread and cause badness around a system, once we had crashes on some PCs due to NaNs infecting the graphics driver. I'd probably feel more comfortable going with 0 here, or do you feel strongly it should be NaN? Zero would also match the Vector3.zero.

src/unity/Assets/Crest/Scripts/Collision/CollProviderDispTexs.cs
huwb: (theres a second Vector2 typo in one other place as well)

src/unity/Assets/Crest/Scripts/Collision/CollProviderDispTexs.cs
moosichu: Yeah, I'm happy changing that.

I would be nice if there was a system where you could turn on NaN's in debug builds only, but also with a feature that tagged every NaN so you could see where they spread from.

But having less potential badness in a system makes sense.

src/unity/Assets/Crest/Scripts/Collision/CollProviderDispTexs.cs
moosichu: You are correct, it's annoying that unity autocasts like that

moosichu:
Thank you for catching those errors, everything should all be good now.

huwb:
looks good nice one - feel free to do the honours!

PR #76, opened 2018-09-23
Fix/ocean depth baseline

huwb:
FYI @dizzy2003 - i'm aware you guys have your own ocean depth shaders? I've changed the meaning of this ocean depth to be 'height above -1000m' instead of what it used to be which was something like 'water depth plus 100'. some more information from the commit:

We need '0' to mean deep water. This is required so that when lods are faded in/out, it can fade to 0 (deep). It also means that outside the biggest lod the value can be 0 and not appear to be shallow. This fix changes the meaning of the depths - it is now the height of the sea floor above some deep value (currently -1000m).

the reason for this change: i discovered that the outer skirt of geometry was being shaded as shallow water. so there was a bright blue region near the horizon. it wasn't noticeable for some views but would become worse with less LODs or for high viewpoints, so im changing the meaning of depth so it works in all scenarios.

if you have any thoughts etc let me know

### DIFF
--- a/src/unity/Assets/Crest-Examples/Scripts/OceanDebugGUI.cs
+++ b/src/unity/Assets/Crest-Examples/Scripts/OceanDebugGUI.cs
@@ -17,6 +17,12 @@ public class OceanDebugGUI : MonoBehaviour
 
     private void Start()
     {
+        if (OceanRenderer.Instance == null)
+        {
+            enabled = false;
+            return;
+        }
+
         gerstners = FindObjectsOfType<ShapeGerstnerBatched>();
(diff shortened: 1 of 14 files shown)


dizzy2003:
Looks fine to me, we ended up using your depth offset anyway.

we do have a heavily modified ocean.shader because we have about 12 ocean
params that we change per pixel (we store this info in textures a bit like
the way you cache the ocean depth, except its at a constant very low res,
rather than a different res per lod)

huwb:
Interesting. Thanks for the note. If there's anything we could simulate in loddata's instead of baking out to textures, let me know, but i can imagine these are artist authored and not replaceable with anything automatic.

Merging this PR.

PR #80, opened 2018-09-26
Make big improvements to flow system

moosichu:
This PR makes the following changes:
 - Add the ability to apply flow to foam (both the sim and how the texture is sampled)
 - Have flow affect the physics of the boat in the water
 - Have flow affect the wave sim


moosichu:
The shape sims which have been modified to use flow data still need to be cleaned-up so that that their ability to sample flow can be enabled/disabled. Which I will do. I am making this pull-request so I can more easily keep track on my progress with that and get feedback on the rest.

moosichu:
Video:

https://youtu.be/a8_nUuVSw84

src/unity/Assets/Crest-Examples/BoatDev/Scripts/BoatAlignNormal.cs
```diff
@@ -59,6 +59,15 @@ void FixedUpdate()
 
         // estimate water velocity
         Vector3 velWater = (_displacementToBoat - _displacementToBoatLastFrame) / Time.deltaTime;
+        if(OceanRenderer.Instance._createFlowSim) {
+            Vector2 surfaceFlow;
+            int lod  = LodDataFlow.SuggestDataLOD(new Rect(position.x, position.z, 0f, 0f), _boatWidth);
```
huwb: we could pull out the lod up above and use it for sampling the other data, but its not required for this PR

src/unity/Assets/Crest/Materials/Ocean.mat
```diff
@@ -74,7 +74,7 @@ Material:
         m_Scale: {x: 1, y: 1}
         m_Offset: {x: 0, y: 0}
     m_Floats:
-    - _ApplyFlowToNormals: 0
```
huwb: should be off by default on the main ocean mat

src/unity/Assets/Crest/Scripts/LodData/LodData.cs
```diff
@@ -143,6 +143,7 @@ public static GameObject CreateLodData(int lodIdx, int lodCount, GameObject atta
                     break;
                 case SimType.Flow:
                     sim = go.AddComponent<LodDataFlow>();
+                    go.AddComponent<ReadbackLodData>();
```
huwb: really happy this worked out :)

src/unity/Assets/Crest/Scripts/LodData/LodDataFlow.cs
```diff
@@ -1,38 +1,82 @@
...
 using UnityEngine;
+using UnityEngine.Rendering;
 
 namespace Crest
 {
     /// <summary>
     /// A persistent flow simulation that moves around with a displacement LOD. The input is fully combined water surface shape.
     /// </summary>
-    public class LodDataFlow : LodDataPersistent
+    public class LodDataFlow : LodData
```
huwb: ahh you moved it to loddata! that seems like the right way to go for now, good stuff

src/unity/Assets/Crest/Materials/Ocean.mat
moosichu: isnt't that turned off?

src/unity/Assets/Crest/Scripts/LodData/LodData.cs
moosichu: it was really nice :)

src/unity/Assets/Crest/Scripts/LodData/LodDataFlow.cs
```diff
@@ -1,38 +1,82 @@
...
+        public override RenderTexture DataTexture { get { return Cam.targetTexture; } }
+        public override void UseSettings(SimSettingsBase settings) {}
+        public override SimSettingsBase CreateDefaultSettings() { return null;}
 
-        public override SimSettingsBase CreateDefaultSettings()
+        protected override void Start() {
+            base.Start();
+
+            _dataReadback = GetComponent<ReadbackLodData>();
+            _dataReadback._textureFormat = ReadbackLodData.TexFormat.RGHalf;
```
huwb: probably worth driving this from TextureFormat property above if you agree so we're not duping it

moosichu: ok, will do that

src/unity/Assets/Crest/Scripts/LodData/LodDataFlow.cs
```diff
@@ -1,38 +1,82 @@
...
+
+        void OnDisable() {
+            // free native array when component removed or destroyed
+            if (_dataReadback._dataNative.IsCreated)
+            {
+                _dataReadback._dataNative.Dispose();
+            }
+        }
+
+        public static int SuggestDataLOD(Rect sampleAreaXZ, float minSpatialLength)
```
huwb: hm wonder if duping this could be avoided. dont think its required for this PR though.

src/unity/Assets/Crest/Scripts/LodData/ReadbackLodData.cs
```diff
@@ -50,7 +50,6 @@ private void Update()
         public void UpdateReadback(Camera cam, LodTransform.RenderData renderData)
         {
             // queue pattern inspired by: https://github.com/keijiro/AsyncCaptureTest
-
```
huwb: perhaps revert this whitespace if you dont mind, the two lines were supposed to be separated, even if it only makes sense to me!!

src/unity/Assets/Crest/Scripts/LodData/LodDataFlow.cs
moosichu: Hmm, I'm trying to work out the best way to do this.
As ReadBackLodData has it's own enum
Should I change that to RenderTextureFormat?

src/unity/Assets/Crest/Scripts/LodData/ReadbackLodData.cs
moosichu: Will do! Sorry, I  didn't even realise I did that.

src/unity/Assets/Crest/Shaders/Simulation/Resources/ShapeSimFoam.shader
```diff
@@ -60,9 +60,14 @@ Shader "Ocean/Shape/Sim/Foam"
...
 					float4 uv = float4(i.uv_uv_lastframe.xy, 0., 0.);
 					float4 uv_lastframe = float4(i.uv_uv_lastframe.zw, 0., 0.);
+					half4 velocity = half4(tex2Dlod(_LD_Sampler_Flow_1, uv).xy, 0., 0.);
+					//velocity = half4(10, 0, 0, 0);
 
 					// sampler will clamp the uv currently
-					half foam = tex2Dlod(_LD_Sampler_Foam_0, uv_lastframe).x;
+					half foam = tex2Dlod(_LD_Sampler_Foam_0, uv_lastframe
+						- ((_SimDeltaTime * i.invRes) * velocity)
+					).x; // TODO: use world coords here.
```
huwb: just to check - is the TODO valid here

src/unity/ProjectSettings/ProjectVersion.txt
```diff
@@ -1 +1 @@
-m_EditorVersion: 2018.2.5f1
```
huwb: This is a funny one - i googled this the other day and apparently it is appropriate to check it in, but its weird if this gets wound backwards I think because one of us is on a slightly different version! I think i'd vote for treating it like a high water mark, otherwise we will end up playing ProjectVersion tennis :). What do you think?

src/unity/Assets/Crest/Scripts/LodData/LodDataFlow.cs
moosichu: The version in LodDataFlow is slightly different from the existing one. Just as a note.

huwb:
Super awesome!!

src/unity/Assets/Crest/Scripts/LodData/LodDataFlow.cs
moosichu: Cool beans

PR #89, opened 2018-10-07
Refactor/drive shader keywords

huwb:
@Midda-C @dizzy2003 @holdingjason

Up until now there were checkboxes for features driven by code such as foam, shadowing, planar reflections, and possibly more things in the future. This has a number of disadvantages:

* Duplicate config - if you forget to turn it on on the material then the feature won't appear and it won't give any warnings or errors
* Duplicate config - you may disable a feature on the ocean renderer but accidentally leave it turned on on the ocean mat, and pay for a feature you're not using
* If you want a feature on in some situations and off in others, the current workflow suggests that you need to duplicate the ocean material, which can create a lot of duplicated data which creates maintenance issues.

In this PR the above mentioned features are driven directly by code. When the LodData activates it will turn on the feature on the ocean material, and when it deactivates it will turn it off. The options for these are no longer exposed in the inspector, and switched off by default.

The switches that are now removed used to serve as group headers in the parameter list, and I faked an indentation using spaces for params in each group. Now that some of these are gone, I found the Header attribute which gives proper headings:

image

Impact - Should be small. If you have any ocean mat variants that purely turn on/off the above features, this may mean you can remove the dupes. The param names should all be the same so there should be no required data to move forward.

If you guys have any thoughts etc let me know, otherwise I'll get it merged.


moosichu:
Awesome, looking at it now :)

Issue #77, opened 2018-09-24
Mask/Cutout Ocean


holdingjason:
Not sure if this has come up but wondering if we had a proper way to not render/cut out an area of the ocean.  Basically we need a way to be able to go below the ocean plane around certain areas of the terrain, caves whatever.

I guess one way would be to apply something similar to how we flatten the ocean out near shores but actually push it down further on the y.

gamemachine:
While I'm assuming a depth mask would work for this, I would propose a bounds based solution that clips geometry based on worldspace bounds passed to the shader.   I have some existing code that goes this I could share, I use it for dynamic entrances into terrains.

The issue with depth masks is aligning them to geometry.  You have to do it exact or you get artifacts of various types, and it's impossible to do well really.

gamemachine:
Gist with the core functionality.  Encode bounds min/max in a float4 you pass to the shader, and call this function with that float4 and world position.  You can pass a vector array also and then loop over several.  It's not really that expensive, especially not in comparison to the overall cost of the water.

https://gist.github.com/gamemachine/d97869b20046c2f3b515679f4c7b266f

holdingjason:
That makes a lot of sense actually.  Need to also try this out with terrain as well.  Right now we do a depth mask on the terrain for holes etc which works fine and we were planning on doing that for entrances ie caves etc as well but this approach might be better.

gamemachine:
Ya I went this route after struggling with dynamic terrain holes.  Being that you can't get at the terrain geometry it's impossible to make depth masks that fit exactly, leaving gaps that are visible at certain angles.

huwb:
Thanks guys. The hole shape could also come from a world aligned texture rather than a procedural square.

Another small addition might be to do the clip in the vertex shader - this is possible by returning NaN in the vertex for the POS. Apparently its undocumented but GPUs support it - they'll discard the vert.

Jeremy Moore talks about this here for punching holes in terrain:
https://www.gdcvault.com/play/1025480/Terrain-Rendering-in-Far-Cry

( This page seems to have died for me but some guy has mirrored it here:
http://www.geeks3d.com/forums/index.php?topic=5381.0 )

This saves the GPU having to do rasterisation and running the pixel shader. However with the dynamically changing mesh density, the bounds of the hole may change slightly, which may be visible depending on the scenario. Just thought I'd throw it in the mix.

huwb:
[...]

PR #95, opened 2018-10-09
#77 - Added a shader/mat that pushes water surface to top of geometry

huwb:
The changes to main.unity should be reverted before merging this, but creating this now for discussion.

### DIFF
--- a/src/unity/Assets/Crest-Examples/Scenes/main.unity
+++ b/src/unity/Assets/Crest-Examples/Scenes/main.unity
@@ -114,6 +114,93 @@ NavMeshSettings:
     debug:
       m_Flags: 0
   m_NavMeshData: {fileID: 0}
+--- !u!1 &82883802
+GameObject:
+  m_ObjectHideFlags: 0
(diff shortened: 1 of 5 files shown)


moosichu:
Will check this out this weekend, looks interesting!

huwb:
Closing - this PR is now pushed separately in 29f404992177cbd9226bf21622b77d9965badb44 .

@GeorgeDziov , thanks for the comment.

It took me a very long time to get around to it but i finally made changes that fix this issue.

Firstly its worth checking my latest post on #77 for the updates to how this kind of shape input works now.

Now I've made it so the displacement from a shape input can be reported through to the ocean system, which means the ocean tile bounding boxes will be enlarged to capture the displacement and the tiles should not get culled (which is what you see in the video above).

Long story short, if you open boat.unity on master then you'll see the Register Anim Waves Input script on the wave particle has the Max Displacement Vertical set to 5.5, which is the amplitude of the wave particle. This means the ocean tile bounds should always encompass the wave particle and should fix the issue.

PR #99, opened 2018-10-15
Fixes #94 - boat instabilities

### LINKED ISSUE #94: Short big jumps in water velocity and height changes
I am seeing brief pops in water and height checks that are very large causing any buoyancy script to pop the object/jitter.  Will work fine for few seconds and then pop.  You can see this, as an example, in the boat align normal script.

Here I am dumping out values from the velWater vector.  The distance also jumps regardless of which distance method I am using ie the ones in boatalign script or sampleHeight.  Seems completely random.

(-0.4, 8.7, -3.1)
(-0.8, 7.0, -2.7)

BOING
(21.9, -27.1, -5.9)
(-0.4, 16.2, -5.9)
(-24.2, 38.9, 1.2)

(-1.2, 6.3, -2.7)
(-1.6, 6.4, -3.1)

Here is a video of it happening.

https://drive.google.com/open?id=1LKf6uOvhXcI-588N6qnvFz6-hYDuTYSL

Weight is set to 1.0 in Shape Gerstner Batched and here is spectrum I am using.

https://drive.google.com/open?id=1a7V5lHP--XaCz6QFQ9fyk_FJ1zZlWWv0

huwb:
Description copied from #94 - fixes for boat instabilities:

* Interpolation added - up until now the wave shape was point sampled. This was fine for using the more detailed lods but giving a boat width of 32 makes the steps really obvious. A useful test is to set Gravity to 0 which freezes the waves and makes this obvious. Performance: this is something to keep an eye on- manually implementing bilinear interpolation generates a surprising amount of code.. and requires 4 reads instead of 1.
* Time stamps recorded with the readback requests so that a velocity can be computed with an accurate DT. Also, I'm computing directly in the readback component instead of in the boat, which is more robust to various things. It requires a bit more CPU memory but i doubt this will make an impact, and there are ways to optimize this down if we need to.
* When lods changed scale, i was computing velocity across two different scales causing the huge spikes in velocity. I now ensure the scale matches the expected and the spikes are gone. The lods change scale based on the viewer - the camera. This is an interesting point - if you get weirdness try freezing the camera position.
* Removed a frame of latency for the physics wave shape, for free (yay)
* Added even more debug - an option that draws crosses in the area around the boat.
For me it now works better than i expected, even in very large choppy waves. Also i was worried that perhaps the async readbacks were dodgy or unreliable, but that doesnt seem to be the case so far, I've managed to make things pretty solid, so thats great.

There is one remaining instability, also related to lods changing scale / camera position. I'm still sometimes using read back data at the wrong lod scale, because readbacks come a frame or two later and the lods can change in the meantime. It will take architectural changes and more time to fix this this is being prepared separately to this PR.

@dizzy2003 - this PR should not affect you guys too much - but the fix that removes the frame of latency is probably of interest! The boat fixed update tells the readback to process any requests that have arrived for this frame before sampling the collision.

        foreach(var aw in OceanRenderer.Instance._lodDataAnimWaves)
        {
            if (aw && aw.DataReadback)
            {
                aw.DataReadback.UpdateProcessRequests();
            }
        }

Not the most beautiful code in the world but the refactor I'm about to start will centralise the LODs and make this a bit nicer. Oh and this would be much nicer in a PreUpdate instead of FixedUpdate but there isn't one, of course!


huwb:
@holdingjason merged!

holdingjason:
Not sure if you noticed this and have no idea what it is but maybe of interest.

https://docs.unity3d.com/ScriptReference/Experimental.PlayerLoop.PreUpdate.html

huwb:
Ha I had no idea! Looks like an ECS thing. I'll keep an eye on it.. Thanks!

https://gametorrahod.com/details-of-the-execution-order-of-ecs-in-conjunction-with-normal-scripts-with-updatebefore-1977de86e920

holdingjason:
Ok that makes sense.  I need to dig into the job system myself soon.  I grabbed that demo they did on flocking behavior using jobs to create thousands of fish which looks fantastic and would like to implement something similar for both fish and birds in the game.

PR #100, opened 2018-10-15
Refactor/centralise gpu readbacks

huwb:
PR not ready to merge yet.

@holdingjason - this is the work to fix up the readbacks once and for all. Changes turned out pretty monumental but its looking good! Restricting which resolutions are read back to the CPU is a nice optimisation. And i haven't noticed any pops with camera movement.

Still some tidy up/fixup to do but thought id post WIP.


holdingjason:
Ok that looks a lot better btw.

holdingjason:
Can you try that out with 64 instead of 32.  Just want to see if that works to ignore waves 16

huwb:
The above turned out to work only when a single shape texture was being read back. Opening it up to multiple shape textures was tricky because

* LOD textures can start/stop rendering based on the dynamically shifting LOD scales, which means they become valid/invalid at random times
* All data comes back asynchronously, to confuse matters even further

It seems to work now in my tests. I've tested a range of different boat widths and spectrums and different viewpoints. I'll sit on it a bit longer and consider if theres anything else that needs testing/changing/etc

holdingjason:
Ok sounds good.  I will pull latest then and run through my tests.

holdingjason:
So a lot better.  Not really seeing the pops, thought I did at first but have not been able to repro so may have just been a glitch on my side (might have had cpu caching turned off when it happens so will see if I can repro with that).

The biggest thing is no matter how high I set the width I cannot get the boat to ignore spectrum 16+.  Mostly interested in ignoring 16 but at some levels it just will not do it.

holdingjason:
@huwb is there anything special I should do to try and eliminate spectrum 16?  Just want to make sure I am not doing something stupid here.  Thanks.

huwb:
Yeah i think i know, and its not very intuitive. I have an idea how to make this user friendly which I'll propose later.

Do you have a Animated Waves Settings assigned to the OceanRenderer?

https://github.com/huwb/crest-oceanrender/blob/refactor/centralise-gpu-readbacks/src/unity/Assets/Crest/Scripts/LodData/Settings/SimSettingsAnimatedWaves.cs

The min/max fields here control which shape textures are read back. I assume in your case it is not transferring any of hte bigger LODs which don't have spectrum 16 removed. Try setting the max field to a large number. 

As i mentioned - not intuitive at all. I aim to make this user friendly and perhaps generate a warning if it does not have the data available to filter out the desired spectrum.

holdingjason:
@huwb Ahhh that did it.  I did not have anything setup for that btw.  So once I did as you suggested I can see it now ignoring correctly.  Thanks a ton.  So what are the ramifications of what this is doing in a bit more detail or point me to the doc when you have time.  Thanks just trying to better understand how all this works.

holdingjason:
@huwb Ok still running into some confusion or issues with this pull and width.

Basically accuracy seems to drop pretty massively the higher I go in the width.  Including a couple of shots showing the problem.

My setup.  

Set objectwidth to 100 or some very large value as you suggested in simsettingsanimatedwaves.
Turn off all spectrums except 32 and set that to 2.  
Weight on gestner is set to 0.3 (dont think it matters)

If I set the boat width (using a simple cube just for ease) to 4 looks great.
image

If I set the boat to 64 (to ignore spectrum 16 and below) looks bad.  You can see how the cross hatches start dropping well below the wave height which did not occur at width 4.  It appears to get worse the higher the width.  So raising the width does appear to not only drop out the lower spectrums but also effect the higher ones.  Was hoping to retain the accuracy on the includes spectrums, ie want to know the height of the water ignoring spectrums below x.  BTW sample height does not appear to work with this at all always is returning the combined height regardless of what is passed into width.

image

holdingjason:
@huwb Was able to hack something (I do mean hack) together to get the action I wanted.  Including the script below.   The only thing that is still busted is we have to have a fixed camera height or we get that popping situation still running width at 64.

The script allows you to apply various force points around your object and applies force based on the distance below surface.  However I had to still check the absolute wave height using SampleHeight to determine if we are above or below the surface and then apply force at various points on the ship based on the SampleDisplacementVel etc.  This is because I want the force that is applied to be based on the non-ignored spectrum not the "real" wave height.  I could not use the displacement value to check though if we are out of the water and should apply force because of the issue I wrote about in my last post.  I tried that first and the ship was constantly applying force at the wrong time either allowing us to go to deep or applying force when it should not (ie out of the water).

This method while ugly as hell does give me the motion I needed.  There are probably ways to improve this so if you have any suggestions please pass them along.  Also feel free to use this anyway you like.

I am sure there are much better ways to do this but best I can come up with so far.  I could not use you boat script because it works great on small boats but on larger ships the motion does not work well pivoting/normal aligning from a central point.

Anyway thanks as always.  

Here is a shot of the layout of my force points inside the ship.  4 along the keel and two on either side.

image


using Crest;
using System;
using UnityEngine;

public class FloaterV2 : MonoBehaviour
{
    public Vector3 CenterOfMass;
    //public float Density = 3;
    public FloaterForcePoints[] ForcePoints;
    public float EnginePower = 10;

    Rigidbody _rigidBody;

    public float TurnPower = 0.5f;

    private const float WATER_DENSITY = 1000;

    ICollProvider _colProvider;
    public float MinSpatialLength;
    public float ForceMultiplier;

    private void Awake()
    {
        _rigidBody = GetComponent<Rigidbody>();


        if (_rigidBody != null)
            _rigidBody.centerOfMass = CenterOfMass;
    }

    private void OnDrawGizmos()
    {
        for (int i = 0; i < ForcePoints.Length; i++)
        {
            FloaterForcePoints point = ForcePoints[i];

            Vector3 transformedPoint = transform.TransformPoint(point.OffSetPosition + new Vector3(0, CenterOfMass.y, 0));

            Gizmos.color = Color.red;
            Gizmos.DrawCube(transformedPoint, Vector3.one * 0.5f);
        }
    }

    private void FixedUpdate()
    {
        if (GPUReadbackDisps.Instance)
        {
            GPUReadbackDisps.Instance.ProcessRequests();
        }

        _colProvider = OceanRenderer.Instance.CollisionProvider;

        UsePhysicsToFloat();
    }

    void UsePhysicsToFloat()
    {
        if (_rigidBody != null)
        {
            //float volume = _rigidBody.mass / Density;
            float archimedesForceMagnitude = WATER_DENSITY * Mathf.Abs(Physics.gravity.y);

            for (int i = 0; i < ForcePoints.Length; i++)
            {
                FloaterForcePoints point = ForcePoints[i];

                Vector3 transformedPoint = transform.TransformPoint(point.OffSetPosition + new Vector3(0, CenterOfMass.y, 0));

                Vector3 undispPos;
                if (!_colProvider.ComputeUndisplacedPosition(ref transformedPoint, out undispPos, MinSpatialLength))
                {
                    // If we couldn't get wave shape, assume flat water at sea level
                    undispPos = transformedPoint;
                    undispPos.y = OceanRenderer.Instance.SeaLevel;

                    Debug.Log("Not Found");
                }

                var waterSurfaceVel = Vector3.zero;

                bool dispValid, velValid;
                _colProvider.SampleDisplacementVel(ref undispPos, out point.Displaced, out dispValid, out waterSurfaceVel, out velValid, MinSpatialLength);

                var dispPos = undispPos + point.Displaced;

                var velocityRelativeToWater = _rigidBody.velocity - waterSurfaceVel;

                float height;

                _colProvider.SampleHeight(ref transformedPoint, out height, MinSpatialLength);
                
                float distance = dispPos.y - transformedPoint.y;

                Vector3 normal;

                if (!_colProvider.SampleNormal(ref undispPos, out normal, MinSpatialLength))
                {
                    normal = Vector3.up;
                }

                if (height - transformedPoint.y > 0)
                {
                    _rigidBody.AddForceAtPosition(archimedesForceMagnitude * distance * Vector3.up * point.Factor * ForceMultiplier, transformedPoint);
                }                
            }

            var forcePosition = _rigidBody.position;

            var forward = Input.GetAxis("Vertical");
            _rigidBody.AddForceAtPosition(transform.forward * EnginePower * forward, forcePosition, ForceMode.Acceleration);

            var sideways = (Input.GetKey(KeyCode.A) ? -1f : 0f) + (Input.GetKey(KeyCode.D) ? 1f : 0f);

            Vector3 heel = transform.forward;

            _rigidBody.AddTorque((transform.up + heel) * TurnPower * sideways, ForceMode.Acceleration);
        }
    }
}

[Serializable]
public class FloaterForcePoints
{
    public float Factor;
    public Vector3 OffSetPosition;
    public Vector3 Displaced;
}

huwb:
Yeah what is that about.. is this the kind of issue you're seeing?

https://drive.google.com/open?id=1Wt9rnfhJBZvDwfAJgZbrualfOJyHpAFk

I have no idea why that would be the case. I'll have to think about it. It might be an off-by-1 error or off-by-half error which becomes exacerbated when the scale of the texture is large.

I'll think about an approach to debug it.

Thanks a bunch for posting that code - my own feeble attempt is in Crest-Examples\BoatDev\Scripts\BoatProbes.cs but i wasnt able to get good results in my brief experiments but i gave up on it. I'd like to try your version out at some point, ill let you know if/when i do.

holdingjason:
I think so.  Basically your green x hashes are not lining up with the water surface.  Some are way below or way above like in your example.  Now I would expect the green X guys to be sometimes below the surface if I was ignoring some lower spectrums that had been set really really high because that is the point ie we are ignoring those spectrums because we don't want the boat to react to that chop.  However I would never expect the green X to be above the entire water surface, that would just be wrong.  Still even the below water ones are not correct either, you can test that by simply turning off ALL spectrums except for one (I did 32) and setting width higher then somthing like 4 say 16, 32 whatever.  The green X guys should line up with the water perfectly but they do not unless your set width to 4 or so.  Then it looks great.

I also rendered a red X cross based on the SampleHeight at the same time and those do not match up at all with the ones being generated now btw.

If you use that script here are the numbers I am using.  Right now some magic number stuff going on which I need to resolve so it works with the mass of the rigidbody and area, but you can tweak until you get the action you want on a specific type of object for now.

image

huwb:
Ok some progress.. my calculations were indeed off by 0.5, so the match should not be tight and good again.

I realised the largest displacement LOD will always pop when waves jump in/out, so its never safe to use this last LOD. I disabled this.

I've added Validation calls - these check whether you're really getting the data you want, and if not they return a reason why not. In BoatAlignNormal i've added warnings when data is not available. It's very chatty right now - it will generate warnings in the first few frames when async data has not come through yet - but its a good first step.

I *think* with those changes things are working to the best of my knowledge, but I'm still mulling if i can make this stuff more user friendly..

huwb:
Btw I no longer believe just filtering out octaves is a good solution for large boats, as it requires very large LODs and it would be expensive to render enough of them to get both small detail and LODs without the large waves. I think manual filtering using buoyancy points is required in these cases like you have done (and of course the align-to-normal physics is all a bit of a hack anyway.. goes a bit nuts in choppy waves..)

holdingjason:
Could you explain a bit more about your last post.  I do not think I followed what it was you wanted me to do.  Are you saying we cannot filter out certain spectrums as we are doing with boat width and if so how do we ignore the smaller chop what I am calling spectrums at like 32 and below?   I agree with not using the align to normal approach and needing multiple bouyancy points.  Though would like to experiment a bit with multiple positions and using some of the normal with those points to move the ship with the wave a bit vs just up and down.  Anyway I think I missed what you where trying to explain.

huwb:
Sure - this is the first time I'm experimenting with large boats (boat width 32, 64 etc) and my last post is my realisation that in order to filter out what you refer to as smaller chop, it requires many LODs. 

My LOD count is 7 which I thought was a larger than necessary number, but if I go to one of the troughs then there with the validation turned on, I get warnings because there isn't a LOD big enough to give waves appropriate for my boat size.

I therefore need to increase the Min Scale of the ocean so that the LODs never scale down this much - but then i would lose my fine details. Or i could increase the lod count to 8 or 9 so that there are some nice big LODs that don't have the choppy waves in them. Both of these sacrifice something (quality or performance), and given the align to normal stuff is a bit dodgy anyway, i feel the only good way to do large boats is with a set of buoyancy probes which will naturally filter/smooth out the water collision when you average over many probes.

huwb:
I unfortunately discovered another type of pop - this time for small boats excitingly! (this has also always been in the system, its not introduced by my recent changes). If the boat is small and wants really choppy waves that arent currently rendering because the viewpoint is too high, but then the viewpoint moves down, the small choppy waves blend in to the rendering, but pop into the collision (the new lod suddenly comes online with weight 1).

One solution i tried was to apply a blend weight to anything rendering into LOD0 based on camera height. This almost worked but getting it from 90% to 100% working turned into a rabbit hole and things were getting more and more complex, and I gave up eventually.

Another solution would be to transition in the new LOD by manual blending in the C# collision code. I couldnt really think of a way of doing this that didnt make a mess of the architecture, and potentially double the cost of collision queries in some situations (to sample twice and blend), and its just nasty.

I'm going to sit on this for now and think of any better solutions, and hope that this is edgy case-y enough that it doesnt affect people in practice, and see if it comes in as an issue.

Let me know if you try the latest and see nay issues, otherwise I'll merge this PR.

holdingjason:
So with the multi-buoyancy probes idea you are suggestion basically doing it the way I am doing it now.  I would still use the boat width but just use multiple probe/physic movers around the ship and the avg should handle it?  Correct?

BTW not sure if you noticed it buried in my other stuff but checking for height using SampleHeight was not working with boat width.  Basically appeared to ignore boat width completely.  Not sure if that is as intended or a bug or if that has been fixed.  Let me know and I will check it out.

holdingjason:
OK then please double check my settings here and see if this all makes sense or if something is amiss. 

I can successfully ignore spectrum 16 and below with these settings.  All appears to be working now and the crosshatches are aligned on the surface correctly it appears.  I am not seeing any popping.  Still not sure if SampleHeight is working so need to test that.  Just FYI for a good reference point all this is to try and mimic a bit of what sea of thieves water acts like from this point of view, really choppy and a lot of action going on around the ship but that is not really effecting the ship, ie large ships.  I may still have the ocean waves a bit to big for the ship but getting closer, we want a nice range of ships from small to very large so there is a nice upgrade path.  Anyway.

LODs set to 9.  
Max Object Width 2048
Boat Width using my floater script set to 33

I am running weight at 0.5 in Gestner and here are my wave settings ( can pass on the actual file if you need it).

512 is off.

image

holdingjason:
@huwb  Oh and if all this looks good then I think we are ok with merging back to main.  Performance wise I am not seeing a big difference going from 7 to 9 LODs.  Actually not seeing any difference which struck me as a bit odd.  Are you seeing an actual difference on your side and if so how are you measuring it?  Maybe see a slight jump in batches etc.

With same setup as above LODs at 9
image

LODs at 7
image


Thanks.

holdingjason:
Removed ship to get a better apples to apples.  So I can see yes there is a jump in vert/tri which accounts for the jump in batches.  Though does not really translate to hit in time much.  Honestly this appears to be a pretty minor hit so livable I would say.  I have much bigger issues with islands, veg and all the other crap :)

LODs 9
image


LODs 7

image

huwb:
Yes that is correctly, I'm suggesting doing it the way you're doing it now.

The height queries should work, so thats something I'll need to look into.

I think/hope the validation function I added will catch the case where there arent enough lods. If you get an "ok" result for all your camera views, you don't need more.

I'll grab some rough numbers for different LOD counts. As always performance is complex. There are different types of cost, some which run in parallel and some serial. But I'll see if I can extract some high level info from a few tests.

holdingjason:
@huwb Ok then sounds good.  Looking forward to it going back to main and hopefully that is the end of these little issues.  Will plan on digging into the underwater skirt stuff next.  See your reward for doing a great job is more work 😒

huwb:
Took me way too long to figure out there was a broken wire - commit reference above. While testing this i added a buoy to boat.unity. Not sure i'll keep it around, but thought it worth pushing in case it helps, it seems to work well for me.

holdingjason:
@huwb It appear to only be an issue with SampleHeight not seeing the CPU Caching have any impact on the other methods of sampling.

holdingjason:
@huwb BTW that is a minor issue so I am OK with merging back without that fixed if that is a big deal.  No big deal either way so up to you obviously.

huwb:
Sigh same issue, different location! Fixed now, looks good to me.

I also centralised the collision settings in the Animated Wave Settings

image

This exposes the option of whether to use GPU transferred collision or CPU calculated gerstner waves, for the first time.

I'll be looking to settle the changes and cleanup etc in the next day or two.

holdingjason:
> Sigh same issue, different location! Fixed now, looks good to me.
> 
> I also centralised the collision settings in the Animated Wave Settings
> 
> image
> 
> This exposes the option of whether to use GPU transferred collision or CPU calculated gerstner waves, for the first time.
> 
> I'll be looking to settle the changes and cleanup etc in the next day or two.

Great news.  I will look forward to that and in the mean time I can bounce over to the skirt and start playing with that to maybe put to bed the underwater issues for swimming etc.  Also will try and tidy up that big ship physics script and pass that along.  Maybe try to also pass along a swim script as well.

huwb:
Decided that while I'm changing everything, may as well clean up the CollProvider interface while I'm at it! Removed some duped up APIs, and after all this trouble with the disp tex approach i thought I'd bring the ShapeGerstnerBatch CPU collision stuff into the fold and gave it a spruce up. Feeling pretty good about the ICollProvider interface now.

@holdingjason i super appreciate the feedback, help and patience.

@dizzy2003 might be quite a merge, but should be a massive improvement in stability. Let me know if you have any questions or feedback.

holdingjason:
@huwb  YOU ARE THE MAN.  Thanks so much for getting this done.  The team is so stoked to get this into the build along with all the other changes and surface visual enhancements that are now in main.

Issue #70, opened 2018-09-10
Detecting Underwater


hayes87:
Hello, this is not actually an issue but a doubt. 
Im trying to figure out the best way to detect if a player is underwater and apply a shader to the part of the screen thats underneath the waves.

Whats the best approach?

Thanks

holdingjason:
@hayes87 Hi.  We are working on the same problem.  We can easily detect when the camera is above or below and modify the shader appropriately but the issue is that the camera is not always fully above or below the water line so you get these spots where your half below half above and it looks wrong.  Not sure yet how to solve it yet so good place to have a discussion and try out some ideas.

huwb:
Thanks for starting this thread. I've seen some games mask it with a flash:

https://www.youtube.com/watch?v=a1ezfA2sb_Q

The look of a camera half underwater is pretty sweet but I'm not sure how easy it is to achieve as you need different post processing for half of the screen?

https://www.youtube.com/watch?v=KIcJgznQWqA

hayes87:
Yeah, i was researching as to how the Witcher3 does this. 
It seems they grab a zdepth of the water, clamp to  get only the line of water that crosses the screen and then some magic to create a gradient with this.

Some screenshots from renderdoc
https://cdn.discordapp.com/attachments/409446549567963151/489451627246387211/unknown.png
https://cdn.discordapp.com/attachments/409446549567963151/489451449592446989/unknown.png

huwb:
Ah thats interesting. Do they produce a half-in-water effect from that? Do you have the rendered result? I looked at some youtube videos but couldn't find a good shot of the camera transitioning into water so far..

hayes87:
Ah sure, forgot to link it  https://cdn.discordapp.com/attachments/409446549567963151/489513816632066051/unknown.png

Heres a link to the full renderdoc capture (300mb)
https://drive.google.com/file/d/1KiR_9kWgQVuowyFlJ97o3lfmeTPWjHWG/view?usp=sharing

hayes87:
Ah sure, forgot to link it https://cdn.discordapp.com/attachments/409446549567963151/489513816632066051/unknown.png

Heres a link to the full renderdoc capture (300mb)
https://drive.google.com/file/d/1KiR_9kWgQVuowyFlJ97o3lfmeTPWjHWG/view?usp=sharing

hayes87:
ops close by accident hah

huwb:
[...]

PR #110, opened 2018-11-01
Underwater rendering with partially submerged view

huwb:
From #70 

Renders a curtain of geometry below water surface to do water fogging/scattering etc. Looks something like this:

image

The code copies the material params onto the underwater material so that they match (when they dont match it creates weird rendering issues). Shader features (like 'Shadowing') are still there and need to be matched manually as these can't be toggled at runtime in standalone builds.


holdingjason:
@huwb Ok pulled the latest and see now that all the values for the underwater skirt are toggles so that I am guessing its syncing up with the ocean shader values.  So anything we want to do to the underwater look would need to be changed in the main water shader ie color etc correct?  I am guessing this is so that we get nice blending between the surface and the underwater skirt?

Issue: Changing transparency on the ocean shader only seems to effect the surface water and not the skirt.  I would has expected the skirt to also become more or less transparent ie can see farther (or less) based on the transparency value, this was I believe how it worked before when they were separate and we needed to sync them manually.  Appears something may have broken in the skirt for this.

Along these same lines does the transparency setting for the skirt (assuming it was bugged above) change at all as you go deeper etc.  I think it is suppose to work as a constant value correct?  Not sure how lighting etc effect it, if at all.  Like what would happen if I bring a light underwater with me for instance.

Minor issue.  I am wondering if we could add something like a depth color similar to what we do with the ocean shader for shallow vs deep water.  Thinking it could allow the skirt color to change as we go deeper.  So near the surface would work like we have ie match surface values, as we go deeper could change.  Just a thought.

holdingjason:
@huwb Ok see what the problem is with transparency and guessing all the settings work this way.  The "copy" occurs on start and so I was changing values at runtime on the ocean shader and not seeing those updated on the fly.  If I mod them first then run it works.  So if I want to change values on fly I am guessing I would do it the old fashion way and update both the shaders at that point via code?

I will check into that but guessing that is correct.

holdingjason:
Does the transparency setting for the skirt (assuming it was bugged above) change at all as you go deeper etc.  I think it is suppose to work as a constant value correct?  Not sure how lighting etc effect it, if at all.  Like what would happen if I bring a light underwater with me for instance.

Minor issue.  I am wondering if we could add something like a depth color similar to what we do with the ocean shader for shallow vs deep water.  Thinking it could allow the skirt color to change as we go deeper.  So near the surface would work like we have ie match surface values, as we go deeper could change.  Just a thought.  Honestly I guess this might fall under custom requirement and probably would be fairly easy for us to add to the skirt shader manually.  Just check depth and lerp in a color mod probably.  Does that sound right to you?

huwb:
Yes i think you could add to ScatterColour and mix in you own stuff based on depth (i_cameraPos.y compared to _OceanCenterPosWorld.y). I'm not sure i'd add this to core Crest, for the time being anyway..

Alright i think tidy enough and good to go, merging.

Issue #106, opened 2018-10-28
Artifacts/Spikes in Simulation by increasing Texels per Wave


gameanimation3d:
If I increase the "Min Texel per Wave" up to 13-15 to create a more decent resolution for close up shots, it start it inject artifacts and after short while these artifacts spike out.
![1](https://user-images.githubusercontent.com/44543590/47614310-55902c80-da9e-11e8-8bb6-30ad346cf5b1.JPG)
![2](https://user-images.githubusercontent.com/44543590/47614311-56c15980-da9e-11e8-8ebd-19149875600e.JPG)
![3](https://user-images.githubusercontent.com/44543590/47614312-5923b380-da9e-11e8-86e0-0ebe79070317.JPG)

I am using a more less static camera on this. Is there a way to prevent this our smooth it out?

huwb:
Wow, you're simulating ripples from rain drops, thats cool!

Something like this needs high frequency updates - the dynamic waves currently update at the same rate as the rendering, which can be quite low. To make something like the above work it would need at least 60hz. What sort of framerate is your test running at before it explodes, out of interest?

I've never set the detail level up that high before, 5 for "min texels per wave" was usually a good value for me. Did you need to set it this high to get these small ripples?

gameanimation3d:
Hey,

Thank you very much for your quick response.
I used an higher "min texels per wave" as we are pretty close to the water and i want to get decent resolution to have finer details. Especially with more particle spawned(heavy rain) and more ripples are created.
We want to use this ocean system not for a game, we are using Unity for Realtime Rendering for an 3D Animation short were want to create the "fluid simulation/water sim" with this ocean system. So the framerate doesn't matter with Timeline but the Visual Quality.

At this particular example it explodes on 53 FPS but it takes 30secs to explode. But this depends on the amount of particle that get spawned. I found that including an Shape Gerstner Batched into the Scene with a low Weight(0.05) kind of reduces the effect. But still mosaic artifcats like in image 1 are shown.

How can i increase the high frequency updates?

huwb:
Wow, thats cool!

In the past I tried and failed to double the wave sim update frequency ( #44 ) - it didnt fail for any good technical reason, but unity just did not seem to allow me to do read/write from the right textures. it might have been a bug though, and i really should get this fixed, so i'll see if i can have another attempt this week.

Midda-C:
[...]

Issue #79, opened 2018-09-25
Optimizations


gamemachine:
So I'm tackling performance and while the optimization readme has some good stuff, what I'm finding is a lot of little things that are adding up.

I'm going to advocate for a performance by default approach.  Having to modify the code to get production performance should not be the default.  Ease of use things like the ability to tune everything at runtime should be secondary.  Real games have to take that approach, there is no other choice (other then having both).

The thing is the little stuff adds up.  I'm closing in on 2ms of main thread cpu time gone from over a dozen little things.  It's all little stuff but in the aggregate it adds up to something substantial.

The challenge is that right now Crest is unusable for many games that would use this type of water.  Games simply don't have 8ms of budget for water, which unless you disable things to the point of the water not even being interesting anymore, is where it's at.  That's half of your main thread budget for your entire game.  Realistically it should be around 2ms, I think that's entirely doable but only if little stuff matters.

There is a lot I like about Crest, and I knew performance was an issue going in.   And that most of the performance issues looked to have fairly straight forward fixes.  My fear is that the lack of attention to performance will lead to it not becoming straight forward as more and more features get piled on.

dizzy2003:
Are you profiling in editor or stand alone, are you using unity profiler or
toggling features and observing the resultant change?

Also I'm a bit confused because you say its taking nearly 2ms then you say
it shouldnt take 8 ms then you say it should take more like 2ms.

huwb:
Yes, please post data and methodology. I'm a games developer myself who typically end up working on perf when finaling projects and I understand the domain.

Performance is a key concern for me and I've built things to be efficient where possible, but I have limited resources and limited access to hardware  so I'll need to rely on you guys to help ID perf problems. That starts with data and we can discuss from there. That's the limit of service that is currently available in crest I'm afraid.

Most of crest features are on toggles as dizzy notes so there's a bunch of scope for experimentation - as long as methodology is good. The frame rate should be unlocked in standalone builds in the crest unity project so I typically run standalone release builds on my laptop with fraps to get a high-level max(CPU time, GPU time) number.

moosichu:
"I'm going to advocate for a performance by default approach. Having to modify the code to get production performance should not be the default. Ease of use things like the ability to tune everything at runtime should be secondary."

[...]

Issue #61, opened 2018-08-30
Crest requires lots of layers


gamemachine:
Number of layers required is a bit much.  The demo requires 4 (Terrain is not mentioned that I could find in the docs).  This actually used up all remaining layers in my project.   Layers tend to be at a premium in large games.

gamemachine:
Saw the ocean needs a layer also.  That's 5 layers now.

huwb:
Yeah, this comes up now and then, theres another team in a similar situation and I'm becoming increasingly aware of the squeeze. It's a shame that the number of layers is so small, and that physics and rendering share layers. But thats the constraints we have i guess!

It's there for flexibility so you can assign whatever you like into the various data like water shape, foam, dynamic sim, etc. If you're not using any of these you can not add the relevant layer (i've not done a huge amount testing in this direction so it might need one or two things unpicked but it should be easy to make it work with only the layers you want/need).

However I get the point and this kind of functionality could perhaps be achieved using tags. Scriptable render pipelines may change this up as well, which is something i plan to do research into soon. I'll leave this open and post back when i have any thoughts or progress, or perhaps someone else looks at it in the meantime.

huwb:
Yeah im thinking along the same lines.. In my past experience I haven't loved doing this kind of thing because theres the state to manage/update when new objects get created, and when stuff gets destroyed you dont want to hang on to references and prevent stuff getting GC'd. Or maybe im making a mountain out of a mole hill. I'll put it on my list to experiment with this stuff soon and find a nice pattern.

gamemachine:
I think it would involve more work then appears at first glance.  It also looks like it wouldn't be too much work to just add work arounds where needed per project until a better overall solution hits.  Big picture I don't see this as a show stopper for real teams.

huwb:
There is some evidence from only THIRTEEN YEARS ago that just doing a FindObjectsWithTag may not be as bad as it sounds, in practice (which would remove the state that needs to be maintained). It seems to be efficient provided you don't use tags for too many GOs, which he suggests people typically dont. Do you use tags much in your projects?

https://forum.unity.com/threads/findgameobjectswithtag-efficiency.3612/

gamemachine:
I mostly see tags used by third party assets.  They do seem like an ok integration point assuming that in context they won't be a performance issue.  And they aren't a limited resource which is a big plus.

One potential downside is what stuff created at runtime.  And I say potential just because I have no idea of the possible issues there not having used tags in a long time.

[...]

PR #117, opened 2018-11-04
Refactor/move to command buffers

### LINKED ISSUE #44: Wave sim intensity is very frame-rate dependent
Hi @huwb !

Looks like Wave sim/Foam sim intensity are very related to framerate.
I have tested this on 3 diferent machines:

* Machine 1: Core i5, Intel HD 620 (Around 20 fps avg)
* Machine 2: Core i7 7500, Geforce GTX 950M (Around 35 fps avg) (approx)
* Machine 3: Core i7 8700, Geforce GTX 1080 (+80 fps)

On machine 1, there is a lot of foam just after starting accelerating the ship a little, and after get some speed, the generated wave is way higher (in height) than desired.

On machine 2, is also noticeable there is a lot more foam than normal, and that waves go too up as well, but not as much as on machine 1.

On machine 3, there is almost no foam (unless you go pretty fast and turning), and wave sim is way smaller, not getting as high as any of the other machines.

I've just double checked this on latest https://github.com/huwb/crest-oceanrender/commit/d4d8536507f7237f75f94a473331a6a296e506a2 commit from master.

I think that some time-dependent values are not multiplied by `Time.deltaTime`, or maybe we are mixing fixed time step with regular time steps. I'll try to figure out what is causing this.

Thank you!

huwb:
(PR Status: tested on branch, ready for further testing or merging into master)

This PR refactors the entire loddata system. I'm moving it to a component manager system because it was wasteful having gameobjects+cameras for each lod data type which were not necessary. 

The new system will build a commandbuffer for all the loddata, should be more controllable, simpler, cheaper, and should allow me to run the wave sim at higher frequency which should close #44 and #106, and on top of all that it will be a big step towards scriptable render pipeline support. It also removes the need for layers completely (#61).

Perf has improved signficantly in a standalone build running at 320x200 resolution, i get 580fps on my laptop. When i recently did this test last time, i got 360fps (#79), so this PR shaves 1ms off the CPU frame time for me.

Impact:
* Render Ocean Depth script changed to Register Ocean Depth Input, to bring it in line with new naming conventions. Usages of the old script need to be switched to the new.
* No more layers - the ApplyLayers component has been removed completely, and will give unassigned script warnings if you are using them in your project. If you had cull include/exclude layers specified they should be applied directly to the culling mask on the camera. Simply removing the unassigned scripts should be enough to get going in most cases.
* I did a big rename and cleanup of shaders and materials which may produce breakages or merge conflicts. This probably mostly affects usages of shaders that add foam/waves/flow/etc. These were not widely used partly because they were confusingly named and inconsistent, which was the reason for this cleanup.
* Script execution order is simplified and made explicit. I've added documentation to the TECHNOLOGY.md page.


huwb:
Finally i think im at the end of this - i just have a few last small tweaks to example data.

It may need some fixup in your projects - see the impact above. If anyone is keen and has time to early-adopt this to help me test the changes, let me know. If not then i'll merge it in to master some time soon.

@moosichu @holdingjason @midda-c @dizzy2003 @mschwarz1

holdingjason:
sure I can pull it over and give it a go.  This is all in refactor/move-to-command-buffers so will yank it from there.

Tsetso:
I'll do my best to try out the latest features as well. Already pulling any new changes aside...

moosichu:
Testing it out now

moosichu:
There's a lot of stuff to work through, and although I'm the inevitable bugs crop-up here and there, it all looks good to me. Tested out upgrading to this on a small project and it worked pretty well!

huwb:
Thanks @moosichu! If anyone else has feedback (or wants more time to test) please let me know.

holdingjason:
A bit more time.  Moved things over into the project but getting a bunch of warnings about my floating stuff and not seeing any waves, acting like its just calm water right now.  Not sure what is going on so comparing my stuff to your example project to see.

Validation failed: NoDataAtThisPosition. See comments on the AvailabilityResult enum.
UnityEngine.Debug:LogWarning(Object, Object)

huwb:
Damn sorry @holdingjason - this was a todo that i didnt mark as such and it slipped through the net. The problem was that there is unfortunately no shared event between forward and deferred paths that i can hook to, so i added code to choose the event based on path with some validation to catch this kind of thing in the future. Could you please grab latest and try with that (commit referenced above)? Thanks..

holdingjason:
Ok that seems to work fine but getting a flicker now and again.  I need to check to see if I was always getting this or not and just noticed it but I did not think so.  I believe it is related to a garbage collection issue since it seems to time up pretty well with the profiler.  It is caused by calls to GPUReadBack ie sampling water surface just like/similar to how you are doing it with the boats and the boatalignnormal script.  I used the boats example and was able to replicate the issue so you should be able to see it as well.  Basically calling it over and over in the fixedupdate as we are generates a lot of GC Allocations which eventually trigger a punch in the head GC Collect (this can 10+ milliseconds in a single frame) which I think is causing the  hiccup.

Here is the deep profile I grabbed during a normal frame (not the GC Collect one).  It is allocating about 1.4KB of memory each frame.  This is using the 3 boat scene.  This of course gets worse and worse the more things that call this sequence of course, so the more floating objects the worse this alloc gets and the more times GC Collect will get triggered and worse it will be.

I believe the issue can be tracked down to foreach loops the gremlins of memory alloc.  Basically you should only use them when your doing one off calls infrequently.  In loops or repeated calls they used to be complete death, this has been improved however here is an excellent article on which lists you can foreach and there impact.  Unfortunately it appears the sorted list is still the worst of the lot.

https://jacksondunstan.com/articles/3805

image

holdingjason:
BTW will be out of town Thur and back next Monday so just FYI bad timing on my part.  I will be available to answer questions just wont be able to test anything but will be able to tonight and tomorrow morning my time (Las Vegas).

holdingjason:
@huwb  Ok here is the video of the flicker.  I verified that I do not see this kind of flicker in the mail build just this one using the command buffers.  I do get a bit of a hiccup in both but its more of the normal hitch you see when frame rate drops off because of an errant process (like GC etc) vs this kind of video flicker.  

https://drive.google.com/open?id=1CyeK75aP99l_1H-a22frcLiruaZNFJzt

holdingjason:
@huwb Got it, not related to the above issues.  I still think we have an issue with for each loops btw because I can still see decent hit 1-2 ms hit with 10+ floating objects which is mostly caused by the memory alloc.

So in the three boats scene I set this up (not sure you need these settings but whatever)

Lod Count max 16.
WavesModerateSmooth weight 1.0
Main Camera Cam Controller Lin Speed = 50

Just take off in forward direction for a bit and you will catch the flicker as I think it bounces between LODs.  Here is a video that captures it with data.

https://drive.google.com/open?id=1D85_kQ32SwdSAF9O4KWVojDfcYKZMJ2I

huwb:
Good catch with that one frame issue. It was an update order problem, i've made the update orders simpler and explicit now.

I did a pretty ugly thing to stop garbage from sorted list. Should be good to go now.

I found another GC from a FindObjectOfType which i had neglected to clean up! This led to a good fix.

Thinking/hoping its pretty solid now. Just remembered you're out of town @holdingjason , i might merge this into master in the next day or two so i can clear this off my plate and start on other stuff. Last call for any testing!

holdingjason:
Yep poor timing so can't test before that so go ahead.  Those we're all I noticed so should be good.

holdingjason:
Great job btw I see a significant performance increase with this method and hope the GC stuff helps as well.

huwb:
Merged, thanks all!

PR #133, opened 2018-11-27
Feature/flow gerstner waves2

huwb:
Up until now flow affected the sims (waves, foam) and the normals, but it did not affect the actual waves.

This change now applies flow to the gerstner waves, and it does so in the combine pass (and only if flow is on), so it came through fairly elegantly. It also sets the period for the flow cycle to be proportional to lod scale so that e.g. large waves are flowed with a large stride, which works really well. Also i think having different flow cycles helps break up the visible 'pulsing' artifacts that are sometimes evident with flow techniques.

Here's a video from the whirlpool scene with the waves jacked up like crazy: https://drive.google.com/open?id=11rn8ebDYomrYPoEBYYGLv_3jPqO13bDs


moosichu:
Holy moly that looks awesome

huwb:
Was a team effort :)

src/unity/Assets/Crest/Shaders/Resources/ShapeCombine.shader
```diff
@@ -44,6 +47,17 @@ Shader "Hidden/Ocean/Simulation/Combine Animated Wave LODs"
 
 			uniform float _HorizDisplace;
 			uniform float _DisplaceClamp;
+			uniform float _CrestTime;
+
+			void Flow(out float2 offsets, out float2 weights)
+			{
+				const float period = 3. * _LD_Params_0.x;
```
moosichu: `OceanNormalMapping.hlsl` also does this (with a different period as well?). So we probably want to have that call this function.

moosichu:
```hlsl
	const float half_period = 1;
	const float period = half_period * 2;
	float sample1_offset = fmod(_CrestTime, period);
	float sample1_weight = sample1_offset / half_period;
	if (sample1_weight > 1.0) sample1_weight = 2.0 - sample1_weight;
	float sample2_offset = fmod(_CrestTime + half_period, period);
	float sample2_weight = 1.0 - sample1_weight;
```

Is the code I'm referring to.

huwb: cool thanks. for the other uses we'll need to decide on the period if we switch to this function, and it would be good to test a similar approach to choosing the period for the normals as these are scaled with lods (whereas foam is not.. currently..). this will take some thinking and testing and i think im going to take the lazy route and tackle that in the future if that sounds acceptible for now.

moosichu: Sounds good 👍

huwb:
@ftejada17 thanks! truly breaking waves are tricky due to extreme stretching of the ocean surface and then the changing topology when it breaks. however something that has the appearance of breaking (steep wave with foam) might be more doable. i'm currently hoping to get some nice rivers working but after that i'll consider taking another look at breaking waves, ill @ mention you if i do.

huwb:
Merged manually. Closing

PR #143, opened 2018-12-10
Refactor/sampling data

huwb:
Big simplification and rewrite of collision provider. Old system had hidden state and was confusing.

Now SamplingData must be provided for every query, and this data has any prewarm info to ensure efficiency.

This has removed a whole bunch of redundant API.


huwb:
I created a new PR for this as I messed up the branches on this one. #145

Issue #140, opened 2018-12-09
Wave height queries appear to be dependent on camera position


skaughtx0r:
I was playing around with switching over from UWS to Crest, but I noticed there's an issue with the wave height queries. It seems like its very dependent on the camera position. See the video, I'm just changing camera angles using Cinemachine and the water seems to change dramatically affecting the physics on the boat.

https://www.youtube.com/watch?v=OIUj39MqQ34

I'm using Crest on master branch.

Am I doing something wrong or is this a bug?

Also with the very low FOV on the TV camera, the water becomes very flat.

huwb:
Hey, cool video!

The ocean has a number of lods that are centered around the viewpoint. See the **LOD data** section here: https://github.com/huwb/crest-oceanrender/blob/master/TECHNOLOGY.md . 

I think in this situation i would not sample the highest detail lods for the physics, as they will only cover an area near the camera. If you are using BoatAlignNormal then increase the *Boat Width* param. This is described in the **Collision Shape for Physics** section (of the above doc).

If you need the physics to be truly independent of viewpoint, you could compute the collision directly from the waves instead of using the LOD data, see the **Gerstner Waves CPU** section.

huwb:
Oh the very low FOV thing - i had never thought about that before. I guess this is an issue for all distance based LODing. One thing you might try is to make the ocean detail follow the boat instead of the camera, maybe at certain times. You can do this by setting the *Viewpoint* on the ocean renderer to be the boat.

Or you can brute force the solution - just jack up the quality params described in **Ocean Construction Parameters** in https://github.com/huwb/crest-oceanrender until the problems go away! (but do test performance!)

Let me know how you get on, and I'd love to see more of your stuff in the future if you ever feel like posting!

huwb:
Sure. 

I naively would have thought running a sim for prediction might be ok enough even if the water is smoother or flatter than it should be - but maybe not.

I'm not sure either about the server, i guess not.

No the cpu only includes the gerstner waves, as you probably suspected. More could be added but im not sure how efficiently.

On efficiency, i have started experimenting with a more efficient way to execute CPU height queries, see my last comment on #127 .

skaughtx0r:
It might be ok, if I change the Viewpoint to the boat that the camera is targeting and make the server use CPU. I will need to play with it a bit. I did have to modify the Viewpoint property on OceanRenderer, it needs to have a setter, so I can update it with the current camera target.

[...]

Issue #127, opened 2018-11-20
Getting the height of water at X,Z without a massive fps drop.


ImtehQ:
I got a script running to get the height of the water like the little boats do in one of the demos, however once you go over lets say 10 calls a frame (that's 1 points) it drops the fps a lot because it takes forever to get the data back from the ocean.

Here is the code im using for it:
`            position = new Vector3(x, 0, z);

            undispPos = Vector3.zero;
            if (!colProvider.ComputeUndisplacedPosition(ref position, ref undispPos)) return -999;
            if (!colProvider.SampleDisplacement(ref undispPos, ref _displacementToBoat)) return -999;
            if (!_displacementToBoatInitd)
            {
                _displacementToBoatLastFrame = _displacementToBoat;
                _displacementToBoatInitd = true;
            }

            var dispPos = undispPos + _displacementToBoat;
            return dispPos.y;`

![Image of debugger](https://i.imgur.com/HDHcp5g.png)

Is there a better way of doing this?

holdingjason:
This might be related to a post I just made over here.  I am noticing an issue with GC for these calls which I think is causing some (most/all/not sure) of the issues.  Can you open up the hierarchy on those and see where the biggest chunk of time is being spent.  You can see on mine that it was GC alloc. 

https://github.com/huwb/crest-oceanrender/pull/117#issuecomment-440372739

ImtehQ:
There no GC Alloc to speak of with 1 point, however there are tons of calculations being made.

![screenshot](https://i.imgur.com/dMyxzMd.png)

holdingjason:
Ok that is weird.  I am doing a bunch of these calls as well inside of a floater script via fixedupdate and not seeing a big hit.

here is a shot of my frame and shows 24 hits to fixed which is doing the following code.  I know I have about 20 of these being triggered right now.  So not sure why you are seeing such a wallop.

![image](https://user-images.githubusercontent.com/42001313/48848804-bf2bef80-ed59-11e8-924c-60d4305f29c4.png)

private void FixedUpdate()
    {
        if (_active)
        {
            if (GPUReadbackDisps.Instance)
            {
                GPUReadbackDisps.Instance.ProcessRequests();
            }

            if (UseNonPhysics)
            {
                UseNonPhysicsToFloat();
            }
            else
            {
                UsePhysicsToFloat();
            }
        }
    }
    void UsePhysicsToFloat()
    {
        if (_rigidBody != null)
        {
[...]

PR #157, opened 2019-01-01
Feature/gerster jobs

huwb:
(PR status: early dev, not ready to merge)

Background: #127 and #140 

Performs gerstner height queries using burst jobs. Moving to jobs means getting work off the critical path (main thread), and using burst gives a consistent 10x performance boost (!), so the queries have become essentially free. On top of this i managed to vectorize the gerstner computation which added another 3-4x speedup. I tested with 1000 height queries against 60 wave components and the total wall clock time for all queries was just under 1ms (performed on idle cores, not on the main thread). **UPDATE** i've now factored out a bunch of calculations and reduced the time down to 0.3ms.

As is being discussed in #140, i think a good setup is to disable GPU readbacks (set collision source to None in animated wave settings asset) and rely completely on the jobs to get water height. this will only work on open water (not attenuated by sea floor) and will not include dynamic waves/wakes. the sea floor could probably be accounted for by reading the ocean depth back and using it in the job, but this will add some complexity to the implementation and is not done for now.

The jobs are turned on by default and are used by BoatProbes. To use the jobs, follow the pattern demonstrated in this script (precisely).

This loops in 2 preview packages (mathematics and burst), it seems to work well for me but i dont know what support/stability is like on different platforms, and taking a hard dependency on these will kill backwards compatibility with previous unity versions. I think before this could be merged back to master i would disable them so its opt in, for now at least.


skaughtx0r:
I tried to do a quick test last night, but ran into some issues that I think are unrelated to this PR. I started integrating the job queries into my modified version of the Dynamic Water Physics asset and it seemed like it was getting the values properly and was using almost no CPU time, but was having issues with my floating objects sinking. I tried switching back to the GPU queries and it was also sinking, so I'm not sure whats going on there, I'll need to dig into it a bit. I'll probably have more time this weekend to test.

skaughtx0r:
The 2nd pass is to get the water height at the center point of the submerged part of a partially submerged triangle. I'll play around with using the center point of the original triangle. My guess is that could lead to inaccuracies if the original triangle is large and only a small portion of it is submerged.

huwb:
Oh yeah - now I get it. Yeah the dependency thing could make sense actually.

But you might just try running the jobs synchronously. Create some query positions, schedule, complete. Create some more at the centroids, schedule, complete. There might be some overheads running this way compared to showing it to run across frames, but I'd wager you might get a good trade-off between speedup and complexity. Your still gonna get ultra fast burst compiled code running on multiple cores.

mocudev:
I'll probably be re-visiting this PR again soon, do you think everything will still work if I roll the math and burst packages back? I'm going to be stuck on 2018.4 for a while since they removed UNet from 2019.

skaughtx0r:
I'll probably be re-visiting this PR again soon, do you think everything will still work if I roll the math and burst packages back? I'm going to be stuck on 2018.4 for a while since they removed UNet from 2019.

huwb:
Yeah I reckon it will

KellanHiggins:
@skaughtx0r I've got this almost done on a branch of crest. Burst and Math will work fine with 2018.4. Although It's been reported that more than one water tester object won't return proper heights, so I gotta do some more testing but it now works with ocean depth as well.

skaughtx0r:
OK thanks, I'm also curious if this can be mixed with GPU height queries. I think the way it is now you can only choose GPU or CPU. For my game when I'm recording replays of the race, I need buoys and boats further from the camera center to get the correct wave heights. I would also like the dynamic effects like boat wakes to affect the physics. I'm not sure if its possible to include those into the CPU calculation or if there is some way to mix the dynamic effects from the GPU readback?

KellanHiggins:
@skaughtx0r No, not really. What the system is doing is calculating a single point with the CPU instead of from the GPU. Right now it only takes into account the waves and the ocean depth cache (which is all I need for my game). What you would need to do is pull the code out of the shaders that calculates the dynamic waves and feed that into the job system. It is possible to do if you understand shader code (which I barely do).

skaughtx0r:
Are there plans to port this to the new Query API? I'm not sure if compute shaders work with headless server, I assume not. In the short term I'll probably just assume no waves on the server, since my water is pretty flat.

huwb:
Thanks for flagging the API breakage. It's been a while since i dug up this branch.

We don't have plans at the moment, but if enough people hit this we'll see if we can invest time in it.

KellanHiggins:
I did a ton of working trying to get this system to work just with jobs @skaughtx0r but in the end it basically required me to copy the shader code into a C# version and use jobs for it. The regular waves worked pretty fine, but adding anything like displacement or dampening due to shallow water was super hard (cause I am not a shader guy). Now that you can ping the GPU for water heights, I think this branch should be scrapped and removed. To use it on a GPUless system, you'll have to copy all shader code as it evolves into C#.

huwb:
Yeah... the problem with this stuff is we're a highly GPU-oriented system. All of the ocean simulation is done on the GPU and has nontrivial things like shallow water (potentially driven by dynamic objects), manual inputs into the shape such as adding to the water height, dynamic wave simulation.

But i think it still stands that if you have open water, and only gerstner waves, it should be possible to compute queries in a job, and @skaughtx0r makes a valid point that we could update the API to match recent changes on the C# side (the new query API). actually the new query api was heavily inspired by this api and should map pretty directly..

skaughtx0r:
@huwb Just a heads up. We finally ran our first ever online test race using Crest and everything seemed to work just fine using a flat water model on the server, so I don't think I have a need for this anymore.

KellanHiggins:
Yeah I think this branch should be deleted. It is going to be a pain in the ass to maintain. The best thing would be to write a system to pull the shader code and convert it to C# burst things. Has anyone tried running Crest on integrated graphics chip? Because if you need a headless server maybe the integrated graphics card can do enough to calculate the levels without actually rendering anything. That way you can run it on a small server.

huwb:
Thanks both, and congrats on the test @skaughtx0r 

Yeah i dont think this will ever merge, but its useful for some so worth keeping around. I might close the PRs though.

I would guess headless is truly headless in most cases, i dont think any graphics is available in these contexts like AWS or such. I could be wrong though.

PR #162, opened 2019-01-06
Complete and reintegrate underwater reflection shader

moosichu:
The previously WIP underwater shader is now much more physically based:
high FOV view of underwater

Here's a screenshot of [Snell's Window](https://en.wikipedia.org/wiki/Snell%27s_window) in effect:

Screenshot of Snell's Window

The `OceanReflection.hlsl` file is also cleaned-up in general to bring it in-line with the rest of the code base, and the refractive indices of air and water are now exposed and adjustable in the Ocean shader.

### DIFF
--- a/src/unity/Assets/Crest/Shaders/Ocean.shader
+++ b/src/unity/Assets/Crest/Shaders/Ocean.shader
@@ -387,12 +387,10 @@ Shader "Ocean/Ocean"
 						ApplyReflectionUnderwater(view, n_pixel, lightDir, shadow.y, i.foam_screenPos.yzzw, scatterCol, col);
 					}
 					else
+					#endif
 					{
 						ApplyReflectionSky(view, n_pixel, lightDir, shadow.y, i.foam_screenPos.yzzw, col);
 					}
-					#else
-					ApplyReflectionSky(view, n_pixel, lightDir, shadow.y, i.foam_screenPos.yzzw, col);
-					#endif
 
 					// Override final result with white foam - bubbles on surface
 					#if _FOAM_ON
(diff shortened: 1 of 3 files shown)


src/unity/Assets/Crest/Shaders/Ocean.shader
```diff
@@ -379,7 +381,18 @@ Shader "Ocean/Ocean"
 					half3 col = OceanEmission(view, n_pixel, lightDir, i.grabPos, pixelZ, uvDepth, sceneZ, sceneZ01, bubbleCol, _Normals, _CameraDepthTexture, underwater, scatterCol);
 
 					// Light that reflects off water surface
+					#if _UNDERWATER_ON
+					if (underwater)
```
huwb: Minor comment - you could avoid the duplicate calls to reflection wrapping the if/else:

#if ...
if(underwater)
{ ...}
else
#endif
{
  reflection sky ()
}

I personally find this easier to follow but it might just be me, ill leave it up to you.

huwb:
This looks excellent. I had only the minor suggestion. Looks clean and well written.

I tried it out and the physical values were too intense for my tastes, however by increasing the IOR for water i could get something i really liked:

image

What do you think? If you think its ok like this i might make this change to the example material in *main.unity*, but let me know if you prefer it with physical values.

In any case thumbs up from me, merge it in!

moosichu:
Yeah, I think going for stylised makes sense. I was looking at footage of other games which actually seem to do the same thing for artistic reasons.

Let me apply the cleanup you suggested, then I'm happy for everything to be merged. :)

moosichu:
@Midda-C I thought I would let you know that this is being merged in as you expressed an interest in it previously.

PR #164, opened 2019-01-07
Fix shadows not removing caustics in underwater curtain

moosichu:
Before:
image

After:
image

### DIFF
--- a/src/unity/Assets/Crest/Shaders/OceanEmission.hlsl
+++ b/src/unity/Assets/Crest/Shaders/OceanEmission.hlsl
@@ -154,8 +154,18 @@ void ApplyCaustics(in const half3 i_view, in const half3 i_lightDir, in const fl
 	{
 		// only sample the bigger lod. if pops are noticeable this could lerp the 2 lods smoothly, but i didnt notice issues.
 		fixed2 causticShadow = 0.;
-		float2 uv_1 = LD_1_WorldToUV(surfacePosXZ);
-		SampleShadow(_LD_Sampler_Shadow_1, uv_1, 1.0, causticShadow);
+		// See comment for underwater code ScatterColour for why this is done
+		#if _UNDERWATER_CURTAIN
+		{
+			const float2 uv_0 = LD_0_WorldToUV(surfacePosXZ);
+			SampleShadow(_LD_Sampler_Shadow_0, uv_0, 1.0, causticShadow);
+		}
+		#else
+		{
+			float2 uv_1 = LD_1_WorldToUV(surfacePosXZ);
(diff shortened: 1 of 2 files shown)


moosichu:
I don't know if my solution is hacky or not, and feedback appreciated!

huwb:
Ah good find. Attempt at a (very tired) review:

Should it just sample from that one lod? Lod index 1? Not saying its wrong but if you could write a few words about how you arrived at that it would be useful, thanks.

You shouldnt need the #define i dont think, you can pass it in as an argument to the function. if the parameter is always constant the shader compiler will optimize out the branch anyway so branch cost should not be an issue.

Minor nitpicks, If you get a chance could you look at the whitespace changes and i think "underwater" is one word, i remember googling that and deciding whether to capitalise!

moosichu:
Sure, I will sort out whitespace changes.

With regards to underwater spelling, I was making it consistent with other functions in the file, but will change those instead.

I can have a look at sampling. I was simply applying the principles found here:
```
if (i_underWater)
	{
		// compute scatter colour from cam pos. two scenarios this can be called:
		// 1. rendering ocean surface from bottom, in which case the surface may be some distance away. use the scatter
		//    colour at the camera, not at the surface, to make sure its consistent.
		// 2. for the underwater skirt geometry, we don't have the lod data sampled from the verts with lod transitions etc,
		//    so just approximate by sampling at the camera position.
		// this used to sample LOD1 but that doesnt work in last LOD, the data will be missing.
		const float2 uv_0 = LD_0_WorldToUV(i_cameraPos.xz);
		float seaFloorHeightAboveBaseline = 0.;
		SampleSeaFloorHeightAboveBaseline(_LD_Sampler_SeaFloorDepth_0, uv_0, 1.0, seaFloorHeightAboveBaseline);
		depth = DEPTH_BASELINE - seaFloorHeightAboveBaseline;
		waveHeight = 0.;

		fixed2 shadowSoftHard = 0.;
		SampleShadow(_LD_Sampler_Shadow_0, uv_0, 1.0, shadowSoftHard);
		shadow = 1. - shadowSoftHard.x;
}
```
In the `ScatterColour` function.

I will make the underwater a function param.

Will do all of the above, tomorrow.

huwb:
Thanks, thats the background i could not remember! if you could copy that last line of the comment about lod1 being missing and put it with your code,, that would be great.

Sorry about the underwater inconsistency :( . yeah i htink it should be one word. thanks.

With that i reckon its good to merge but id like to have a test with it, should be able to tonight.

Thanks!

huwb:
Looks good to me, tried it locally and it seems to work for me. Go ahead and merge it!

PR #173, opened 2019-01-19
Vectorised gerstner batch shader

huwb:
Hand vectorised the gerstner computation shader. The wave computations are mostly scalar, so 4 wave components can be computed simultaneously.

On my laptop i wasnt able to measure a significant difference but im hoping it helps a lot on switch. Would be super interesting to hear how it goes if you get a chance to try it @dizzy2003 (at default quality settings would be awesome for comparison, if possible).

There is more that can be done - much of the computation in the gerstner loop body can be baked into the incoming uniforms. Only the stuff that depends directly on the world position needs to be computed per pixel - in theory at least. However without seeing a noticeable perf difference im flying a bit blind here and would be interested in hearing how this first round of optimisations lands before going further.


huwb:
Ok i figured out how to get nsight graphics to not crash (using DX12 instead of DX11 seems to do the trick) and so i pushed forward with optimising further.

Similar to the gerstner jobs, i pushed all possible computations out of the shader.

On my laptop the total cost of rendering the gerstners has dropped from 0.36ms to 0.18ms (reading the time for the gerstner shader group/range). It's probably at the point where other things are bottlenecking it now. I hope the speedup could be even more substantial on switch.

huwb:
The 2x speedup also seems to hold on switch - 3.9ms down to 2.05ms. Merging in.

PR #174, opened 2019-01-23
Fix the fact that Planar reflections were not being distorted

### LINKED ISSUE #172: Planar Reflections Broken?
In the latest master, planar reflections no longer distort with the waves? they look like slapped on the screen.  You can test it with main crest example just enable planar reflections, now changing wave height does not make any difference on the reflections at all. 

p.s. This was working perfectly from a build a few weeks ago. I can provide screen shots if you like. Not sure what could have changed.

moosichu:
Resolves #172


moosichu:
@superjayman Can you please let me know if this fixes the issues you have been having? It should do the trick but just wanted to double check.

PR #181, opened 2019-02-03
Feature/decoupled geo density

huwb:
Alright I got what might hopefully be another chunky perf improvement for you @dizzy2003 .

Up until now the geometry/vert density exactly matched the lod data texture resolution (so e.g. every displacement texture value maps to 1 vert). Having this detail in the geometry is good for the shape but we can get away with significantly lower geometry detail as long as the shading still takes the detailed normals, with little impact on visuals. Also small tris can spell trouble for quad utilisation. @moosichu and I discussed this some time ago but it took me a while to figure out a good approach to implement it.

Here is an example with full geometry density (note the tri count - 460k):

image

Here it is with 4x lower resolution geometry, corresponding to a 16x reduction in verts so just 32k:

image

This is how it looks with wireframe which shows just how much geometry has been decimated:

image

My plan would be to make the crest default to a 2x reduction in geometry, but i reckon you could get away with at least 4x on switch.

On my laptop this 4x test reduced the cost of drawing the ocean surface from 2.22ms to 1.58ms, a 29% reduction in cost. Hoping that we see the same or better on switch.


dizzy2003:
Very cool, I look forward to trying it tomorrow.

huwb:
It appears to give up to a 2X speedup for the ocean render pass on switch so thats great.

I'm setting it to a 2x reduction by default, which should roughly quarter the geometry that is normally drawn, at very little visual impact in my tests.

PR #207, opened 2019-04-01
this fixes zillion if Underwater material instances creation

ShadowTeolog:
fixes new material instance creation every frame,
resulting slowdown and crash after few minutes.
material property surprising

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/OceanChunkRenderer.cs
+++ b/crest/Assets/Crest/Crest/Scripts/OceanChunkRenderer.cs
@@ -18,6 +18,7 @@ namespace Crest
         Mesh _mesh;
         Renderer _rend;
         MaterialPropertyBlock _mpb;
+        private Material _lastmaterial;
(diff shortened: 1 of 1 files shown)


huwb:
Damn! Thanks for this. I don't fully understand the issue, but I will merge it for now if it fixes it for you and experiment with it more later.

huwb:
Ok so simply reading the material property causes the material to be instanced, thats not good! I made a slight adjustment - I'm using sharedMaterial which should stop any instancing occurring and seems to work..

Thanks again for tracking this down and for the quick fix

ShadowTeolog:
I also think about using shared material property, but still not know all KREST resource workflow, shared material changes may affect on original material, so i use more safe way, to prevent changes on borrowed material to affect original material.
You know you project better, so can do this.

huwb:
Great point! I hope that it wont do that as I'm not changing the properties themselves, but i will test that later!

PR #209, opened 2019-04-02
Fix planar reflections search for several cameras on one display

ShadowTeolog:


huwb:
Thanks for posting this.

So camera instance ID will uniquely identify both different displays as well as several cameras in one display, is that right?

Are you doing split screen for your project by the way? I haven't tried out the ocean much in split screen, if they are two free views i guess it would required having two oceans and given that the ocean is a global singleton right now, that seems likely to require a bunch of work.

ShadowTeolog:
Yes, this code can handle any count cameras with planar reflection script,both on one or multiply displays, only restricted by performance, and I not sure about camera aspect handling in final reflection. For now two camera is limit for my scene, to keep near 60 fps, also i will try to add texture update not every frame, and split reflection update to different frames for different camera. Also currently any camera without reflection script(editor scene camera too) use black reflection ,but not shows wrong reflection map, this look not good. 
I already try multiple camera installation with one ocean instance. I may case all camera in one position or somethere nearby but has different direction and projection settings. This for multi-monitor visualization. Two oceans look impossible to handle by perfomance and too hard to handle multiply instance interraction inside ocean components. Some troubles still here:
 if one camera go underwater, surface on over camera stop render normally, but in my situation this not big deal, simular kind of camera will be grouped on one computer
Also has trouble with ocean teleport, ocean already has this function, but after all object teleport, wave form little changes, look like teleport missing something inside ocean.

huwb:
Regarding teleport - i added some notes in this commit - do these help?

34a95aa7bd0cfd7f09b19f477a76fa3f25837eee

PR #212, opened 2019-04-04
Planar reflection upgrade

ShadowTeolog:
Replacement for previous pull request.
Add render range limit for reflection(1000 by default)
Copy render path from original camera
Add reflection refresh frame interval(1(every frame)-by default)
Add oddity attribute for relax reflection refresh to different frames(if true, script work only on odd frames)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,18 +4,17 @@ Thank you for your interest in contributing to Crest! :)
 
 To propose changes/additions, please fork this repository and then open a pull request.
 
-Time and effort has been put into maximising quality of the Crest codebase. Below a number of stylistic conventions are outlined that aid in readability and understanding. The code has gone through many iterations and many refactoring passes to smooth out kinks and simplify complexity wherever possible. Unfortunately this volume of change may cause some pain for forks. Now that the existing code is maturing we will establish releases to provide less volatile branches which will hopefully help give stability.
-
 There are a few stylistic conventions followed in the code:
 
-* Member variables follow lower camel case with a single underscore prefix: \_exampleVariable
-* Single blank line between functions and between logically separate code
-* Public member variables typically grouped together above private ones
-* Comments placed on their own line above the relevant code
-* No commented code in the master branch, unless it is instructive to include it and not distracting for readers
-* No code that is not used in the master branch; experimental or unfinished code belongs in development branches
-* [Documentation comments](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/xmldoc/xml-documentation-comments) preferred when useful for documentation
+* The Format Document command in vanilla VS2017 is used to apply code formatting after every code edit (shortcut Ctrl+K, Ctrl+D)
 * Simplicity is king, with modularity favoured when it does not hurt readability or workflows
+* Member variables follow lower camel case with a single underscore prefix: \_exampleVariable
+* Single blank line between functions and between logically separate bits of code
+* Tweakable/serialised variables are private and tagged with *SerializeField*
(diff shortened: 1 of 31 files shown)


huwb:
Thanks for this. Apologies if I'm a bit slow to go through this, have been very busy at work. Sounds like good changes though and I'll let you know once I've had a chance to look at it.

huwb:
So i went through this and the changes look good.

There are some minor nitpick stuff, like variable naming conventions. i dont know if this is interesting or motivating for you to look at, i'd be happy to do a vanity pass to 'crest-ify' the code if you dont mind me doing so. i just did a review and spring clean of the coding standard page which covers the main points  https://github.com/huwb/crest-oceanrender/blob/master/CONTRIBUTING.md

Regarding refresh rate - a comment refers to RefreshFrameOffset but the final code uses a bool. I think have the offset makes more sense to me - i could specify that my reflection cams update 1 in every 3 frames, and then i could assign offsets of 0, 1, 2, etc to control which offsets happen when. Is there a reason you settled on a bool?

Regarding use of the volatile keyword, do you actively expect to be accessing the data from multiple threads or is it a 'good habit' type thing to include it, or is there a reason i'm missing? I'm mostly just curious.

Besides that these look like really good practical changes, thanks for contributing it back to Crest.

ShadowTeolog:
Good to listen that this changes useful for you. With code convention- I don't do this manually, this result of Resharper's default code convention with auto-completion hints. Resharper makes public member variables,different with private variables. It is better to keep the project in the same style, so code convention is good idea. About RefreshFrameOffset, initially this was int variable,but later I simply code to use only 2 cameras.
"volatile" useless now, but can be surprises in future versions of Unity with Update/Render multithreading.
Totally this result of code adoptation for our current needs, will also contribute back some changes related to server to clients sea shape sync. This my next week trouble:). Also I am found some gpu spikes while ocean surface calculation, this not my priority now, but currenlty don;t know why here used CommandBuffer on start of rendering(adding some GPU utilisation while GPU already full loaded) but not ComputeShader on update with no wait and receive result on next frame(eat GPU while Update process where GPU not loaded). Anyway this very good project this good adjustable code.

huwb:
Thanks @ShadowTeolog! I did many refactorings to try to keep code complexity under control so I'm glad to hear you're getting on well with it.

Yes good to hear your thoughts about the update! I was considering making it run async, but didnt have a chance to try it. So it would start running at the beginning of the graphics queue and would do the update while the GPU is doing the pre-z, shadows, opaque pass, and then put a fence to make sure its done before starting the transparent pass. What you describe sounds even better - the ocean update is fully prepared in LateUpdate (by BuildCommandBuffer), there might be a win to kick it off at that point so that it is running while the render thread is preparing the render. Would be great to try this out .. i'll let you know if i get any experiments done on my side.

Your contributions are great, please keep them coming.

huwb:
Ok i did some messing around, as you probably already know, the ocean command buffer is not suitable to go onto async compute as i'm not using compute shaders at the moment. I'm pretty sure it could all be moved to compute and probably be optimised a bunch while doing so due to group shared memory and stuff.. might take some time to move all the shaders across though.

moosichu:
I'm investigating moving some GPU operations to compute shaders @ShadowTeolog, see issue #214 if you are interested. 🙂

huwb:
Thought I'd bump this PR.

Regarding formatting do you plan to do a pass or do you want me to handle it? I'm happy with either.

What do you think about the issue i raised with the cameras? I currently feel we should not hardcode it for two cameras - for the crest version of the code.

ShadowTeolog:
Ok. I fix to camera hardcode to n camera in few days.

ShadowTeolog:
Fix frame render offset to int and cleanup

huwb:
I did some tests with 2 cameras and it looks great to me! Nice work and thanks for doing those changes and the cleanup. Merging!

PR #215, opened 2019-04-08
Fix wave stutter on 16bit GPUs

huwb:
Fixes issue where waves jitter after leaving running for some minutes
on platforms like iPhone.


huwb:
I'll merge this in as its had a bit of testing now.

@mingoddess

PR #224, opened 2019-04-16
Add reproducible cross-platform random System.Random implementation

ShadowTeolog:
Use local random source with cross-platform implementation, to allow reproduce same spectrum on server without Unity

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Helpers/MersenneTwisterRandom.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Helpers/MersenneTwisterRandom.cs
@@ -1,0 +1,286 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace WaveShapeCalculator
+{
+    using System;
+
+namespace Random
+{
+    /* C# Version Copyright (C) 2001 Akihilo Kramot (Takel).       */
+    /* C# porting from a C-program for MT19937, originaly coded by */
+    /* Takuji Nishimura, considering the suggestions by            */
+    /* Topher Cooper and Marc Rieffel in July-Aug. 1997.           */
+    /* This library is free software under the Artistic license:   */
+    /*                                                             */
+    /* You can find the original C-program at                      */
+    /* http://www.math.sci.hiroshima-u.ac.jp/~m-mat/MT/emt.html    */
+    /*                                                             */
+
+    /// <summary>
+    /// Implements a Mersenne Twister Random Number Generator. This class provides the same interface
+    /// as the standard System.Random number generator, plus some additional functions.
+    /// </summary>
+
+    public class MersenneTwister: System.Random
+    {
+        /* Period parameters */
+        private const int N = 624;
+        private const int M = 397;
+        private const uint MATRIX_A = 0x9908b0df; /* constant vector a */
+        private const uint UPPER_MASK = 0x80000000; /* most significant w-r bits */
+        private const uint LOWER_MASK = 0x7fffffff; /* least significant r bits */
+
+        /* Tempering parameters */
+        private const uint TEMPERING_MASK_B = 0x9d2c5680;
+        private const uint TEMPERING_MASK_C = 0xefc60000;
+
+        private static uint TEMPERING_SHIFT_U( uint y ) { return ( y >> 11 ); }
+        private static uint TEMPERING_SHIFT_S( uint y ) { return ( y << 7 ); }
+        private static uint TEMPERING_SHIFT_T( uint y ) { return ( y << 15 ); }
+        private static uint TEMPERING_SHIFT_L( uint y ) { return ( y >> 18 ); }
+
+        private uint[] mt = new uint[N]; /* the array for the state vector  */
+
+        private uint seed_;
+        private short mti;
+
+        private static uint[] mag01 = { 0x0, MATRIX_A };
+
+        /// <summary>
+        /// Create a twister with the specified seed. All sequences started with the same seed will contain
+        /// the same random numbers in the same order.
+        /// </summary>
+        /// <param name="seed">The seed with which to start the twister.</param>
+
+        public MersenneTwister( uint seed )
+        {
+            Seed = seed;
+        }
+
+
+        /// <summary>
+        /// Create a twister seeded from the system clock to make it as random as possible.
+        /// </summary>
+
+        public MersenneTwister()
+            : this( ( (uint) DateTime.Now.Ticks ) )  // A random initial seed is used.
+        {
+        }
+
+
+        /// <summary>
+        /// The seed that was used to start the random number generator.
+        /// Setting the seed resets the random number generator with the new seed.
+        /// All sequences started with the same seed will contain the same random numbers in the same order.
+        /// </summary>
+
+        public uint Seed
+        {
+            set
+            {
+                seed_ = value;
+
+                /* setting initial seeds to mt[N] using         */
+                /* the generator Line 25 of Table 1 in          */
+                /* [KNUTH 1981, The Art of Computer Programming */
+                /*    Vol. 2 (2nd Ed.), pp102]                  */
+
+                mt[0] = seed_ & 0xffffffffU;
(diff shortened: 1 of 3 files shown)


huwb:
Thanks @ShadowTeolog, great contribution again. I had no idea System.Random was not deterministic across platforms - thats a bit crap.

I can't take the commented out warning (it indicates that waves are being lost, and can result in pops when the viewpoint moves and they come back) but i'll take the rest. I'm fairly sure I can add commits to this, I'll see if i can remember how!

ShadowTeolog:
yes. This commented warning passed to pull request by mistake. Also I update this part later, because look like ocean state look not reproducible after waver change and move, look like need to keep original state at zero position for basis for predictable state at point generation.

moosichu:
It's also probably worth checking that the Mersenne twister license is compatible with the project's MIT license and that we appropriately handle it. Going to do some research into that now.

ShadowTeolog:
Look like artistic license compatible to MIT, GPL, etc. Only require to any changes of original code mast not be hidden from pulling to original version

moosichu:
OK, so this version of the file appears to either come from this project:
https://github.com/stratisproject/HashLib/blob/master/TomanuExtensions/Utils/MersenneTwister.cs
or this one:
https://github.com/stratisproject/NStratis/blob/master/Hashing/TomanuExtensions/Utils/MersenneTwister.cs

Which seems to be an adaption from this implementation:
https://github.com/vpmedia/template-unity/blob/master/Framework/Assets/Frameworks/URandom/MersenneTwister.cs

All of the above reference the original C program from which they are adapted: http://www.math.sci.hiroshima-u.ac.jp/~m-mat/MT/MT2002/CODES/mt19937ar.c, but only the latter includes the original license. 

So the C# port appears to *originally, orginally* come from this now dead page [http://www.centerspace.net/resources/free-stuff/mersenne-twister](https://web.archive.org/web/20140909002055/http://www.centerspace.net/resources/free-stuff/mersenne-twister).

moosichu:
So, part of the code appears to be licensed under [the Artistic license](https://en.wikipedia.org/wiki/Artistic_License) (see also [here](https://opensource.org/licenses/Artistic-1.0)). However, I couldn't find any evidence of which version of the license it was originally distributed under.

The original C program has this license (which applies to the C# program as well):

```
A C-program for MT19937, with initialization improved 2002/1/26.
   Coded by Takuji Nishimura and Makoto Matsumoto.

   Before using, initialize the state by using init_genrand(seed)  
   or init_by_array(init_key, key_length).

   Copyright (C) 1997 - 2002, Makoto Matsumoto and Takuji Nishimura,
   All rights reserved.                          

   Redistribution and use in source and binary forms, with or without
   modification, are permitted provided that the following conditions
   are met:

     1. Redistributions of source code must retain the above copyright
        notice, this list of conditions and the following disclaimer.

     2. Redistributions in binary form must reproduce the above copyright
        notice, this list of conditions and the following disclaimer in the
        documentation and/or other materials provided with the distribution.

     3. The names of its contributors may not be used to endorse or promote 
        products derived from this software without specific prior written 
        permission.

   THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
   "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
   LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
   A PARTICULAR PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
   CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
   EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
   PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
   PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
   LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
   NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
   SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.


   Any feedback is very welcome.
   http://www.math.sci.hiroshima-u.ac.jp/~m-mat/MT/emt.html
   email: m-mat @ math.sci.hiroshima-u.ac.jp (remove space)
```

ShadowTeolog:
So, trouble will fixed if original "bla-bla-bla bsd" code will added back to currently exists file?
At least half of mathematics applications + boost, office, etc use code based on original C implementation and I don't think what this impossible

huwb:
Let's just make sure we credit the source with that link you have
currently. The notice is not easily available as you've shown Tom.

huwb:
> Also I update this part later, because look like ocean state look not reproducible after waver change and move, need to keep original state at zero position for basis for predictable state for setorigin recalulation

I havent fully understood - do you plan to make further changes or is this PR ready to be merged?

ShadowTeolog:
This look not simple, initial data contains seed,wind speed and direction end base offset position. This results some changes. Also need change in offset origin code. Not sure then ca do this immediately/

huwb:
I don't really get how setorigin is affecting this code - my impression from what you wrote above is that the seed is dependant on the position/base offset, in which case I would understand why this would break when these move... however in the code the seed is set by the user, which should always be consistent.

There are some comments in FloatingOrigin.cs about requirements to get pop-free setorigin() and i wonder if you're seeing the normal map popping.

I don't mind leaving this open in any case, no pressure from my side!

ShadowTeolog:
Not this code, here only one part related with random reproduction, over parts has more impact to different files, and I not pull them before finish and clean. Trouble is copy ocean shape starting in middle of simulation, with any wind and any time moment. This require "Update conditions" function affecting different code parts. Also current code make origin changes from current position but not initial position. This result impossible to return to original state in cycle:
move->change waver-> move back. Final result depending on position there waver changed. This make result not matched with server. But simple solution - rebuild in zero position and offset again after any waver change results huge ocean shape move after small wind change.
So now I don't see ideal solution here and not pull something related, only small peace of code, useful and harmless.

huwb:
I'm not confident i understand still, but this feels like it should be solvable.

We had the ability to render the gerstner waves from geometry laid out in the world. As a concrete example if you want a big storm in the middle of the open ocean, it is possible to render a circular patch of geometry with whatever wave condition you want there:

image

The numbers represent vertex weights to blend out the stormy waves smoothly at the edges.

This would move around with the rest of the geometry in the world and should support the teleport and should also be 100% deterministic and consistent throughout the world. We can discuss bringing this mechanism to life if it its useful.

In any case please let me know if/when you think this PR is ready to merge, as its not clear to me, and I'll leave it open in the meantime.

moosichu:
https://docs.unity3d.com/Packages/com.unity.mathematics@1.1/api/Unity.Mathematics.Random.html?q=random

This is a thing now, so will look at this as an alternative :)

ShadowTeolog:
> This is a thing now, so will look at this as an alternative :)
May be yes,or may be not. No equal implementation exist for this API to embed in normal .NET server application. This only option for Unity based server.

13E12K:
Just considering to add a good water system to a multiplayer project, and I came across this precious site.

Regarding syncing different clients on network, I would like to ask the logic behind. I don't know the crest script itself, but is not this kind of ocean re-creations a result of some Fourier transformations, which creates a real time simulation from a frequency based distribution? In that case, form different wavelengths and frequencies into the time domain?

If so, I could not get the logic how syncing the time would help to solve this issue. Even with the same parameters and same time, an infinite random outcomes are possible. Or is not it the case for Crest?

ShadowTeolog:
if you have a reproducible source of random numbers that works the same on all clients and the server, as well as synchronized time, the waves will play exactly the same. To do this, it is enough to set the initial spectrum at time 0, the local point as a base and the source of random numbers. This is tested on a multi-display system with several computers displaying side-by-side images with a full angle of 270 degrees.

PR #225, opened 2019-04-16
Light scattering enhancements

huwb:
Trying to incorporate some of the light scattering stuff used by rare for SoT. i think they are using the size of the horizontal displacement to drive the scattering intensity, and perhaps only for a particular scale of waves. I've implemented a variant of this where the scattering intensity is computed per wave octave and divided by wavelength to normalize which seems to hold up better in different wave conditions.

The effect is controlled in the **Subsurface scattering** section of the material, with the *Base Mul* param scaling this contribution (and the *Colour* setting the colour).

image


VizzzU:
Cool stuff , the lighting is very smooth now and noticeably  more realistic but i have strange results underwater . The fog becomes dark blueish whereas the ocean undersurface gets clear green . Even with the copy parameters each frames checked . 
If everything is good on your side i have certainely forgotten a commit so don't care  .  (and i am using the godrays stuff dont know if it can interct with this )

huwb:
Another iteration, got a lot of the weirdness when camera changing altitude fixed. Reallllly want this to work.

Theres still a subtle pop :/, and i haven't looked at underwater yet.

image

VizzzU:
You are doing an amazing job !

moosichu:
A reference as this is soon to be merged: 😄 
Before:
image

After:
image

huwb:
I must have left this open? These changes were merged in in #284. Closing.

PR #227, opened 2019-04-26
Abstract materials and mpbs to make CS port easier

moosichu:
In order to make the transition to Compute Shaders as smooth as possible, we are staging the process so that they will initially be implemented with the same structure (and alongside) the existing shaders. As compute shaders do not use Materials for parameter initialisation, we need to roll our own system which can share an interface with the same abstraction that the existing materials use.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgrShadow.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgrShadow.cs
@@ -25,6 +25,14 @@ namespace Crest
         RenderTexture[] _sources;
         PropertyWrapperMaterial[] _renderMaterial;
 
+        static int sp_CenterPos = Shader.PropertyToID("_CenterPos");
+        static int sp_Scale = Shader.PropertyToID("_Scale");
+        static int sp_CamPos = Shader.PropertyToID("_CamPos");
+        static int sp_CamForward = Shader.PropertyToID("_CamForward");
+        static int sp_JitterDiameters_CurrentFrameWeights = Shader.PropertyToID("_JitterDiameters_CurrentFrameWeights");
+        static int sp_MainCameraProjectionMatrix = Shader.PropertyToID("_MainCameraProjectionMatrix");
+        static int sp_SimDeltaTime = Shader.PropertyToID("_SimDeltaTime");
+
         SimSettingsShadow Settings { get { return OceanRenderer.Instance._simSettingsShadow; } }
         public override void UseSettings(SimSettingsBase settings) { OceanRenderer.Instance._simSettingsShadow = settings as SimSettingsShadow; }
         public override SimSettingsBase CreateDefaultSettings()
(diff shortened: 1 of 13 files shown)


crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgrShadow.cs
```diff
@@ -184,29 +184,27 @@ public override void UpdateLodData()
...
 
                 lt._renderData.Validate(0, this);
-                _renderMaterial[lodIdx].SetVector("_CenterPos", lt._renderData._posSnapped);
-                _renderMaterial[lodIdx].SetVector("_Scale", lt.transform.lossyScale);
-                _renderMaterial[lodIdx].SetVector("_CamPos", OceanRenderer.Instance.Viewpoint.position);
-                _renderMaterial[lodIdx].SetVector("_CamForward", OceanRenderer.Instance.Viewpoint.forward);
-                _renderMaterial[lodIdx].SetVector("_JitterDiameters_CurrentFrameWeights", new Vector4(Settings._jitterDiameterSoft, Settings._jitterDiameterHard, Settings._currentFrameWeightSoft, Settings._currentFrameWeightHard));
-                _renderMaterial[lodIdx].SetMatrix("_MainCameraProjectionMatrix", _cameraMain.projectionMatrix * _cameraMain.worldToCameraMatrix);
-                _renderMaterial[lodIdx].SetFloat("_SimDeltaTime", Time.deltaTime);
+                _renderMaterial[lodIdx].SetVector(Shader.PropertyToID("_CenterPos"), lt._renderData._posSnapped);
```
moosichu: @huwb This was the bit that I was most unsure about, because on one hand forcing ourselves to do this explictly is a good reminder that we should be creating property IDs for all shader params up-front. However, it is a bit clunky when done inline like this. I'm happy to implement this conversion inside the property wrapper if you think that is preferable.

huwb: Thanks Tom. This was lazyness on my part - this class should store the IDs as member variables. Also clunky but that's the system. I'll see if I can point to an example..

huwb: Didn't find a perfect simple example. But as I said it could store the IDs as static int member variables and initialise them on construction or init. Let me know if that makes sense.

moosichu: 👍 Will do that now

moosichu:
I have changed the structs to classes, but reading this:

https://stackoverflow.com/questions/63671/is-it-safe-for-structs-to-implement-interfaces

You can avoid boxing by using generics, although that then does produce more code and is more complex, so I've gone with the simple solution. Completely forgot about boxing! Good point, and it would have led to potentially weird bugs.

moosichu:
Sadly it's no longer more deletions than additions 😢 

But it is a bit of technical debt out of the way 😄

huwb:
Looks good! It touches many of the systems so we'll need to test most of the functionality. Could you let me know what tests you have done/plan to do and I'll fill in any gaps as needed (i'll also do a very quick sanity check test as well)? Thanks!

moosichu:
I was thinking about testing the other day actually, for now I can confirm that all the example scenes still work. I've also tested shadowing as well.

Do you think setting up automated tests would be useful or introduce too much friction? Unfortunately, all the free CI systems don't support GPUs for testing, (except for maybe Azure Pipelines https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops ). So if we set tests up, we would have to do it locally. Not that I'm even sure what these tests would look like, (potentially only regression tests for bugs that get reported?).

I did get Unity testing setup for CPU stuff a few years back:

https://github.com/moosichu/part-ii-project

https://travis-ci.com/moosichu/Part-II-Project (it's old now so no longer works, and all the history has been wiped).

PR #228, opened 2019-04-27
Null check for uninitialized SamplingData tag.

jabza:
Ran into this null reference when sampling data immediately. Not sure what was the cause, but this check makes sure the SamplingData._tag attribute is set, and fails gracefully otherwise.


huwb:
Thanks, oddly enough this null check is omitted on purpose, _tag should never be null. How are you creating the sampling data?

(I will add an assert to make it clearer that something has gone wrong)

huwb:
Closing as will not merge.

PR #238, opened 2019-05-04
Experiment - focus waves in towards point

huwb:


VizzzU:
Ahaha you are a bot or an alien , btw the result is incredible !  exactly what i expected  visually . Wow

moosichu:
If anyone is interested, this would be to simulate the fact that waves always appear to approach islands from every angle, independent of the overall directions of the waves themselves. (from discussion on Discord, found by ceebee):

https://www.scubageek.com/articles/wwwisland.html

huwb:
I pushed this - added it to gerstner script as an option (default to off). When turned on, two gizmos appear in the editor that indicate the region of influence of this behaviour (white full strength, black 0 strength).

image

This is now turned on in main.unity.

PR #246, opened 2019-05-13
Store depths simply instead of relative to a baseline

huwb:
The depths simply represent water depth (not relative ot some deep baseline)  and now the depths can be visualised/debugged directly

image

This makes the code simpler/better and the system less complex.


huwb:
@KellanHiggins heads up this will mean the depth value represents simply depth of water, no shenanigans with the -1000m baseline. so 2 means water is 2 meters deep.

PR #254, opened 2019-05-17
Improvements and fixes for fog, shadowing and lighting

jjxtra:
Reduced dir light scatter and specular by taking shadow to third power.

Added two external shader hooks:
- Fog: In Ocean.shader, you can wire up a fog function by turning on a #define and including your own shader file to provide the fog function
- Shadows: In UpdateShadows.shader, you can wire up a shadow function by turning on a #define and including your own shader file to provide the shadow function


huwb:
Thanks! yeah I think I can use the hooks. Could you illustrate why you powered up the shadow values?

There are sharp and soft shadows and these are configurable in case this helps.

jjxtra:
No shadow:

image

Default shadow with a large cube. In my opinion, too much scattering and specular.

image

Power of 3 shadow. Eliminates the scattering and specular.

image

huwb:
If possible could you please push the cube so I can the same scene here?

jjxtra:
It is done

jjxtra:
As a side note, how does one disable the shadow fade in the far distance? Even setting fade to 0 in the shadow shader does not do it.

huwb:
Thanks again for this.

I'll take the hooks but not the powered up shadows. I don't understand still what it is solving for, and it makes a strong edge in the scattering. There are shadow settings to make the transition stronger if desired.

image

The shadows are sampled from the shadow maps and stored in the LODs. The lods have limited range and data is not available in the distance. The shadow cascades are also limited in extent for performance reasons, so the data is not available at all after a certain range. We could look at making the default value that it fades to configurable in the settings, so e.g. it fades to 50% shadowed.

Let me know what you think. The next step might be to reduce this PR to just the hooks if you agree.

jjxtra:
Ok, I can rollback the power of 3 shadows.

jjxtra:
Looks like the edges still have some shadowing issues. As I rotate the camera, especially rapidly, I notice fading shadows in the water that comes in view. Is this by design? It takes the shadows about 1/2 second to fade to the correct value.

image

jjxtra:
In this shot the shadow edge artifacts is more pronounced. Also now that I have removed my power of 3 shadow code, I see the specular of the sun through the storm clouds, it is fainter but still very noticable.

image

huwb:
It's a TAA-style filter, _currentFrameWeightHard and _currentFrameWeightSoft in the settings can help to speed it up. There might be some way i can speed up the filter myself if the data is unitialised or soemthing. ill think about it

huwb:
`I see the specular of the sun through the storm clouds`

I have no idea why this would happen. I'm verifying here with a very bright sun and it blocking the spec. Is it possible that the storm clouds are not 100% shadowing?

jjxtra:
The storm clouds return a shadow based on shadow strength of dir light. In my case, I can replace the shadow shader with a value of 0.1 and still see the specular. In my opinion, a shadow of 0.1, very heavy shadow, should block all specular. Even a light shadow, say 0.5 or 0.8 should block the specular.

huwb:
Ah that will be it. I'm not doing anything fancy with this value - i'm just multiplying the light specular response by it the way shadowing would normally be applied.

jjxtra:
So I could return 0 for my storm clouds but this causes VERY dark shadows over land and looks really bad.

huwb:
I think clouds block a very high percentage of the directly specular light, and scatter it into indirect light. i don't know how this should integrate into unity to get a correct looking response on surfaces.

jjxtra:
Storm clouds yes, but even regular scattered clouds should block all the specular if the sun goes behind the clouds. Or even a thin piece of posterboard will block the specular if placed above the water in direct view of the sun. The shadow it casts is not super dark but is enough to block the specular.

huwb:
The value in crest needs to stay as is - it has a physical, standard meaning - percentage of the direct specular light blocked. Remapping it or etc is not an option. Sorry i cant be of more help here.

As a suggestion, I guess you're adding a hook, if you need a high proportion of the direct light blocked could you do it in the hook? Hopefully this can be solution.

jjxtra:
Shadow changes rolled back, I did add planar reflection, shadows, shadow cube and Unity sky to the main scene, I think it is helpful for debugging and seeing everything configured to the max.

huwb:
Thanks @jjxtra, i'll discuss the data changes with a colleague and get it merged in, nice one.

huwb:
This is a long shot but ill ask anyway - is EXTERNAL_FOG_PASS_FUNC a standard name that other sky assets are using. Would be useful if there was a standard. Otherwise the hooks in ocean.shader only really make sense for your asset, rather than being useful for the general case - if i'm understanding correctly? Apologies if not, i'm quite tired.

jjxtra:
It's just a define I came up with. Any other asset using fog with depth buffer would do it the same way.

huwb:
Ok. I'm cringing while writing this .. sorry i could not have given this feedback from the beginning, i'm terrible at reading/reviewing code .. but if the code does not suit the general case, i can't take them into crest.

Azure has documentation on how to add it to existing shaders. Could you add a section for crest which instructs what code to add? I can improve the comments around the insertion point with something custom if you want, like a reference to your asset.

(sorry)

huwb:
Wait actually i think i can refactor this into something that doesnt make any assumption on the sky asset code, but still gives the user a leg up. I will follow up with a proposal shortly!

jjxtra:
I am going to close this and keep referring my users here, but will keep an eye on your solution.

huwb:
Ok sure.. On further consideration today I decided my idea would water this down and would be a middle ground that doesn't really help anyone.

One thing I noted in this pr, I believe the insertion point needs to be where the comment is, otherwise it's double fogging. 

I'll merge on the UV fix separately.

jjxtra:
Switched it to be where the comment is

PR #268, opened 2019-06-08
Use depth caches in gerstner jobs (attempt 2)

huwb:
From @KellanHiggins  :

To get the wave height at any point, you add a monobehavior called `CrestWaterSampler` to your base object. This script has an array of **local** Vector3s and a array of floats with the resulting heights at those points. Every Update(), the script will ask the static gerster jobs to update the list of heights. At any time you can use this list of floats to feed into whatever other scripts need them. Remember to transfer them from local to world. The boat floating script does not use this, instead it calls on the gester jobs directly.

Outline of registering
1. Register your sampler with the jobs system
2. It adds all the local points to the jobs system
3. Now the job system will update those every round.

Outline of Job
1. Convert local positions to world positions via job
2. Process all points via job

Caching Depths
Whenever a depth cache is added to the world, it is automatically added to the gester jobs. Then during the calculation phase, it finds the shallowest depth and tells the job processing the heights what depth the point is at. I ripped the code directly from the shader, so it should work almost the same. There is some blending that doesn't happen though. Like when the depth cache's edge is at 200m deep and then it goes right to 1000m deep. If there are two overlapping caches, it finds the shallowest point and uses that.

To write your own sampler, you just have to feed in a list of local points relative to a transform and you will get a list of heights back.

Changes
- The biggest change is now you input local positions relative to a monobehaviour transform. Since most of the time you need this attached to a boat, I thought it made sense. Not sure the use case where you wanna grab a world coordinate.

Possible TODO
-The wave heights are still slightly off. I think there has to be interpolation run instead of just finding the closest point. But should work for all but the smallest boats in the biggest waves as is.


KellanHiggins:
I have resolved the conflicts. I just overwrote everything with my updates since it was almost exactly the same. Next step will be to outline here what I have changed.

KellanHiggins:
To get the wave height at any point, you add a monobehavior called `CrestWaterSampler` to your base object. This script has an array of **local** Vector3s and a array of floats with the resulting heights at those points. Every Update(), the script will ask the static gerster jobs to update the list of heights. At any time you can use this list of floats to feed into whatever other scripts need them. Remember to transfer them from local to world. The boat floating script does not use this, instead it calls on the gester jobs directly.

Outline of registering
1. Register your sampler with the jobs system
2. It adds all the local points to the jobs system
3. Now the job system will update those every round.

Outline of Job
1. Convert local positions to world positions via job
2. Process all points via job

Caching Depths
Whenever a depth cache is added to the world, it is automatically added to the gester jobs. Then during the calculation phase, it finds the shallowest depth and tells the job processing the heights what depth the point is at. I ripped the code directly from the shader, so it should work almost the same. There is some blending that doesn't happen though. Like when the depth cache's edge is at 200m deep and then it goes right to 1000m deep. If there are two overlapping caches, it finds the shallowest point and uses that.

To write your own sampler, you just have to feed in a list of local points relative to a transform and you will get a list of heights back.

Changes
- The biggest change is now you input local positions relative to a monobehaviour transform. Since most of the time you need this attached to a boat, I thought it made sense. Not sure the use case where you wanna grab a world coordinate.

Possible TODO
-The wave heights are still slightly off. I think there has to be interpolation run instead of just finding the closest point. But should work for all but the smallest boats in the biggest waves as is.

huwb:
Sorry its taken me a while to look at this. Thanks for posting this.

I looked through the changes. I understand now that the ocean depths are computed in MatrixTransformJob and I think i understand what the code is doing.

Could you talk more about the use case for CrestWaterSampler ? I'm currently a little unsure of the utility of this. As you've pointed out the boats can go straight to GerstnerHeightJobs . Is there a scenario you had in mind where something would want to send a bunch of local-space query points off to be queried? In my experience theres a few things which have batches of query points, and then lots of misc things in world space like cameras,projectiles, gameplay queries, etc, but i'm curious if you had something in mind.

The depth cache is a good fix, but we need as sure as possible that it is correct. I spent some hours getting it to be 100% accurate, i received bugs whenever something was even slightly off - i thought it was good enough many times but it was one of those things that needed to be perfect. Is drilling down further into the possible TODO above? I wish to help but i dont have bandwidth in the short term.

A final trivial point - we'll want to remove any unnecessary data changes and white space changes. That's not critical for now. I can help with this where needed - i know how fiddly whitespace is :/ currently crest uses the formatting of VS with default settings, but i know people use different editors, and its a pain...

moosichu:
I've created PR #271 which should hopefully help any formatting.

KellanHiggins:
@huwb The use case for the Crest Water Sampler is I need it ;). I have boats that need about 100 - 150 points every frame and they use this sampler to find the query heights. I may have 10 to 15 boats and need the local heights to do the floating calculations. For any external programs or floating systems, they can use this instead of the other more integrated boat probe systems. 

Best way to test if the depth cache is to turn on the sampler and create a 64 by 64 sample grid (options in there) and then put in a bunch of depth caches and look to make sure they are right. I did a bunch and they all seemed fine. The heights can sometimes be a bit off, but this is not the depth cache, rather the base calculations (I think it needs interpolation). The only problematic ones was when the depth cache went from lets say 500m deep right to 1000m (hard edge). The heights didn't blend like they do in your shader. I am not sure how you blend these though.

Glad @moosichu can fix my heavy hand on those enter and white spaces. What do you mean by data changes?

huwb:
Thanks. To explain a bit more where i'm coming from, I need to minimize code I'm adding to crest, every line adds bug fixing cost, maintenance cost, complexity, compile time, etc.

I'm not sure i understood still, why do you not just go directly to GerstnerHeightJobs, and instead use a middle man? Of course, I don't want to critique your code or argue against you using it :), I'm making sure every line of code that goes into crest has earned its place. Alternatively you could keep this class locally in your project, and we're hoping to set up a wiki where people can share scripts which could be another good way of disseminating this to people that want it.

Ok then i think the next step with the depth caches is to do some testing. I'll look out for the case you mention regarding discontinuities. Unfortunately i have a lot on my plate with some big things currently, but these are starting to land. I'll post back when i make progress.

To apply the formatting, someone needs to open each source file and press Ctrl K Ctrl D if using VS, or whatever the "Format Document" command is in the editor (the editor needs to support editorconfig files, which i think a bunch do). Not a very exciting task but I'd be happy to do it.

Finally, data changes - if you look at the "Files Changed" view above there are changes to misc things like materials which should be reverted (or justified for merging into crest). Also there's a scene for burst jobs that i think we should just replace rather than adding a second scene, but i can look at this later (i would probably remove mine and keep yours after doing a few checks).

huwb:
Closing for similar reason we closed the original gerstner job branch

PR #271, opened 2019-06-14
Add editorconfig to project

moosichu:


huwb:
Thanks tom. I will grab this later and do the format document command in VS and see what happens!

PR #272, opened 2019-06-16
Replace individual textures with texture arrays for LODs

moosichu:
# What problem does this change solve?

Draw calls can have a fairly large overhead, and Crest makes a seperate one for each LOD that is calculated. Changing the LOD textures from individual textures to texture arrays will allow us to use geometry shaders to render to the LODs in a single draw call, reducing this overhead. This is demonstrated in the OceanDepthCache in this pull request.

This change also has the benefit of making the code in shaders a bit clearer as LODs are explicitly indexed.

# What other changes does this PR include?

  - An example of a geometry shader being used to target different slices of a texture array is demonstrated in the OceanDepthCache.

  - The persistent LODs (foam and dynamic waves) are now made in single compute shader dispatches.

  - Shadowing is now done using compute shaders, but not a single dispatch.

# What are potential downsides to this change?

## (Older mobile) devices without compute shader support will not longer be able to run Crest

We couldn't avoid using compute shaders, because the ShapeCombine shader now writes and reads from the same texture array in order to combine the LODs together, something that can only be done in compute shaders.

However because of this, we have been able to bite the bullet and port some of the LodDataMgrs to compute shaders as well.

## Shaders are fairly heavily modified

Whilst we have done our best to ensure the change is as non-intrusive as possible, it is a fairly big change, which is why we want to merge it into master before taking advantage of all the benefits it could bring.

# What are the next steps once this is merged?

  - Ensure as many LODs as possible use a single draw call/dispatch. 
  - Move compute shaders to async command buffers.


moosichu:
As this is such a big change, testing from anyone would be really appreciated! I've been able to test in-editor in Unity 2018.4 and 2019.1, and will also be able to test on Android.

moosichu:
I'm also happy to try and help anyone who is having difficulty integrating this change into their projects.

moosichu:
Geometry shaders aren't supported on metal, so I've made a PR which should allow this change to work on Apple devices (#273).

So please test that branch if you are testing on an Apple device. I cannot do so as I do not have an iOS device or Mac.

moosichu:
I believe this change is now ready to be merged. 😸

huwb:
BOOM

PR #273, opened 2019-06-17
Make geometry ocean depth cache shader optional

moosichu:
Geometry shaders aren't supported on some platforms, so this
demonstrates that feature for those platforms where they are definetely
not supported.


moosichu:
Based in in-person conversation with @huwb, going to merge this change in.

PR #276, opened 2019-06-23
Experiment with using the async compute queue

moosichu:
This branch and PR are for allowing people to expriment with moving compute tasks to the async compute queue. We will need to properly profile the branch to analyze what the performance gains are (if any). If all goes well these things will eventually be merged into master.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgrShadow.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgrShadow.cs
@@ -155,7 +155,7 @@ namespace Crest
             {
                 _bufCopyShadowMap = new CommandBuffer();
                 _bufCopyShadowMap.name = "Shadow data";
-                _mainLight.AddCommandBuffer(LightEvent.BeforeScreenspaceMask, _bufCopyShadowMap);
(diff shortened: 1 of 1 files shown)


moosichu:
Reading the Unity documentation, we don't need to check if async compute queues are supported because:
https://docs.unity3d.com/ScriptReference/Graphics.ExecuteCommandBufferAsync.html
```
If the target platform does not support async compute queues then the work is dispatched on the graphics queue.
```
However, apparently:
https://docs.unity3d.com/ScriptReference/SystemInfo-supportsAsyncCompute.html
```
Note that asynchronous compute queues are only supported on PS4.
```
This may be out of date, but we will need to do some more investigation, if this is true the benefits of async compute queues are likely to be limited.

moosichu:
Yes, so it turns out that the Unity documentation is out of date, reading this here:
https://forum.unity.com/threads/async-compute-support-for-more-platforms.583540/
However, it is only supported on DX 12:
```
 It definitely is DX12 only though. You can call the async APIs on DX11, but they'll fall back to just using the graphics queue. 
```
DX12 support is still classed as experimental in Unity 2019.1, cannot find information on it becoming stable in a given Unity release yet.

moosichu:
`SystemInfo.supportsAsyncCompute` returns false for me in DX12 in Unity, but then, *sighs* I get other log messages telling me that things are running asynchronously seperately. So, who knows what is going on there.

PR #278, opened 2019-06-25
Refactor/lod hierarchy

huwb:
This is not high priority but would be good to merge in when we're happy master is settled.

I've been meaning to simplify the ocean hierarchy for a long time now, finally got around to doing it. i've removed the intermediate LODx from the hierarchy, which will make it cheaper to update, and the lod scales are now computed instead of being read from transform.lossyScale. I would wager that the unicorn bug where gaps appear in the surface comes from error in the lossy scale.

Oh i couldnt resist running the Format Document command on OceanBuilder which made a LOT of changes. This PR is best reviewed per-commit, rather than Files Changed.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/OceanBuilder.cs
+++ b/crest/Assets/Crest/Crest/Scripts/OceanBuilder.cs
@@ -211,7 +211,7 @@ namespace Crest
             for (int i = 0; i < ocean.transform.childCount; i++)
             {
                 var child = ocean.transform.GetChild(i);
-                if (child.name.StartsWith("LOD"))
+                if (child.name.StartsWith("Tile_L"))
                 {
                     child.parent = null;
                     Object.Destroy(child.gameObject);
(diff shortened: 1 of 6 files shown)


crest/Assets/Crest/Crest/Scripts/OceanBuilder.cs
```diff
@@ -211,24 +211,22 @@ public static void GenerateMesh(OceanRenderer ocean, int lodDataResolution, int
...
             {
                 bool biggestLOD = i == lodCount - 1;
-                GameObject nextLod = CreateLOD(ocean, i, lodCount, biggestLOD, meshInsts, lodDataResolution, geoDownSampleFactor, oceanLayer);
-                nextLod.transform.parent = ocean.transform;
+                //nextLod.transform.parent = ocean.transform;
 
                 // scale only horizontally, otherwise culling bounding box will be scaled up in y
-                float horizScale = Mathf.Pow( 2f, (float)(i + startLevel) );
-                nextLod.transform.localScale = new Vector3( horizScale, 1f, horizScale );
+                //nextLod.transform.localScale = new Vector3(horizScale, 1f, horizScale);
```
moosichu: Is there any particular reason why this is being commented-out instead of being deleted?

moosichu: There's another similar line above this section as well which is commmented out.

PR #281, opened 2019-07-01
Refactor LodTransform Behaviour

moosichu:
The goal of this change is to simplify the run-time behavior of Crest, reducing the amount of work it has to do and remove dependencies on compile-time hardcoded values.

This is done with two changes:
- Having LodTransform cache data that is often sent to the GPU, so that parameters can be more easily bound in LodDataMgrs. 
- Replacing hardcoded fixed-length arrays with compute buffers in shaders, increasing their flexibility and reducing dependencies on `MAX_LOD_COUNT`. 

Garbage that was generated by the depth manager has been eliminated as well.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Helpers/PropertyWrapper.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Helpers/PropertyWrapper.cs
@@ -16,6 +16,7 @@ namespace Crest
         void SetFloatArray(int param, float[] value);
         void SetVector(int param, Vector4 value);
         void SetVectorArray(int param, Vector4[] value);
+        void SetBuffer(int param, ComputeBuffer value);
         void SetTexture(int param, Texture value);
         void SetMatrix(int param, Matrix4x4 matrix);
         void SetInt(int param, int value);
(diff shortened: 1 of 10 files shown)


moosichu:
NOTE: I'm not expecting this to get merged, this is more the result of me playing around with a few ideas and seeing if they were interesting. This change *could* be merged but I would want to clean it up a tad-more before than as not too happy with some of the variable names (although suggestions more than welcome).

moosichu:
I've also updated my editor to run the `.editorconfig` format on save, but I've put all of the formatting changes in separate commits.

huwb:
I think it would be good to talk this through. I have some half baked thoughts around making the lods less uniform in resolution/size/count and making it so that e.g. the ripple sim only has a few smaller lods and not the larger ones, and other stuff like this. another example - it might be that the foam needs less resolution than the displacements. the bindata stuff has lingered where it is master as i'm on the fence about whether it should be unified or not. i think talking it through would help my thoughts converge.

That all feels separate from the compute buffers part of these changes, which i dont really have a strong handle on. i'd look to you to give your thoughts about pros cons and whether the change is suitable, etc etc. if you're not sure about these things either then we could talk it through to help my brain along.

splitting out logically separate changes into separate PRs might help divide and conquer here, if this is possible/sensible.

huwb:
I think we should close for now. I think we did discuss this in person since my last comment above. To restate, i do think we'll want to vary parameters on a per-simulation basis at some point, so probably worth leaving the structure as is. However let me know if im closing prematurely.

PR #282, opened 2019-07-02
Fix shadows on Vulkan platforms

moosichu:
HACK INCOMING:
For some reason we get the following error on platforms which target Vulkan:

```
Shader error in 'UpdateShadow.compute': Unknown parameter type (54) for _LightShadowData at kernel UpdateShadow
```

We skirt around this by defining our own _LightShadowData as a float4, and
then making sure that the *real* _LightShadowData variabel is knocked-out.


moosichu:
Was approved by @huwb on Discord.
For the record, this change does make me v. unhappy.

But also delighted that all features now work on Vulkan.

moosichu:
Fixes #283

PR #284, opened 2019-07-02
Implement improved height-independent subsurface scattering

huwb:
This is a variation of the sea of thieves light scattering heuristic.

It also takes the logic that ensures large wavelengths are smoothly blended across lods and pushes it into the lod data anim waves, so that other large wavelength shape inputs can benefit from teh same logic, and it removes this complexity from the gerstner script, which now uses this common api to submit gerstner batches.


huwb:
thanks, it doesnt seem to be related to this change, it seems that reading shadows in the underwater stuff makes it flicker like crazy. for now i've knocked out the shadowing term when underwater.

PR #293, opened 2019-07-11
Add Pirate Cove and Critter Cove to Showcase Gallery

moosichu:


moosichu:
You can see the outcome here:

https://github.com/crest-ocean/crest/tree/readme/gallery#showcase-gallery

huwb:
Yeah this can work.

I never told you what i originally had in mind here, i just realised. i'll state it in case you think its a good direction. i found a way to have an image that links to a youtube video, and takes the thumbnail from the video.

this would mean there is a minimal text, minimal upkeep way to achieve a gallery. the idea came from how three.js does it:

https://threejs.org/

so it would be like that but the images would be thumbnails, then the youtube video would be on the publishers channel with whatever text they want etc etc, so would require minimal upkeep. i didnt ask if anyone was ok with this as no material was submitted.

anyway as i said that was my original intention but i'll leave it up to you what makes sense.

moosichu:
So just keep everything as-is but just remove the text so that we only have the video thumbnail + link?

moosichu:
Updated so that it looks like this now. @huwb is this what you were after? 
image

huwb:
Looks good!

Issue #263, opened 2019-05-30
Dynamic water simulation is time independent


UnknownSoldier1:
Hello,

I have a game where the player can speed up or pause the game simulation and I wrote a custom time provider that applies that to the water as well. As it appears though, this does not affect the dynamic wave simulation and therefore waves created by boats are still simulated, even if my time provider doesn't advance. Also, when my time provider advances faster than normal the dynamic waves also don't simulate faster as they should.

I don't know how hard it is to fix or if this is actually desired behaviour but from my perspective this probably is a bug.

huwb:
If i set time.timescale to 0 then it seems to freeze the dyn waves for me.  How are you controlling the time?

(unfortunately it seems that if i decrease and icnrease timescale quickly then it adds tons of acceleration to the sim for some reason :/ so thats a bug)

UnknownSoldier1:
I'm using this piece of code to do this:

`public class WaterTimeSync : TimeProviderBase
    {
        public override float CurrentTime => time;

        private float time;

        private void Update()
        {
            time += Time.deltaTime * GameManager.current.SpeedManager.GameSpeed;
        }
    }`

You mentioned time.timescale? Is that an independent thing from the time provider?

huwb:
Oh right i see the problem - the sim delta time is driven by unity's delta
time.

            GetSimSubstepData(Time.deltaTime, out numSubsteps, out
substepDt);

I guess ideally the time provider would provide a time and a delta time.
I'd be interested to see a PR with this change if you feel up for trying
it. Otherwise i'll add it to my list and post back when i've had a chance
to look at it.

The workaround for now might be to use Unity's Time.timeScale  for slowing
down time as the sim will repsond to this..

UnknownSoldier1:
Unfortunately using Time.timeScale won't fix my issue as it slows down everything, including UI etc. That said, it's not an urgent fix and I can certainly wait for a fix for this, so tackle this issue when you have the time.

huwb:
In #295 i add delta time to the time provider and this is merged into master now.

Issue #232, opened 2019-04-29
Demo Scene on iOS produces strange disturbance


klauskobald:
I build a test for iOS and get these strange patterns after a few minutes:
<img width="336" alt="Ohne Titel" src="https://user-images.githubusercontent.com/278428/56915141-15921c80-6ab6-11e9-8305-9bba930a4e30.png">

huwb:
Thanks for reporting, thats interesting. While I'm not targeting iOS as a primary platform, this seems like an issue that should be fixed regardless.

huwb:
One thing that might address this issue is to reduce the max sim timestep. Create Crest/Dynamic Wave Sim Settings in the project view and reduce the "Max Substep Dt" to something like 0.012 . This will mean the sim steps at 12ms or less instead of the default 16.67ms. At 30fps, it will do 3 steps. If necessary it could be reduced further.

The unfortunate byproduct of this is an extra sim step has to be done per frame, which will mean slightly more cost.

If anyone tries this and finds it adresses the issue, please post here.

huwb:
Oh this should be resolved on latest on master branch here. Apologies i forgot to confirm.

Issue #55, opened 2018-08-23
Pause bug is back


holdingjason:
Pause  the game, wait awhile, ocean goes nuts when unpausing (that is not a word).  Settles down after a few.

Midda-C:
Is this the same thing as mentioned in #38?  Because yeah, we're still seeing this at very low timescales (and I believe sometimes when paused).

huwb:
Thanks guys... Issues with the sim timestep seem to be resilient super bugs :/. I'll post back when I've had a new attempt.

huwb:
See my last comment on issue #44

huwb:
Should be fixed on 9756b495ad33daee0b1bfd2720425e9aed6fecbe . I tried different patterns of pausing/unpausing, including tabbing away from the editor. Simulation substep counts and dts are now carefully controlled so the above behaviour should not be possible anymore.

Issue #38, opened 2018-08-06
Dynamic Waves and time scale


Midda-C:
It appears as though the force applied to the ocean continues to accumulate if Unity's timescale is set to zero.  We use the numpad for setting the time scale for debug purposes (5 = %50, 1 = 10%, etc), and we noticed that if we're interacting with the water and pause the game by setting the speed to 0, the ocean will explode with tidal waves when we unpause.

A simple way to replicate this is in the threeboats scene, press play, and while running, open the Time Editor (Edit > Project Settings > Time) and set the time scale to 0.  Wait a few seconds, then set it back to 1.  You should see all the wake from the boats go a bit crazy.  The longer it's paused, the crazier it'll be.

On a related note, I have also noticed similar behaviour if there are hitches in the framerate while playing, where the dynamics will suddenly be very strong after a breif (e.g. 0.5 sec) hiccup.  I assume this is the same issue?

Midda-C:
One of the coders on our team has found that this is still happening at very low time-scales (which he often uses for debugging).  I just repro'd the issue on my machine to.  We're seeing crazy dynamics at time-scales of 0.04x or less (but completely paused is fine).

Any ideas on what might be happening there?

huwb:
Ah ok im seeing the same thing here. I changed the approach slightly, it should now go all the way down to 0 ok. However going back up seems to add loads of energy to the system. Maybe this behaviour ok-ish for debugging. I could not get to the bottom of why this is happening, the dts seem ok.

If the remaining issue is a blocker then lets reopen this issue - let me know.

Midda-C:
I just grabbed latest data, and the issue is still present.  In fact, it seems to be back to occurring even when paused, though to to the degree it originally did.

We're not planning to set the speed this slow during our actual game (outside of pausing), so it's not a show-stopper, but one of our programmers does frequently set things this low when debugging things like player movement and such, and currently he just has to disable all of the dynamics.

If you're having trouble reproducing it, let me know if there's anything I can provide to help.

PR #295, opened 2019-07-13
Rewrite wave sim

huwb:
Best reviewed by files rather than by commit. I reviewed data changes and I believe they're all intentional.

Aims to solve a few issues:

* General instabilities in sim - #38, #106, #232
* Issues with changing dt/pausing - #55 
* Allows deltatime to be driven by time provider - #263 
* Wave sim code untidy/unfinalised

Computes the required number of sim steps and sim dt at runtime which is nice. Smaller lods need more sim steps, so calculation is now lod'd. In the future we could look at varying number of updates per lod.

The sim step/dt code should guarantee stability. If the max steps are hit, dynamic waves will slow down slightly (rather than becoming unstable).

Also the sim is changed from a PBD style approach to a normal (height, velocity) representation, so that the velocity does not need to be computed which gives much more stable results.

Besides all the issue fixes, i made a few additions:

* Gravity exposed to enable speeding up the waves
* Reconfigured/tweaked the three boat types. All boats work fairly well now.

Made a video and posted here: https://youtu.be/rkN-CMtafCg


crest/Assets/Crest/Crest/Scripts/LodData/Settings/SimSettingsWave.cs
```diff
@@ -9,16 +9,19 @@ namespace Crest
     [CreateAssetMenu(fileName = "SimSettingsDynamicWaves", menuName = "Crest/Dynamic Wave Sim Settings", order = 10000)]
     public class SimSettingsWave : SimSettingsBase
     {
-        [Header("Simulation Settings")]
-        [Range(0f,32f), Tooltip("The wave sim will not run if the simulation grid is smaller in resolution than this size. Useful to limit sim range for performance.")]
+        [Header("Range")]
+        [Range(0f,32f), Tooltip("NOT CURRENTLY WORKING. The wave sim will not run if the simulation grid is smaller in resolution than this size. Useful to limit sim range for performance.")]
         public float _minGridSize = 0f;
-        [Range(0f, 32f), Tooltip("The wave sim will not run if the simulation grid is bigger in resolution than this size. Zero means no constraint/unlimited resolutions. Useful to limit sim range for performance.")]
+        [Range(0f, 32f), Tooltip("NOT CURRENTLY WORKING. The wave sim will not run if the simulation grid is bigger in resolution than this size. Zero means no constraint/unlimited resolutions. Useful to limit sim range for performance.")]
```
moosichu: What do you mean but this not currently working, should we remove these then?

huwb:
Which issues, could you expand? Is it the water protruding through the boat issue #274? It might be that this change makes it slightly worse, potentially because i've amped up the strength of the object-water interaction while testing. I could dial them back down.

Whats really needed is a thorough pass on object-water interaction to try to combat this problem, or maybe the waves could be forced down below the level of the boat. It's all tricky because the boat affects the water, and the water affects the boat, so there is an undesirable feedback loop here. there's a possible solution where dynamic waves do not affect height queries, but then dynamic waves wouldn't affect objects at all which also doesnt sound desirable.

So lots of issues but this change aims purely to address the issues listed above to improve the sim itself. Having said that i could dial down some of the interaction forces which should help with #274 - style issues.

crest/Assets/Crest/Crest/Scripts/LodData/Settings/SimSettingsWave.cs
huwb: These seem to have broken at some indiscriminate time in the past. I swear they worked once. I thought the move to TAs might have broken them but it seems they were broken before that. I could comment them out, but if anyone has settings for these it will kill their data, so was going to leave as so. Let me know what you think.

moosichu:
Sorry, that was very, very vague feedback. The scenes are bugged or anything like that. I will record some footage because that will likely help. The main thing is that the boats all start crashing into each other when they didn't do so before - it's probably a case of tweaking the settings in the scenes themselves more than anything else.

crest/Assets/Crest/Crest/Scripts/Interaction/ObjectWaterInteraction.cs
```diff
@@ -20,7 +20,7 @@ public class ObjectWaterInteraction : MonoBehaviour
         [Range(0f, 1f), SerializeField]
         float _noiseAmp = 0.5f;
 
-        [Range(0f, 20f), SerializeField]
+        [Range(-1f, 1f), SerializeField]
         float _weight = 6f;
```
moosichu: This default value doesn't seem to fit in the new range?

crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgrDynWaves.cs
```diff
@@ -38,7 +38,7 @@ public override SimSettingsBase CreateDefaultSettings()
         static int sp_Damping = Shader.PropertyToID("_Damping");
         static int sp_Gravity = Shader.PropertyToID("_Gravity");
         static int sp_LaplacianAxisX = Shader.PropertyToID("_LaplacianAxisX");
-
+
```
moosichu: Lingering whitespace, might be worth doing a format on all of the changed files (I'm happy to do so if it saves time).

crest/Assets/Crest/Crest/Scripts/LodData/Settings/SimSettingsWave.cs
moosichu: Shall we create an issue for this once this is merged?

crest/Assets/Crest/Crest/Scripts/Shapes/OceanWaveSpectrum.cs
```diff
@@ -87,11 +87,11 @@ public float GetAmplitude(float wavelength, float componentsPerOctave)
             return a;
         }
 
-        float ComputeWaveSpeed(float wavelength)
+        public static float ComputeWaveSpeed(float wavelength, float gravityMultiplier = 1f)
```
moosichu: To make this consistent with the implementation in the shader, it might be better to have the second parameter just be the value of gravity in ms^2 instead of a multiplier of Earth's gravity.

crest/Assets/Crest/Crest/Scripts/LodData/Settings/SimSettingsWave.cs
```diff
@@ -9,21 +9,26 @@ namespace Crest
...
+        [Range(1, 8), Tooltip("How many simulation substeps are allowed per frame. Run at target framerate with the OceanDebugGUI visible to see how many substeps are being done when the camera is close to the water, and set the limit to this value. If the max substeps is set lower than this value, the detailed/high frequency waves will propagate slower than they would in reality. For many applications this may not be an issue.")]
+        public int _maxSimStepsPerFrame = 3;
 
         [Header("Displacement Generation")]
         [Range(0f, 20f), Tooltip("Induce horizontal displacements to sharpen simulated waves.")]
         public float _horizDisplace = 3f;
         [Range(0f, 1f), Tooltip("Clamp displacement to help prevent self-intersection in steep waves. Zero means unclamped.")]
         public float _displaceClamp = 0.3f;
+
+        public float _gravityMultiplier = 1f;
```
moosichu: Add a ToolTip here and possibly even change it just to gravity in terms of ms^2, I will leave the latter up to your judgement.

crest/Assets/Crest/Crest/Scripts/OceanChunkRenderer.cs
```diff
@@ -160,10 +160,10 @@ public static void ExpandBoundsForDisplacements(Transform transform, ref Bounds
         {
             float boundsPadding = OceanRenderer.Instance.MaxHorizDisplacement;
             float expandXZ = boundsPadding / transform.lossyScale.x;
-            float boundsY = OceanRenderer.Instance.MaxVertDisplacement / transform.lossyScale.y;
+            float boundsY = OceanRenderer.Instance.MaxVertDisplacement;
             // extend the kinematic bounds slightly to give room for dynamic sim stuff
-            boundsY = Mathf.Max(boundsY, 1f);
-            bounds.extents = new Vector3(bounds.extents.x + expandXZ, boundsY, bounds.extents.z + expandXZ);
+            boundsY += 5f;
```
moosichu: This isn't feedback, but just curiousity about why the logic here was changed.

crest/Assets/Crest/Crest/Scripts/Interaction/ObjectWaterInteraction.cs
huwb: didnt see this feedback. good point will do.

crest/Assets/Crest/Crest/Scripts/OceanChunkRenderer.cs
huwb: i've increased the bounds a bit - tiles were getting culled too aggressively

crest/Assets/Crest/Crest/Scripts/Shapes/OceanWaveSpectrum.cs
huwb: discussed on discord

crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgrDynWaves.cs
huwb: I've done it

crest/Assets/Crest/Crest/Scripts/LodData/Settings/SimSettingsWave.cs
huwb: yeah, maybe i'll get your help with this one if you're up for it?

PR #296, opened 2019-07-17
Max detail height affected only by waves

huwb:
Max detail height - camera height relative to sea level where ocean is to minimum scale to maximise detail - was affected a lot by . It caused it to lower which mean ocean was not at full when camera was around sea level. Some lowering is ok - helps for gerstner waves which drop the average surface height lot - but the whirlpool should not lower it, so i've split out some so it can distuingish between gerstner displacements and other.

Editorial note: i doubt this is going to make any sense. maybe we can discuss it tomrrow. a second pair of eyes on testing might be useful if you find the time, no worries if not.


crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
```diff
@@ -320,19 +320,22 @@ void LateUpdateLods()
         /// Shape scripts can report in how far they might displace the shape horizontally. The max value is saved here.
```
moosichu: Probably makes sense to update this comment as it appears to have aged a bit.

huwb:
Merged, thanks

Issue #220, opened 2019-04-11
Underwater curtain doesnt completely cover sides of screen in VR


huwb:
Discussed on discord. Factor '2.1' in UnderwaterCurtain.shader could be increased, seems to work well at 2.5.

Issue #183, opened 2019-02-04
Underwater curtain begins glitching as distance from world center increases (and when viewed from a steep angle)


Nimuecrypto:
This will probably be an easy fix since it's likely very similar to the most recent underwater curtain issue you resolved.  

Here is a video showing the issue:

https://www.youtube.com/watch?v=bz9WVLInW6s&feature=youtu.be

huwb:
It's weird, i wasnt really able to repro this. At (100,000, 0, 100,000) i see a slight flicker, but nothing at 50,000. I parented a sphere to the camera and looked straight up and straight down.

Some instability is inevitable due to precision issues (games i worked on in the past stayed within 4000 units from teh origin) so I dont expect this to work perfectly further out. Still its odd that the issues seem so much more severe in your tests. I'll have a think about what the difference could be.

huwb:
Oh also could you let me know which commit or version you were running if you have this info.

Nimuecrypto:
@huwb I was using a version that was a couple weeks old, but after seeing your comment I downloaded the version you've been testing on.

Result is exactly the same as before the update.  Yeah I don't know why you can't repro.

As you said, some of the test distances are pretty ridiculous.. but I start seeing noticable degradation of the shader on my machine at far smaller distances (<10000).   

When I get a moment I'll see if I can do further testing to see if I can discover what may be causing my results to differ from yours.

In case the difference turns out to be hardware related,  I am using a dual-monitor setup with both monitors connected to a Radeon R9 290X, running driver version 18.12.1.1

huwb:
Ok let me know.

Yeah it could be hardware dependent, maybe some kind of precision issue. My ability to track down hardware dependent issues that i cant repro is severely limited, as I dont have the resources to do more.

On previous projects that i've worked on we had to limit the playable area to around 4km squared. Other things like physics would also become unstable beyond this kind of distance. Is limiting your playable area an option? Alternatively, we recently added floating origin functionality to avoid large distances from the origin, is this an option?

Nimuecrypto:
I don't think I can limit the space to 4km, but I'll give the floating origin branch a shot and see if it helps.  I'll report back on it in the next few days.

huwb:
I'm going to close for now. I'll get notification if you post here.

Issue #147, opened 2018-12-16
Underwater shader failure at depths below 100 meters.


Nimuecrypto:
I was torn about whether to create a new issue for this because it's probably related to the earlier issue I posted regarding the underwater shader failing when viewed from below.  Please excuse me if I should have just posted this as a new comment in that thread.

Anyway, I have a new video showing extreme behavior and failure of the underwater shader at depths below -100 meters (at a scale of 1 unit = 1 meter).   

Beginning at just about the moment the camera passes below 100 meters, it seems the underwater fog stops functioning, and then the edges of the screen effect begin to clip and move inward.  At certain angles the effect striates completely and is visible only as lines.  These effects only compound as depth increases until the underwater effect is no longer visible at all.

https://www.youtube.com/watch?v=h42FJDkWPJ0&feature=youtu.be

huwb:
Thanks, linking to #139 . I think this is fixable, i wont be able to get onto it in the very short term but will post back when i do.

Nimuecrypto:
Not a problem.  I may even get around to it myself, but I'm furiously writing game and netcode at the moment.  I believe in making things functional before making them pretty, so shader bugs kind of take a back seat until I have that other stuff done.

Nimuecrypto:
@huwb I have done as you requested.  I have upgraded to the newest Crest release and performed several experiments. It seems the result is highly dependent on the distance of the camera from the world center.  

In the video below I have the Crest Ocean object at 0,0,0.    I then proceed to test at various distances:

- with player at 0,0,0 (world center) everything works as expected.
- with player at 2000,0,2000 the underwater curtain fails just as the camera goes below water level.
- with player at 5000,0,5000 the shader fails when viewed from directly below and we begin to see little white specks at the LOD edges.
- Things get progressively worse from there.  I tested all the way out to 250000, 0, 250000 (which I know is an unreasonable distance due to floating point limitations), but at that distance you can clearly see speckles and lines where the LODS meet, striations in the underwater shader, and it continues to fail from below.

At various points in the video I click on the Player object as well as the Orbit camera so you can see their transforms.  Obviously keep in mind that the Orbit camera is a child of the player, so it's transform will be in local space.  

Follow the Youtube link below to see these tests.

https://youtu.be/Zhp2XhG1T_M

huwb:
Thank you - thanks for the rigorous testing and findings, really helpful. I'll post back when i have news. I'm confident these issues are solvable.

[...]

Issue #139, opened 2018-12-06
Underwater shader fails when viewed from directly below.


Nimuecrypto:
Two posts in one day... this is a new record for me :)

I am having the issue in the the title - the underwater effect shader fails completely when viewed from beneath it at steep angles.  just from playing around it looks like it occurs at all angles between about 85 and 90 degrees.

![shaderfailure](https://user-images.githubusercontent.com/6428158/49601463-e4248300-f953-11e8-8ff0-abea57a77cea.png)

huwb:
Thanks, looks like a bug. I'll let you know when i get around to looking at it. It might help give me a leg up if you could give me the camera transform (screenshot of the transform component of the cam would work) and perhaps the submarine transform as well. That's usually enough to repro these things.

Nimuecrypto:
Sorry it took me so long, I've been busy this weekend.

Anyway that's easy enough.  In the video below the sub is at 0,0,0 and rotation 0,0,0.    Engines are off, so the only sub motion should be from surface interactions (my own buoyancy calculations combined with a modified version of your "align to normal" stuff).     

I have the orbit camera enabled and set as the viewpoint in the Ocean script.  The underwater curtain and meniscus objects are enabled on the orbit camera.  The first person camera is disabled, as are the underwater objects attached to it.

Here is the result:   https://www.youtube.com/watch?v=djyrnhfJKp0&feature=youtu.be

huwb:
@Nimuecrypto the commit reference above should fix this. It's not perfect but should work "enough".

PR #299, opened 2019-07-20
Replace underwater effect with underwater post-processing shader 

moosichu:
**Status:** Blocked as it is incompatible with existing PP effects, and requires a complex workaround to fix. See issue below.

Simplifies underwater rendering and allows it to be pixel-perfect (replacing the geometry-based skirt).

It would be good to do some profiling as well to see how it compares to the previous implementation.

Explanation on how the system works coming soon.

Issues of interest: #183 #220 #147 #139

Things to complete:
- [x] Fix artifacts on Vulkan.
- [x] Strip-out and clean the ocean mask shader.
- [x] Find a solution that allows this to work with camera roll.
- [x] Confirm that this works with features like edit-and-continue. (@huwb - i dont think it does, but something else breaks edit and continue worse than this, so ill look at fixing it later)
- [x] Profile the performance of the new system. (0.6ms at almost 1080p on huws laptop) @huwb
- [x] Explain the tradeoffs made by the new system, and which problems it aims to solve.
- [x] Reinstate shadows. @huwb . Merged #348 into this branch. Not perfect but better i guess.
- [x] Improve masking values for underwater meniscus. @huwb
- [x] Update documentation so that it explains how to use new system. @moosichu
- [x] Comment and format code properly. @moosichu
- [ ] Update [the Wiki](https://github.com/crest-ocean/crest/wiki/Cut-out-visuals-underwater) so that there isn't a loss of functionality for those upgrading. @moosichu 
- [x] Prevent material edits from being saved to disk at runtime @moosichu
- [x] Assign default materials to UnderwaterCamera script @moosichu
- [x] Review general setup of how the underwater effect should be enabled and attached to camera. @huwb 
- [x] Ensure the system is compatible with VR @moosichu
- [x] Look into performance of OnRenderImage. Function is called only once but frequently costs 0.5-0.7ms (on huws laptop).
- [x] Prime GetInterpolatedProbe (conversation below) @huwb - added profiling tag for now.
- [x] Final checks - working on supported unity versions (2018.4, 2019.1, 2019.2, 2019.3). Tested on windows, added issue below.
- [x] Issue below - how to handle this? (both from a code point of view - runtime errors? - and from a documentation point of view)

Screenshot of new effect:
image

**Issues**

* 2019.1.3 , 2019.1.12, 2019.2.0, 2019.3.0b4 - if post process layer script is enabled, input to postprocess is black . think this is a unity bug, but it means SMAA and other PP cant be used. Maybe related to this? https://github.com/Unity-Technologies/PostProcessing/issues/811 , or this: https://github.com/Unity-Technologies/PostProcessing/issues/433 . The latter has an ugly workaround - put the Post Process Layer on a child object. It seems to work but is pretty ugly..


crest/Assets/Crest/Crest/Scripts/Helpers/UnderwaterCamera.cs
```diff
@@ -0,0 +1,215 @@
...
+                _textureMask.format = RenderTextureFormat.RHalf;
+                _textureMask.Create();
+
+                _depthBuffer = new RenderTexture(source);
+                _depthBuffer.name = "Ocean Mask Depth";
+                _depthBuffer.format = RenderTextureFormat.Depth;
+                _depthBuffer.Create();
+            }
+
+            // Get all ocean chunks and render them using cmd buffer, but with
```
moosichu: I guess this comment needs fixing, haha

crest/Assets/Crest/Crest/Scripts/Helpers/UnderwaterCamera.cs
```diff
@@ -0,0 +1,215 @@
+﻿using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.Rendering;
+
+namespace Crest
+{
+
+    [RequireComponent(typeof(Camera))]
+    public class UnderwaterCamera : MonoBehaviour
+    {
```
moosichu: We should probably add a longform comment here explaining how everything is meant to work.

crest/Assets/Crest/Crest/Scripts/OceanValidation.cs
```diff
@@ -52,13 +52,13 @@ public static void RunValidation(OceanRenderer ocean)
...
-            var underwaters = FindObjectsOfType<UnderwaterEffect>();
+            // UnderwaterPostProcessor
+            var underwaters = FindObjectsOfType<UnderwaterCamera>();
             foreach (var underwater in underwaters)
             {
-                if (underwater.transform.parent.GetComponent<Camera>() == null)
+                if (underwater.GetComponent<Camera>() == null)
                 {
-                    Debug.LogError("Validation: UnderwaterEffect script expected to be parented to a GameObject with a Camera. Click this message to see the script in question.", underwater);
+                    Debug.LogError("Validation: UnderwaterPostProcessor script expected to be parented to a GameObject with a Camera. Click this message to see the script in question.", underwater);
```
moosichu: This text needs to be fixed

crest/Assets/Crest/Crest/Shaders/OceanHelpers.hlsl
```diff
@@ -52,3 +52,9 @@ void SnapAndTransitionVertLayout(float i_meshScaleAlpha, inout float3 io_worldPo
 	if (abs(offset.x) < minRadius) io_worldPos.x += offset.x * o_lodAlpha * GRID_SIZE_4;
 	if (abs(offset.y) < minRadius) io_worldPos.z += offset.y * o_lodAlpha * GRID_SIZE_4;
 }
+
+bool IsUnderwater(const float facing, const float forceUnderwater)
+{
+	const bool backface = facing < 0.0;
+	return backface || forceUnderwater > 0.0;
+}
```
moosichu: This file needs a newline

huwb:
I dont have much time but i got latest and just had a quick browse through some code. I can act on these if you agree with them.

* UnderwaterCamera - i think this makes me assume its some kind of camera. I would call it UnderwaterPostprocess perhaps?
* We should probably avoid calling _currentCamera.GetComponent<UnderwaterCamera>(); for each tile. we could potentially cache this off somewhere, in BeginCameraRendering if that gets called? or maybe another way. i can think about it.

huwb:
Testing on 2019.2.0f1 and 72882abf4a1dbbfd0d8b7fc9e685218a3d5335d9 , i can't seem to get it to work at all, the input to the underwater postproc is a black texture, so the output is always a black texture. tried restarting, updating postprocessing package, deleting library folder.

crest/Assets/Crest/Crest/Scripts/Helpers/UnderwaterCamera.cs
huwb: Done

crest/Assets/Crest/Crest/Scripts/OceanValidation.cs
huwb: Done

crest/Assets/Crest/Crest/Shaders/OceanHelpers.hlsl
huwb: Done

crest/Assets/Crest/Crest/Scripts/Helpers/UnderwaterPostProcess.cs
```diff
@@ -0,0 +1,218 @@
...
+
+        // This matches const on shader side
+        private const float UNDERWATER_MASK_NO_MASK = 1.0f;
+        private const string FULL_SCREEN_EFFECT = "_FULL_SCREEN_EFFECT";
+        private const string DEBUG_VIEW_OCEAN_MASK = "_DEBUG_VIEW_OCEAN_MASK";
+
+
+        public void RegisterOceanChunkToRender(Renderer _oceanChunk)
+        {
+            if (_oceanChunksToRenderCount >= _oceanChunksToRender.Count)
```
huwb: I don't think most of this is necessary. Simply Add and Remove from the list, and use the list Count. The list should do all of this efficiently. I can do this change, just validating my thinking here.

huwb: I confirmed using dumb Add/Clear on list doesn't generate garbage, so replaced this

crest/Assets/Crest/Crest/Scripts/Helpers/UnderwaterCamera.cs
huwb: Had a stab, feel free to review

kinglouie:
I just saw this PR and was wondering if this would also help with my issue https://github.com/crest-ocean/crest/issues/213

My issue was that if I create a porthole the meniscus and curtain won't align perfectly. Additionally the porthole curtain was rendered with the camera perspective as if the curtain is directly before the camera. Maybe you could consider possible requirements for doing portholes is this rewrite?

cowhunt:
@kinglouie yes potentially! i'll need to discuss portholes with @moosichu but our new approach feels flexible here.

if you would have time to grab this branch and test it and see if it works for you let us know - it would help us getting this stuff shippable. thanks!

kinglouie:
I just tested this branch and am really impressed by the results 👍  I just clipped the oceanshader and the maskshader in z direction to get this result:

crestscreen1

Theres only 2 issues left for my needs

1. The sky is rendered as if underwater if the camera is to far below surface. I think there is some kind of detection if there are only back facing ocean vertices visible and if this is the case then the sky gets rendered as if underwater. I think I can fix this by myself if I look more into it.

2. The Underwatereffect is obviously calculated relative to camera position, but in my case it should only be applied up to the cross section which is at the z-axis zero point. To fix this I'm lacking the maths knowledge :( 

I'm really looking forward to this PR :)

crest/Assets/Crest/Crest/Scripts/Helpers/UnderwaterPostProcess.cs
```diff
@@ -0,0 +1,244 @@
...
+
+            _underwaterPostProcessMaterial.CopyPropertiesFromMaterial(OceanRenderer.Instance.OceanMaterial);
+
+            if (_viewOceanMask)
+            {
+                _underwaterPostProcessMaterial.EnableKeyword(DEBUG_VIEW_OCEAN_MASK);
+            }
+            else
+            {
+                _underwaterPostProcessMaterial.DisableKeyword(DEBUG_VIEW_OCEAN_MASK);
```
huwb: I'm not sure what the overhead of calling DisableKeyword each frame might be. i guess it might cache it internally and early out cheaply?

crest/Assets/Crest/Crest/Scripts/Helpers/UnderwaterPostProcess.cs
```diff
@@ -0,0 +1,244 @@
...
+                // lead Thomas Hourdel.
+                bool saveChunksToRender =
+                    XRSettings.enabled &&
+                    XRSettings.stereoRenderingMode == XRSettings.StereoRenderingMode.MultiPass &&
+                    _mainCamera.stereoActiveEye != Camera.MonoOrStereoscopicEye.Right;
+
+                if (!saveChunksToRender) _oceanChunksToRender.Clear();
+            }
+
+            _underwaterPostProcessMaterial.CopyPropertiesFromMaterial(OceanRenderer.Instance.OceanMaterial);
```
huwb: I'm unsure about calling this every frame. in the old implementation there was a checkbox on the script to copy the params every frame. it would default to on so be friendly by default, but allows people to turn it off if they know they don't need it. would that be an option here or does it need to happen every frame?

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
```diff
@@ -233,6 +236,10 @@ void LateUpdate()
                 Debug.LogError("_viewpoint is null, ocean update will fail.", this);
             }
 
+            LightProbes.GetInterpolatedProbe(_viewpoint.position, null, out _sphericalHarmonicsL2);
```
huwb: this could be a quite expensive call, seems like its doing a lookup into tetrahedra. should we put it in a profile tag so that people will report it back to us if they see it?

the second arg seems to be there to help speed this up. i wonder if we can give it a renderer? one idea might be to give it one of the ocean tiles' renderer, or maybe theres another we could use?

huwb:
I gave the changes a scan through and put some notes, i didnt see major things (such as structural issues). Nice one

huwb:
Are you ok with turning the meniscus on by default (3ae51f46a2f38b35918c65a90e6e20eff1d5e855). My eyes bleed without it :)

huwb:
Profiled it in nsight, on my laptop, at 1680x1050 (so almost 1080p). took 0.6ms it seems - constructing the mask and then doing the fullscreen effect.

This is "not nothing" but an acceptable cost for now i think (SMAA took a little over 1ms). In the future we could potentially explore writing the mask at the same time as rendering the surface using an RWTexture (or did we decide this wasn't possible?).

moosichu:
@huwb I had a chat with Unity's post processing lead, and discussed options. He said that the most optimal thing would be to split-out front and back-face ocean rendering into two seperate draw calls (which would reduce in-shader branching) and then write the mask to the stencil buffer. We would then have to render a quad in front of the camera (because the built-in pipeline wipes the stencil buffer before the post-processing stage) and run the underwater effect then. I think for now he said that this is a good and sensible solution, but we coud definetely look at optimisations in the future. How does it fair compared with the skirt shader?

Also, it's worth profiling the ocean shader itself in the comparison as well, as this change removes underwater fog from that implementation.

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
moosichu: If I remember correctly I couldn't find much documentation on how this works, so I'm happy to do anything that might make this more efficient. :)

crest/Assets/Crest/Crest/Scripts/Helpers/UnderwaterPostProcess.cs
moosichu: That was me going for the "safest" option with regards to keeping the two materials in-sync. I'm happy following your lead on this here. I guess we could have a checkbox to enable copying on every frame, and also add a button to copy the material params in the editor and expose a function to drive that from code?

moosichu: I'm happy to implement this if you think that these would be good approaches.

crest/Assets/Crest/Crest/Scripts/Helpers/UnderwaterPostProcess.cs
moosichu: I would really hope so. We could profile this and check?

huwb:
Yeah that could work, however we have a relatively high number of verts, and each vert does a bunch of texture reads. I think doing a pass of the ocean is likely an expensive operation for this reason and breaking into passes might make it worse.

If this assumption is correct, writing mask values during the ocean shader could be an elegant solution. it would also remove the need to register ocean chunks, and the latter is a blocker on HDRP for annoying reasons - on hdrp we dont have a way of knowing which camera is rendering, so we dont know which underwater script is relevant, so the register calls are currently knocked out.

I tried an experiment binding a rw texture for the ocean pass: https://github.com/crest-ocean/crest/tree/feature/post-process-underwater-single-pass . for some reason it doesn't bind the texture successfully, in renderdoc its unbound. if you get a chance could you take a look and let me know if you have ideas why its not binding, or tests i could do? thanks dude

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
huwb: Cool, nice one. Let me know if i can help

crest/Assets/Crest/Crest/Scripts/Helpers/UnderwaterPostProcess.cs
huwb: I tried to make it copy the params once on start, and then make the subsequent copies optional. Unfortunately it doenst work - the underwater appears black.

My best guess is that one of our ocean shader uniforms does not get set before init() time, and this value is therefore not set on the underwater material. This is a shame but i guess we'll leave it as is.

crest/Assets/Crest/Crest/Scripts/Helpers/UnderwaterPostProcess.cs
huwb: Yes that would be good. I did a very quick pass with the profiler and there's a fair amount of time being spent in this function - around 0.3ms-ish on average, but a number of times 0.5ms, highest 0.7ms).  i'll add a todo to poke around disabling chunks of the function to determine where this time is going..

huwb:
Hey, could you speak to what this keyword does: FULL_SCREEN_EFFECT

I did not understand precisely when this would kick in and what it does.

I've made the code disabled the postprocess if the camera is more than 2m away from the water. In this case, is this keyword now redundant, or is it still useful?

crest/Assets/Crest/Crest/Scripts/Helpers/UnderwaterPostProcess.cs
huwb: I've pushed a slightly different arrangement that essentially does this

crest/Assets/Crest/Crest/Scripts/Helpers/UnderwaterPostProcess.cs
huwb: I didnt notice a huge overhead calling this. (the lions share of hte cost seemed to be in submitting the ocean tiles)

moosichu:
> Hey, could you speak to what this keyword does: FULL_SCREEN_EFFECT
> 
> I did not understand precisely when this would kick in and what it does.
> 
> I've made the code disabled the postprocess if the camera is more than 2m away from the water. In this case, is this keyword now redundant, or is it still useful?

Yes, I think we want to keep it. It does the opposite of disabling the post processing effect when we are far above the water: if we are far-enough below the water we don't bother checking the mask to determine if we need to render the post-processing effect.

huwb:
Ah now i get it. In this instance could we push this check outwards and avoid constructing the mask completely? Thats where a bunch of the cost is going (0.2-0.3ms roughly of cpu to submit the tiles), and i guess a good chunk of the 0.6ms GPU time as well.

moosichu:
@huwb I was thinking that as well. Although currently the post processing shader is using the depth buffer generated by the mask, so we would need to change that.

huwb:
Gotcha, makes sense. In this case lets leave it.

The ocean does render depth during its normal render pass, but not into the 'opaque depth' buffer that would get passed to postprocessing. As a future optimisation we could look at retrieving this depth buffer instead of rendering it with the mask. I'll add a note to the code.

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
huwb: I ended up taking this on. I added some comments to the last few commits, but exec summary is i moved the lookup into the postprocess and added profile tags to keep an eye on perf.

daleeidd:
I was able to reproduce the black screen issue in 2019.3.0f1. It only occurred in multi-pass VR. Version 2.3.0 of the Post Processing package fixes the issue. You might need to enable "Show preview packages" and it is located under the "See All Versions" dropdown. From the [change log](https://docs.unity3d.com/Packages/com.unity.postprocessing@2.3/changelog/CHANGELOG.html):
> Fix for XR multipass and legacy built-in renderer (case 1197855, 1152584)

Looks great nice in VR - especially with a really low near clip plane.

The only issue is that PP effects aren't applied to the underwater effects. If I had a strong red tint, the underwater fog is still blue.

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterPostProcess.cs
```diff
@@ -0,0 +1,204 @@
...
+            if (OceanRenderer.Instance == null)
+            {
+                Graphics.Blit(source, target);
+                _eventsRegistered = false;
+                return;
+            }
+
+            if (!_eventsRegistered)
+            {
+                OceanRenderer.Instance.ViewerLessThan2mAboveWater += ViewerLessThan2mAboveWater;
```
skaughtx0r: This is leaking EventHandlers every frame. Need to set `_eventsRegistered = true` in here.

skaughtx0r: @moosichu Not sure if you saw my comment...

moosichu: Thanks for noticing! Was focused on other issues and other things have been blocking this from getting merged anyway. But will do that now.

skaughtx0r:
Just want to note, I just tested on 2019.3.0f3 with PostProcessing 2.3.0 and I'm still getting the black screen issue.

daleeidd:
@skaughtx0r Thanks. I must of been seeing a different issue. Are you perchance using bloom?

skaughtx0r:
@moosichu Do you know the work-around to fix the black screen? All the links to the PostProcessing issues are dead.

moosichu:
Cheers for the update, will have a look into it!

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.cs
```diff
@@ -0,0 +1,165 @@
...
+            {
+                _camera = GetComponent<Camera>();
+            }
+
+            if (_cameraFrustumPlanes == null)
+            {
+                _cameraFrustumPlanes = GeometryUtility.CalculateFrustumPlanes(_camera);
+            }
+
+            Enable();
```
daleeidd, line 102: Here we would wrap this and check which RP is active and call the appropriate Enable.

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.cs
```diff
@@ -0,0 +1,165 @@
...
+            Instance = this;
+        }
+
+        void OnDisable()
+        {
+            Disable();
```
daleeidd, line 108: Here we would wrap this and check which RP is active and call the appropriate Disable.

daleeidd:
@huwb I think this is ready for review and merge. It went through a pretty serious refactor and many fixes and improvements along the way.

- Meniscus has been fixed/improved
  - Now samples in the opposite of the ocean surface normal
  - Now works went camera is rolled by using the horizon normal
  - Fixed artefacts caused by imperfections in the mask (from the VFACE bug I believe)
- Brought over far plane multiplier from HDRP
- Remove legacy VR stuff and reduced VR code overall
- Fixes XR multi-pass here and will also for HDRP (finally)
- Small C# optimisation with checking whether to force full-screen
- Brought back the old curtain as "obsolete" and made them coexist until URP XR SPI is sorted
- Now uses run-time materials
- Brought over some improvements from the PPv2 branch
- Refactored using partial classes and UnderwaterRenderer can be used across pipelines with minimal changes downstream
- Moved a lot of RP specific code out of common static methods
- Other minor fixes and improvements

### Refactor

I gave the refactor a lot of thought and went with partial classes. The utilities file has been dissolved. The goal is to make development easier in BIRP and across RPs. There are now three files:
- UnderwaterRenderer
- UnderwaterRenderer.Mask
- UnderwaterRenderer.Effect

These are all for the same UnderwaterRenderer MonoBehaviour.

UnderwaterRenderer component is the one component to work across all RPs. It will require maybe a couple of lines of changes downstream to enable the HDRP custom pass or URP render feature (I have added review comments as to where this will happen). BIRP is driven by OnPreRender and AddCommandBuffer as before, but these won't run when SRP is being used which makes a nice fallback. It should generally only have stuff common to both Effect and Mask and MonoBehaviour stuff.

The utility methods etc are now grouped together into their respective files (Mask and Effect). And they can be accessed for HDRP and URP like so UnderwaterRenderer.PopulateMask etc.

### Documentation

Will handle this once everything is sorted out downstream.

### Issues

I don't think there are any. It works with the PPv2 Stack. Only issue might be MSAA causing an outline but I think that was an issue already and is the usual problem of non MSAA depth with MSAA colour texture discrepancy.

I might have to make another PR as I don't know if I can take this out of draft.

crest/Assets/Crest/Crest-Examples/Shared/Scripts/CamController.cs
```diff
@@ -45,6 +45,9 @@ public class CamController : MonoBehaviour
     [System.Serializable]
     class DebugFields
     {
+        [Tooltip("Disables controller preventing the camera from rolling (rotating on the z axis).")]
+        public bool disableCameraRollPrevention = false;
```
huwb, line 49: "disable prevention" -> tempted to invert: "enable camera roll". but debug code so doesnt matter too much!

crest/Assets/Crest/Crest/Scripts/Editor/BuildProcessor.cs
```diff
@@ -60,6 +60,11 @@ public void OnProcessShader(Shader shader, ShaderSnippetData snippet, IList<Shad
             {
                 ProcessUnderwaterShader(shader, data);
             }
+
+            if (shader.name == "Crest/Underwater Curtain")
```
huwb, line 64: i think downstream this code calls a function to check if its one of the underwater shaders. doesnt really matter but just a note in case you want to pull it up stream to here.

crest/Assets/Crest/Crest/Scripts/Helpers/PropertyWrapper.cs
```diff
@@ -21,11 +21,34 @@ public interface IPropertyWrapper
...
+    }
+
     [System.Serializable]
     public class PropertyWrapperMaterial : IPropertyWrapper
     {
         public PropertyWrapperMaterial(Material target) { material = target; }
-        public PropertyWrapperMaterial(Shader shader) { material = new Material(shader); }
+        public PropertyWrapperMaterial(Shader shader)
+        {
+            Debug.Assert(shader != null, "Cannot create required material because shader is null");
```
huwb, line 36: preference for "material property wrapper" instead of just "material" as one might confuse this for a general unity message (maybe) if it got reported on discord

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
```diff
@@ -1011,38 +1015,61 @@ void LateUpdateLods()
...
+            var definitelyUnderwater = false;
+            var volumeExtinctionLength = 0f;
+
+            if (isUnderwaterActive)
             {
-                return;
+                definitelyUnderwater = ViewerHeightAboveWater < -5f;
+                var density = _material.GetVector("_DepthFogDensity");
+                var minimumFogDensity = Mathf.Min(Mathf.Min(density.x, density.y), density.z);
+                volumeExtinctionLength = -Mathf.Log(_underwaterCullLimit) / minimumFogDensity;
```
huwb, line 1028: out of paranoia i think i'd like us to bound _underwaterCullLimit from below. i know it already has a Range() on the param, but perhaps it gets 0 due to some dodgy serialisation scenario, or maybe someone sets it from code at runtime.

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Effect.cs
```diff
@@ -0,0 +1,421 @@
+// Crest Ocean System
```
huwb, line 1: I assume the XYZ.Effect.cs naming is standard? have not come across it before so just curious

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Mask.cs
```diff
@@ -0,0 +1,131 @@
...
+            Material oceanMaskMaterial,
+            bool debugDisableOceanMask
+        )
+        {
+            GeometryUtility.CalculateFrustumPlanes(camera, frustumPlanes);
+
+            // Get all ocean chunks and render them using cmd buffer, but with mask shader.
+            if (!debugDisableOceanMask)
+            {
+                // Spends approx 0.2-0.3ms here on dell laptop
```
huwb, line 113: potential candidate for removing.

otherwise maybe specialising it to be concrete: "2018 Dell XPS 15"

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.cs
```diff
@@ -0,0 +1,165 @@
...
+        public class DebugFields
+        {
+            public bool viewOceanMask = false;
+            public bool disableOceanMask = false;
+            public bool disableHeightAboveWaterOptimization = false;
+        }
+
+        Camera _camera;
+        bool _firstRender = true;
+        static int _xrPassIndex = -1;
```
huwb, line 78: 's_' for static

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.cs
```diff
@@ -0,0 +1,165 @@
...
+        [Predicated("useHorizonSafetyMarginMultiplier"), Range(0f, 1f)]
+        [Tooltip("A safety margin multiplier to adjust horizon line based on camera position to avoid minor artifacts caused by floating point precision issues, the default value has been chosen based on careful experimentation.")]
+        public float horizonSafetyMarginMultiplier = 0.01f;
+
+        [Predicated("useHorizonSafetyMarginMultiplier"), DecoratedField]
+        [Tooltip("Dynamic resolution can cause the horizon gap issue to widen. Scales the safety margin multiplier to compensate.")]
+        public bool scaleSafetyMarginWithDynamicResolution = true;
+
+        [Space(10)]
+        public DebugFields debug = new DebugFields();
```
huwb, line 67: '_' missing

huwb:
Amazing. finished a pass over the water, all looked good, just nitpicks/questions. i will approve once i've had the chance to test it

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Effect.cs
daleeidd: I don't think there is an official standard. But Unity does this in Graphics repository and apparently Microsoft does it too.

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.cs
daleeidd: It's public. Aren't "_" for private members?

crest/Assets/Crest/Crest/Scripts/Helpers/PropertyWrapper.cs
daleeidd: It is now prefixed with "Crest.PropertyWrapperMaterial: " similar to other debug/asserts we have. Will that work?

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Mask.cs
daleeidd: Thanks. I've updated with the specific device.

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.cs
daleeidd: Done.

crest/Assets/Crest/Crest-Examples/Shared/Scripts/CamController.cs
daleeidd: I agree that would be much nicer wording. For debug options I believe they should be checked to be enabled (ie always disabled by default) from a UX perspective.

daleeidd: I'm not a fan of "prevention". Does disableCameraRollClamp or disableCameraRollConstraint sound better?

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.cs
huwb: no the underscore is for non-static members. properties have no underscore, but members do

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Effect.cs
huwb: ok gotcha.. if we're going with this i think we shoudl add this to our coding standard - in which case i'd like to understand when to use it and how. my gut tells me tha the first part of the name should perhaps match the namespace, but im just guessing.

but if we do introduce it to our standards there will be scripts that dont conform, and renaming scripts is ugly, so i think im going to be boring and vote to keep it consistent with how we usually do it - simply UnderwaterRendererEffect, unless that feels bad.

crest/Assets/Crest/Crest/Scripts/Helpers/PropertyWrapper.cs
huwb: sure, its not showing me that change on github but im probably just failing at it. but what you described sounds good. one of these days ill do a pass over all asserts/logs to add a prefix!

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
daleeidd: Done. I used clamp to enforce minimum and maximum.

crest/Assets/Crest/Crest-Examples/Shared/Scripts/CamController.cs
huwb: i feel like "enable camera roll" is disabled by default, and checked to be enabled. but disableCameraRollClamp is probably my favorite of the lot so thumbs up for that :)

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Effect.cs
daleeidd: We wouldn't have to change anything since it is only for partial classes which is also how Unity uses it. It isn't possible to include namespaces as Unity requires component names and file names to match exactly.

daleeidd: I'm happy to go with whichever you want to though.

huwb: oh so its for partial classes. that works great. approved!

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.cs
daleeidd: Sorry. Looks like I need to brush up on the contributing document again... should I make them all private too? I don't think anyone would really need them public.

crest/Assets/Crest/Crest-Examples/Shared/Scripts/CamController.cs
daleeidd: Oh... I feel dumb :^). Sorry. I have renamed to "enable camera roll"

crest/Assets/Crest/Crest/Scripts/Editor/BuildProcessor.cs
daleeidd: Done. Thanks.

crest/Assets/Crest/Crest/Scripts/Helpers/PropertyWrapper.cs
daleeidd: Thanks. Same. I had to go to the files tab otherwise it will show the changes at time of review.

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
huwb: great

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.cs
huwb: if in doubt, private is a good default choice i reckon. i feel theres no point creating serialised data (that might need to be managed/migrated/fixed in the future) if we're not sure.

huwb:
image

lol

I tested in editor and standalone (win64) and did not encounter issues with it. thumbs up for merging

huwb:
🎉

PR #313, opened 2019-08-04
Ramp planar reflection distortion with pixel Z

huwb:
Not ready to be merged - experimental for now.

### DIFF
--- a/crest/Assets/Crest/Crest/Shaders/Ocean.shader
+++ b/crest/Assets/Crest/Crest/Shaders/Ocean.shader
@@ -68,7 +68,7 @@ Shader "Crest/Ocean"
 		// Dynamically rendered 'reflection plane' style reflections. Requires OceanPlanarReflection script added to main camera.
 		[Toggle] _PlanarReflections("Planar Reflections", Float) = 0
 		// How much the water normal affects the planar reflection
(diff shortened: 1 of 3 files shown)


huwb:
Closing for lack of feedback. We can revive this if it comes up again.

huwb:
@Winux1062 tested this and said it works well, so reopening for reconsideration

daleeidd:
@huwb I brought this one up-to-date as someone asked for it. I think all it needs is a range property to scale this distortion multiplier to give users control. Thoughts?

huwb:
Yeah I guess so. I suppose this is birp only?

daleeidd:
Also URP until SG swap.

daleeidd:
All done. I didn't add another multiplier we already have one.

huwb (commented):
Thanks. So this should work for existing cases and not break anything for existing users? Just checking.

huwb:
I tried to approve but it didn't let me. That's a new one. Consider it approved..

daleeidd:
> Thanks. So this should work for existing cases and not break anything for existing users? Just checking.

It will be different. I could add it to "breaking"? Or we could hold off until 5.0.

PR #315, opened 2019-08-05
Sample in data from larger cascade for foam and shadow

huwb:
This avoids objectional boundaries between cascades that visible when the camera moves around, in the shadow and foam. The result is much more robust now.

### DIFF
--- a/crest/Assets/Crest/Crest/Shaders/Resources/UpdateFoam.compute
+++ b/crest/Assets/Crest/Crest/Shaders/Resources/UpdateFoam.compute
@@ -30,25 +30,33 @@ void UpdateFoam(uint3 id : SV_DispatchThreadID)
 	const float2 input_uv = IDtoUV(id.xy, width, height);
 	const float2 worldPosXZ = UVToWorld(input_uv, sliceIndex);
 	const float3 uv_slice = float3(input_uv, id.z);
-	const float3 uv_slice_source = WorldToUV_Source(worldPosXZ, sliceIndexSource);
-	// #if _FLOW_ON
-	half3 velocity = half3(_LD_TexArray_Flow.SampleLevel(LODData_linear_clamp_sampler, uv_slice, 0).xy, 0.0);
-	half foam = _LD_TexArray_Foam_Source.SampleLevel(LODData_linear_clamp_sampler, uv_slice_source
-		- ((_SimDeltaTime * _LD_Params_Source[sliceIndexSource].w) * velocity), 0
-		).x;
-	// #else
-	// // sampler will clamp the uv_slice currently
-	// half foam = tex2Dlod(_LD_TexArray_Foam_Source, uv_slice_source).x;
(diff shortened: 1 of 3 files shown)


crest/Assets/Crest/Crest/Shaders/Resources/UpdateFoam.compute
```diff
@@ -30,25 +30,33 @@ void UpdateFoam(uint3 id : SV_DispatchThreadID)
...
-		foam = 0.0;
+	if (sliceIndexSource < 0 || sliceIndexSource >= depth)
+	{
+		// Leave foam at 0 if going off the array
+	}
+	else if (max(r.x, r.y) <= r_max)
+	{
+		foam = _LD_TexArray_Foam_Source.SampleLevel(LODData_linear_clamp_sampler, uv_slice_source, 0).x;
+	}
+	else
```
moosichu: shouldn't this be:
`else if (sliceIndexSource + 1 < depth)`
?

crest/Assets/Crest/Crest/Shaders/Resources/UpdateShadow.compute
```diff
@@ -124,15 +124,25 @@ void UpdateShadow(uint3 id : SV_DispatchThreadID)
...
 
-	fixed2 shadow = 0.0;
+	half2 shadow = 0.0;
+	const half r_max = 0.5 - _LD_Params_Source[_LD_SliceIndex_Source].w;
 
 	// Shadow from last frame - manually implement black border
 	float3 uv_source = WorldToUV_Source(shadowCoords._WorldPosViewZ.xz, _LD_SliceIndex_Source);
 	half2 r = abs(uv_source.xy - 0.5);
-	if (max(r.x, r.y) < 0.49)
+	if (max(r.x, r.y) <= r_max)
```
moosichu: do we want similar array-bound checks to foam? Or should we be good here? Sorry, haven't thought too hard about it but just checking in-case the inconsistency isn't intentional. Could also be something we discuss outside of this change.

crest/Assets/Crest/Crest/Shaders/Resources/UpdateFoam.compute
huwb: yeah thanks. there was also another guard missing. theres now a lot of dynamic branches, will think if it can be reduced.

crest/Assets/Crest/Crest/Shaders/Resources/UpdateShadow.compute
huwb: ok i guess the index in shadow is fine, as _LD_SliceIndex_Source will be clamped, but needs careful checking for foam

huwb:
I've pushed a new version with feedback addressed, let me know when you get a chance to check it.  Shadow needs less checks because _LD_SliceIndex_Source is clamped in c#.

Issue #301, opened 2019-07-24
Waves not showing up


Chadedala:
**Describe the bug**
Waves not appearing. Both in editor and standalone build.

**Screenshots / video**
![image](https://user-images.githubusercontent.com/939901/62424501-fd690500-b683-11e9-9a8f-d8415196e9ff.png)

**Log**
No errors in Editor log.

**Versions**
Version: Unity Standard Builtin Render Pipline, no HDRP no LWPR
Unity Version 2018.4.4.f1
Zip file was downloaded 24th July 2019 , so it was probably commit 6a6805e . (Huw: I tried it and it worked fine on my laptop).

**To Reproduce**
I start Unity, import the Package, no waves.

**Platform**
In editor. 
Build modules active for Mac, Linux, Windows and WebGL.

**Hardware**
GTX 1080 SLI
Driver Version : 22.21.13.8205

CPU 2x Intel I 7 6800 k
System Memory 64 GB.
Operation Software: Windows 8.1

**Misc information**
Other data types seem to work:
![image](https://user-images.githubusercontent.com/939901/62424799-c1846e80-b688-11e9-9d31-49cbbb5df38a.png)

huwb:
Just realised - which package are you importing? We don't maintain packages on this site anymore due to time constraints, the last package is from quite some time ago now.

The way the github version is installed is by using the "Clone or download" button on the main page. You'll find the core crest files and the example content here: https://github.com/crest-ocean/crest/tree/master/crest/Assets/Crest

huwb:
Just bumping this in case you have more info for us.

huwb:
From discord discussion: Zip file was downloaded 24th July 2019 , so it was probably commit 6a6805e376a094aa09a2290e06967d6df7b00264

huwb:
From a windows 7 setup:

![image](https://user-images.githubusercontent.com/939901/62425049-8e8faa00-b68b-11e9-8757-c05c89529363.png)

Windows 7 stock,
Graphic Card; GTX 750
Driver Version : 26.21.14.3039
Problem the same. So: This is current build from Github ( I download some Minutes ago ) 
Main Demo Scene, also no waves, same Problem with slider.
oh, systemMemory 8 GB

huwb:
I sent a build made from my laptop, result is the same:

![image](https://user-images.githubusercontent.com/939901/62425211-7db55780-b6d0-11e9-9834-f1d398b3c533.png)

huwb:
[...]

Issue #291, opened 2019-07-11
Switch doesn't support texture arrays properly


dizzy2003:
EDIT FROM @moosichu : Scroll to bottom to see where things have landed.

Switch ocean is very flat and boat is hardly moving, debug info is all wrong :(

![image](https://user-images.githubusercontent.com/15023384/61016114-df381e00-a3d1-11e9-9885-c543fe1f9c84.png)

PC Image
![image](https://user-images.githubusercontent.com/15023384/61016093-bf085f00-a3d1-11e9-926a-f4dbed03ebaf.png)

Double checked shaders are compiling fine, no errors in log

I had to change #if SHADER_API_VULKAN to

#if SHADER_API_VULKAN ||SHADER_API_SWITCH

moosichu:
It looks like something is going wrong with the way texture arrays are being handled, an issue independent from the shaders themselves (seeing how it's affecting all of the lods in the same way). @huwb has a lot more experience debugging consoles than I do, do you have any thoughts on how to proceed with this?

moosichu:
Hi @dizzy2003,
I've chatted with @huwb and we think we have a strategy to start debugging this. Having done some research, we don't see any reason why the new system shouldn't work on the Switch.

I'm going to work on putting together a branch this evening which will help test for some things. 

However, in the meantime, it would be really helpful to be able to see the following:
- Which version of Unity are you using?
- If you go to the whirlpool scene, what is the flow LodData showing in the debug menu?
- If you change line ~112 in LodDataMgrSeaFloorDepth.cs from:
  `buf.ClearRenderTarget(false, true, Color.white * 1000f);` to
  `buf.ClearRenderTarget(false, true, Color.white * 0.01f * lodIdx);`, what do you see in the SeaFloorDepth lods?

And just to confirm, this is with `UseGeometryShader` set to false right?

Also, you can email me at `readcuttingt at gmail dot com`if you are more comfortable sharing any extra information you have there.

moosichu:
For the last bullet point, you should see something like this:
![image](https://user-images.githubusercontent.com/10348641/61173749-7b5b5400-a58f-11e9-81b4-61ebfd58d6d9.png)

(Although we don't expect the geometry to render for obvious reasons, this was taken with the camera position of `(0, 300, 0)` in the main scene.

moosichu:
@dizzy2003 I'm going to have a look at this tonight, but I was wondering if you managed to make any headway, obtain any new information including the extra lod info that could help.

[...]

Issue #279, opened 2019-06-29
Waves not forming properly, LOD and other strangeness


Draydak:
**Describe the bug**
Fresh download of Crest (29th June 2019) from Git. Open in Unity, open any demo scene and the waves are all Muted / flat. The Whirlpool scene was craziest, see video. There does not appear to be any wake / bow wake on any boats. But it could just be so shallow, it's not visible.

LOD appears to be behaving oddly, also only appearing in bottom image slot (again see video)

**Screenshots / video**
[Crest Issue](https://youtu.be/_DSPuQbSBBc)

![image](https://user-images.githubusercontent.com/939901/63832846-8b759a00-c969-11e9-9bcd-832046d7b234.png)

**Log**
No error's / problems in Log at all.

**Versions**
Git version, master branch downloaded (29th June).
Unity 2018.4.2f1

**To Reproduce**
Download from Git master branch to empty folder.
Open in Unity, open any demo scene.
**No** changes, **no** modifications, **not** added to existing project etc.

**Platform**
Editor and standalone
Windows 10
Nvidia Geforce GT 650M

**Additional context**
I believe the graphics card is fully DX11 compatible, so should be capable of handling required shaders. (it worked with earlier versions).
All drivers are up-to-date.

moosichu:
Hi @Draydak,

A few things which could help us debug this issue:
- Could you please enable the debug view of all the other LodMgrs and send a screenshot of that (tick the boxes on the bottom of the game screen).
- Could you please add ` #pragma enable_d3d11_debug_symbols` to line 27 in `AnimWavesGerstnerBatches`
- Could you please [capture, save and upload](https://docs.unity3d.com/Manual/RenderDocIntegration.html) a renderdoc capture so we can try and debug the issue.

Doing the first one should be fairly easy, so do that first. If you can take the time to do this it would really help because we are unable to recreate the bug on our end!

moosichu:
Oh, another quick thing that you could do is a take a screenshot of what things look like when you uncheck `Shape combine pass` in the overlay on the left.

Draydak:
![image](https://user-images.githubusercontent.com/39652045/60394039-3ad6f180-9b1e-11e9-85f3-93629e8365e0.png)

Draydak:
Unchecked shape combine pass, this made big difference
![image](https://user-images.githubusercontent.com/39652045/60394093-fac43e80-9b1e-11e9-8986-1b06e478a6f4.png)

[...]

PR #316, opened 2019-08-06
Replace Texture Arrays with Individual Textures

moosichu:
DO NOT MERGE, IN DEVELOPMENT AND NOT CURRENTLY FUNCTIONING

Creating this PR now so we can evaluate the scope of the changes and analyse the tradeoffs that this change is making.

Problems this could solve.

✔️ #291 - Switch doesn't support texture arrays properly (NOTE: Unity have said they are aware of this bug and are working on a fix)
✔️ #279, #301 (MAYBE), not sure about this, but the change could be relevant to these issues (and at least allow us to test potentially causes for them).
✔️ Allow for the independent ping-ponging of simulations at different lods (eg. for the dynamic sim where we might want a different number of steps at each scale).

Problems potentially introduced.
❌ Some code complexity increase: can make things harder to maintain, as each simulation needs to keep track of more state. (Also, we will have to set material properties per-lod again, which also presents CPU overhead)
❌ Sampling in data from across cascades #315 - we will need to carefully think about how to tackle problems like these going forward
❌ Remove the potential for further optimizations and decreases in complexity which we have yet to take advantage of.
❌ All multi-lod single-dispatch shaders will no longer be possible. Increasing CPU overhead (we need to properly profile the costs here).

This change aims to mitigate the above issues by restricting this change *only* to textures and keeping the existing naming semantics.  The reason for this is that we may want to move back to using texture arrays, especially if the mentioned bugs all get squashed by Unity (could we reach-out to them about some of them?).

Once this PR is in a working state, I will also have a look and see how we could potentially factor code so that we can limit the changes that need to be made - abstracting TA behavior such that this can be achieved.


moosichu:
Every TODO is marked with `TODO(SRT):`

huwb:
Thanks dude, great summary. Couple of small clarifications that i didnt understand while reading:

Reduce the need for some branching in the shader to handle sampling out of TA bounds <- is this still a thing? i think we're solving this problem by duplicating the data for the last cascade, but maybe i'm missing/forgetting some branches in some shaders?

Some code complexity decrease <- could you please cite which ones? not trying to nitpick, just to build a clear picture.

moosichu:
You are right on the former, and on the latter I think that was just referring to the branching, so that can go as well.

dizzy2003:
Is there any ETA on this branch working, we have a switch build to deliver at the end of the month and will have to take a stab at this ourselves soon. 

If we do it, any pointers on how you would like it done would be good, so we can hand our changes back. I agree abstracting out the texturearrays would be the best solution

moosichu:
Hi @dizzy2003,

We are currently trying a different approach now because Crest does now have some behaviour now which would break without TAs. However, until the Switch version of Unity is patched to better support texture arrays, you are more than welcome to try this branch out. It isn't "clean" but should be serviceable enough for a demo.

Animated Waves and Depth work in this branch, but I haven't done so for the other LodData, if you could let me know which of the others you use I can focus on getting them working. However, this would be as a temporary fix for the upcoming build you have to deliver, we have another solution in the works that should fix the issues people have been having on their PCs whilst keeping texture arrays. Would that work for you in the long term once Unity has been fixed?

dizzy2003:
> Hi @dizzy2003,
> 
> We are currently trying a different approach now because Crest does now have some behaviour now which would break without TAs. However, until the Switch version of Unity is patched to better support texture arrays, you are more than welcome to try this branch out. It isn't "clean" but should be serviceable enough for a demo.
> 
> Animated Waves and Depth work in this branch, but I haven't done so for the other LodData, if you could let me know which of the others you use I can focus on getting them working. However, this would be as a temporary fix for the upcoming build you have to deliver, we have another solution in the works that should fix the issues people have been having on their PCs whilst keeping texture arrays. Would that work for you in the long term once Unity has been fixed?

Hi Moosichu, we use foam and dynamics, a temp fix would be fine if something better is on the horizon.  Will give this branch a test as is, and see how it runs on switch.

huwb:
Closing. this was a super useful exercise and great to have on record, but i guess we're going to stick with TAs. If that changes lets reopen.

huwb:
Oh just read the last comments @dizzy2003 , sorry these fell by the wayside. i'm not sure this PR is a good place to track them, should we create an issue if they're still occuring?

its worth noting we're hoping to replace the way collision is done, as in #344 . this may impact the bug you mention.

Issue #310, opened 2019-08-02
Register ocean depth failing


huwb:
@moosichu as discussed on discord.

Repro: Create a large sphere and assign the Register Ocean Depth script, it won't input its depth to the system so the waves ignore it.

I see the depth cache renders using the GS, but the shader here is OceanDepths.shader which doesnt have GS, but i'm not really clear on what the solution is?

PR #317, opened 2019-08-08
Add support for LodDataInput shader options

moosichu:
So far this only extends to allowing for LodDataInput to exist which
support geometry shaders, but this could be extended further to support
compute shaders (for example), or even static textures directly (through
simplified interface). For now making this limited change in order to
fix a bug due to the switch to sea floor depth geometry shader support.


moosichu:
This should fix #310.

TBH the more I think about it, the more I'm personally inclined and willing to drop geometry shader support completely. *but* I think this particular increase in complexity is worth it anyway as it could be helpful for supporting further input types down the road like compute shaders which could allow for things like single-dispatch support there.

huwb:
I get a very interesting and profound image when i add a sphere and register it for depth:

image

Guessing the transforms are wrong?

crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
```diff
@@ -39,6 +40,58 @@ public class OceanDepthCache : MonoBehaviour
         GameObject _drawCacheQuad;
         Camera _camDepthCache;
 
+        public static bool UseGeometryShader
```
huwb: is UseGeometryShader specific to OceanDepthCache or is it now used by multiple lod types? If the latter we could put it perhaps in LodDataMgr..

crest/Assets/Crest/Crest/Scripts/LodData/RegisterSeaFloorDepthInput.cs
```diff
@@ -27,4 +27,31 @@ protected override void OnEnable()
...
+
+
+    /// <summary>
+    /// Tags this object as an ocean depth provider with geometry shader support (for advanced users).
+    ///
+    /// This is similar to RegisterSeaFloorDepthInput, but allows geometry depth to be rendered to all lod levels in a
+    /// single dispatch. The Ocean Depth Cache supports this mode of rendering, so use it's implementation as a
+    /// reference on how to do this.
+    /// </summary>
+    public class RegisterSeaFloorDepthInputGeometry : RegisterLodDataInput<LodDataMgrSeaFloorDepth>
```
huwb: i asked on discord but guess it makes sense to register here as well - could the presence of a GS be automatically detected from the shader, so the user doesn't have to configure and maintain a separate "Register" script type?

crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
```diff
@@ -15,26 +15,33 @@ public interface ILodDataInput
...
     {
         public abstract float Wavelength { get; }
+        public virtual LodDataInputType LodDataInputType => LodDataInputType.Conventional;
 
         public bool Enabled => true;
 
         public static int sp_Weight = Shader.PropertyToID("_Weight");
 
-        static Dictionary<System.Type, List<ILodDataInput>> _registrar = new Dictionary<System.Type, List<ILodDataInput>>();
+        static Dictionary<(System.Type, LodDataInputType), List<ILodDataInput>> _registrar = new Dictionary<(System.Type, LodDataInputType), List<ILodDataInput>>();
```
huwb: Cool to see tuples being useful!

moosichu:
Jacob Frye on Discord reported that this branch solves their problem, which is positive.

moosichu:
After a conversation with @huwb, we have decided that the extra complexity incurred by geometry shader isn't worth it, so we are going to remove it completely for now. Will create a new PR for this in a bit.

PR #320, opened 2019-08-17
Add one to the size of generated texture arrays

moosichu:
This will help prevent out-of-bounds texture array accesses in shaders.

### DIFF
--- a/crest/Assets/Crest/Crest-Examples/Shared/Scripts/OceanDebugGUI.cs
+++ b/crest/Assets/Crest/Crest-Examples/Shared/Scripts/OceanDebugGUI.cs
@@ -1,9 +1,6 @@
 ﻿// This file is subject to the MIT License as seen in the root of this folder structure (LICENSE)
 
-using Crest;
 using System.Collections.Generic;
-using UnityEngine;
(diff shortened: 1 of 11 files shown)


moosichu:
@huwb It probably makes sense for you to tackle this one as you have a clearer vision of what you want to do, just thought I would help by kick-starting the process.

huwb:
Oh sorry i think i haven't made the issue clear enough. the problem is with the arrays of uniforms, not with the TA. if slice index is at its maximum value (largest cascade), code like this will read off the end of the initialised data:

	const float wt_biggerLod = (1. - wt_smallerLod) * _LD_Params[_LD_SliceIndex + 1].z;

If all lods were active this would read off the end of the array.

i've pushed changes now which i think resolve the issue, let me know. i didnt see visual change on my kit, but it may have affected others.

huwb:
Weirdly it wont let me request your review (i can request my own review which is handy). can you review the changes? cheers

moosichu:
These changes look good to me. 👍

huwb:
Cool. I can't merge either so check it in whenever convenient!

PR #321, opened 2019-08-18
Remove geometry shader support completely

### LINKED ISSUE #310: Register ocean depth failing

moosichu:
After conversations with @huwb, despite the theoretical benefits of
a reduction in drawcall overhead, the maintenance costs aren't
currently worth it. Especially as many mobile platforms do not support
geometry shaders at all - and they are the ones which are the most
drawcall bound anyway.

Resolves #310.


moosichu:
It is always nice to have a commit which deletes a lot of code!

moosichu:
This replaces #317

cowhunt:
This is my favourite PR ever. Looks great to me. I ran a quick test with registering a depth input and it worked for me.

huwb:
This is my favourite PR ever. Looks great to me. I ran a quick test with registering a depth input and it worked for me.

PR #327, opened 2019-08-20
Generate error if collision queried before system init'd

### LINKED ISSUE #325: Height Sampling To Soon Stops returning proper height info.
Using Demo from latest build that includes the rowboat or you can add the rowboat prefab that has the BoatProbes script.

Move RB getcomponent to awake from start.

Call FixedUpdate once form awake.  System will only return 5 forever after that in FixedUpdate calls which is the height of the Y pos for the ocean.  

Also if you do not do the fixedupdate call so just setting everything back to default Demo with no script changes if you debug out the first x calls to fixedupdate you will notice you will get back a 0 for the height queries for the first few frames.  So again some timing issue is going on here that could potentially cause issues depending on what your doing with that height info.  So we need someway to know that the system is ready because making these bad calls.  

You can also add the following code to OceanRenderer.cs for ease.  Call this in any Awake function will also lock up everything.

        SamplingData _samplingData = new SamplingData();

        public float GetWaterHeightAtPosition(Vector3 position)
        {
            float height = 0;
            var r = transform.lossyScale.magnitude;
            var collProvider = OceanRenderer.Instance.CollisionProvider;
            var rect = new Rect(new Vector2(position.x - r, position.z - r), 2f * r * Vector2.one);

            if (collProvider.GetSamplingData(ref rect, 2f * r, _samplingData))
            {
                var pos = position;
                collProvider.SampleHeight(ref pos, _samplingData, out height);
                collProvider.ReturnSamplingData(_samplingData);
            }

            return height;
        }

huwb:
Fixes #325 .

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/LodData/Settings/SimSettingsAnimatedWaves.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/Settings/SimSettingsAnimatedWaves.cs
@@ -58,6 +58,7 @@ namespace Crest
                     break;
                 case CollisionSources.OceanDisplacementTexturesGPU:
                     result = GPUReadbackDisps.Instance;
+                    Debug.Assert(result != null, "Sampling collision too early, collision system has not been initialised.");
                     break;
                 case CollisionSources.GerstnerWavesCPU:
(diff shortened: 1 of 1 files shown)


huwb:
Thanks

PR #328, opened 2019-08-21
Add dyn waves from texture

huwb:
As discussed with @holdingjason on discord.

Adds a shader to inject dynamic waves from a texture (similar to add foam from texture).

Was almost straightforward, but its annoying the _SimDeltaTime needs to be set to integrate the forces into the sim correctly. An alternative approach would be to make it a global shader var which may save costs in some ways, but i felt it was unnecessary to go global with it.

Let me know what you think


huwb:
@moosichu bump :)

PR #329, opened 2019-08-21
Refactor static params

huwb:
Didnt really set out to do this, but started pulling on a thread
while doing something else and ended up with this. I think
moving these param IDs into the loddatamgr makes semantic
sense and reduces some of the code.

Also added two long over due IDs in oceanrenderer.

Let me know what you think..


huwb:
@moosichu , gonna get merged as its been open for a bit and it shouldnt be too controversial, hope thats fine. if you have feedback later do send through though. thanks!

PR #335, opened 2019-09-04
Implement Gerstner Waves using compute shaders

moosichu:
Implement Gerstner Waves using compute shaders instead of using pixel shaders.

This change should result in identical behaviour to what we had before, but is a prelude to a fix for #279 which will be the result of a change to texture formats, as we need to be able to output the gersnter wave results to multiple textures in parallel.


moosichu:
All TODOs should be marked as:
`TODO(TRC)`

There is some cleanup and refactoring left to do, but the functionality itself is ready to test.

moosichu:
Closing this as it wasn't needed for what we were doing in the end. It isn't a high priority issue.

PR #339, opened 2019-09-16
Aspect ratio of water camera fix

### LINKED ISSUE #257: Planar reflection and skybox issue
I have a planar reflection in the default main scene. I have switched to the default Unity skybox.

There are two issues:

1- The sun is off center. You can see it move in the water as you rotate the camera.

image

image

2- As sun drops below horizon, it continues to light the water from underneath causing a strange green appearance. I even tried adding a large cube underneath everything as a double sided shadow caster but it had no effect.

image

furkandinc:
Fix #257 
Tested on Unity 2019.2


moosichu:
Thanks for the contribution! :)

PR #340, opened 2019-09-16
Add support for texture slices for each anim wave channel

### LINKED ISSUE #279: Waves not forming properly, LOD and other strangeness

moosichu:
The second attempt to fix #279, but with less drastic changes.

https://docs.microsoft.com/en-us/windows/win32/direct3d12/typed-unordered-access-view-loads

As explained above *reading* from typed UAVs that aren't floats, uints or ints is unsupported in DX11 versions below DX11.3 - this was a problem as merging LODs required reading from an ARGBHalf typed UAV resource.

This change introduces the option of using a single float texture array slice for each displacement dimension - allowing older hardware to be supported. This does currently come with the following two caveats:

- It doubles the memory used (with an increase in associated bandwidth requirements) of animated waves when the fix is enabled.
- It won't work with GPU readbacks, however, this should no longer be an issue once #279 has been worked on.

We are going to wait and see if it fixes the issues people have been having, and if so we will get this merged.

It's worth noting that packing two halfs into a single in texture could be an option to save on memory requirements - but would add a lot of complexity and take time to complete. This workaround should work fine for now.

Tasks:
- [ ] Look into detecting if typed UAVs are supported on a given piece of hardware at runtime. (Could just check if DX11.2 or below?)
- [ ] Disable the workaround by default.
- [ ] Find a way to print a warning if waves don't show? (Or throw it in validation somehow).
- [ ] Investigate the best way to toggle WORKAROUND_UAV_LIMITATION in the shaders (is there a way to do this in code?). (Marked `TODO(TRC)`)

### DIFF
--- a/crest/Assets/Crest/Crest-Examples/Main/Data.meta
+++ b/crest/Assets/Crest/Crest-Examples/Main/Data.meta
@@ -1,0 +1,8 @@
+fileFormatVersion: 2
+guid: fa180ad7bf100dd478bb36448899d289
+folderAsset: yes
+DefaultImporter:
+  externalObjects: {}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 
--- a/crest/Assets/Crest/Crest-Examples/Main/Data/AnimWaveSettingsMain.asset
+++ b/crest/Assets/Crest/Crest-Examples/Main/Data/AnimWaveSettingsMain.asset
@@ -1,0 +1,19 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!114 &11400000
+MonoBehaviour:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  m_GameObject: {fileID: 0}
+  m_Enabled: 1
+  m_EditorHideFlags: 0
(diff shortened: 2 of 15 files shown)


moosichu:
It looks like there isn't a way to query supported UAV types directly (unless we make a native plugin?) But we could query if DX11.2 or below? https://docs.unity3d.com/ScriptReference/SystemInfo-graphicsDeviceVersion.html

huwb:
@moosichu . It seems like we're taking an alternative path so closing for now. Let me know if this is/isn't accurate, if so we can reopen.

moosichu:
I agree, it's a path worth considering once compute height queries is stable, but has too many issues for now.

PR #341, opened 2019-09-17
Ping pong combine

huwb:
Another variation of a fix for #279 . I think this might be a 'winner' (or at least the best loser) - doesnt add complexity to the code beyond the minimum (a new combine code path).

Adds a few texture copies, but avoids a few downsides of #340 .

I've lazily left both code paths in the runtime. This has an overhead of loading an additional shader, and allocating an additional widthXheightX1 texture. To fix this we could move a bunch of stuff into defines potentially. Might be a good idea to deploy it as a dynamic switch which might help with debugging and then bake it into a define later perhaps? Or just leave it dynamic.


huwb:
Awesome. Agreed about write up, if i get time ill add it to #279

PR #343, opened 2019-09-25
Fix the UAV type of shape combine texture array

moosichu:
_LD_TexArray_AnimatedWaves_Compute is now half4 instead of float4
to match the type of the underlying texture.

### DIFF
--- a/crest/Assets/Crest/Crest/Shaders/Resources/ShapeCombine.compute
+++ b/crest/Assets/Crest/Crest/Shaders/Resources/ShapeCombine.compute
@@ -21,7 +21,7 @@ float _HorizDisplace;
 float _DisplaceClamp;
 float _CrestTime;
 
-RWTexture2DArray<float4> _LD_TexArray_AnimatedWaves_Compute;
+RWTexture2DArray<half4> _LD_TexArray_AnimatedWaves_Compute;
 
 void Flow(out float2 offsets, out float2 weights)
 {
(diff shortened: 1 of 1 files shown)


moosichu:
I realise that we tested if this fixed the UAV issues (it didn't), but it is probably a good thing to checkin regardless!

cowhunt:
Thanks.

I don't know if it's worth lumping into this pr but there's a bit more fixup we should do with the you're of the slice index. It's using floats to index into arrays so there will be conversions from float to int happening a bunch. I think we should try changing to int starting from the sample functions and can probably avoid conversions in a bunch of places. Just mentioning it as if we're looking at types this would be a low hanging fruit that would be good to hit.

huwb:
Thanks.

I don't know if it's worth lumping into this pr but there's a bit more fixup we should do with the you're of the slice index. It's using floats to index into arrays so there will be conversions from float to int happening a bunch. I think we should try changing to int starting from the sample functions and can probably avoid conversions in a bunch of places. Just mentioning it as if we're looking at types this would be a low hanging fruit that would be good to hit.

moosichu:
We could do, but it's worth pointing out the the desired type is *uint*, not *int*.

moosichu:
Done, is this good? :)

huwb:
I tackled some remaining type issues separately. With those and this PR the types are in pretty good shape. There's still a few integer to float conversions but some go into SampleLevel as a float3 and didnt feel worth trying to eradicate.

PR #344, opened 2019-09-26
Height queries in compute shader

huwb:
**Status:** ready to merge pending final tests.

To test: open any scene with boats. the default collision provider is now the new compute-based system.

This change adds a new, default collision source to the animated wave settings - compute queries.

These upload query positions to the GPU, which are then executed in a compute shader, and the results are asynchronously transferred back to the CPU.

This is a huge improvement over the previous system, the memory bandwidth is reduced massively. There used to be a high CPU cost to copying out the readback data (up to 3-4ms in some cases) which is gone. The complexity of choosing which LOD to sample is gone as well, its all done dynamically.

To facilitate batching queries and sending them off to the GPU, there are Query() functions added to ICollProvider which receive arrays of query positions. The other ICollProvider functions are now deprecated. The new Query() functions mirror what is done in the gerstner height job branch, so this approach could be added naturally to the new APIs in the future.

**Note**: The old flow texture readback has been removed. For implementation reasons it was not easy to support old and new paths for querying flow info, so it has been hard switched to the new path.


dizzy2003:
I tried this on ps4, the boats behaved very badly, kinda floating off into the sky

(I specifically tried this on ps4 as the current readyback async is majorly blocking ps4 speed)

dizzy2003:
seemed ok on pc, running on an amd 2200G

dizzy2003:
btw on ps4 the small cubes are working, but am not sure if they are using the compute heights system.

huwb:
Thanks @dizzy2003 I also saw some instabilities which I need to track down next.

huwb:
@dizzy2003 there was a fairly nasty issue which would make boats pop around which i fixed now.

hopefully this is now good to go.

in this branch, the compute-based system is now the default collision provider and so will be turned on in every scene, unless you have Animated Waves Settings which select one of the other collision types.

if you get a chance to test this I'd be keen to hear how it fares.

huwb:
Merged into master

PR #346, opened 2019-09-29
Add a proxy monobehaviour for the post-process effect

moosichu:
This means we can enable/disable post-processing from the main update loop based on runtime parameters.


huwb:
Gotcha. I think my thoughts are the same - we should explore using an event. This may be simpler, and avoids the overhead of the update. What do you think?

huwb:
The event based solution seems to have worked so going with that to avoid the update, and parking this.

PR #348, opened 2019-10-05
Make shadows work(ish) for underwater

huwb:
Now that we have TAs, its easy to compute proper UVs and sample shadows at any world position, which works better.

Two problems:

* They're noisy because of the jittered sampling of the shadows. Not
sure what to do about that. Could average over multiple samples..
* Moving backward in shadowed water will sample unpopulated data
which will be super bright. Fix could be to offset sample along camera
forward vector.

I guess this should be cleaned up and merged in as it fixes the worst of the strobing, but is no silver bullet.


huwb:
I've merged this into the underwater branch so that i could work with it there. We could close this if/when the former gets merged into master.

huwb:
Merging in as underwater branch is deep in rewrite land.

PR #352, opened 2019-10-11
Move Foam Light Scale under Foam header

daleeidd:
Foam Light Scale isn't specific to Foam 3D Lighting. It is also applied to foam generally on this [line](https://github.com/crest-ocean/crest/blob/master/crest/Assets/Crest/Crest/Shaders/OceanFoam.hlsl#L80).


huwb:
Nice one, thank you

Issue #350, opened 2019-10-08
Caustics - support colour separation


huwb:
From discord user americanbean

> I've been playing with caustics to see what sort of effects can be achieved. I don't know how difficult it would be to add but, I'd be nice to have control over an "Emmissive" value for the  Caustics texture to ramp those values in HDR space. Also, I edited the original caustics texture to get some color separation. However, color information seems to be disregarded in the caustics projection. Can this be include in the future? Right now, I can use bloom, but this ramps up all values universally which isn't ideal. here's an example: https://youtu.be/kB3O7NZJMvo

![image](https://user-images.githubusercontent.com/939901/66377932-6b92b680-e9aa-11e9-8065-cc687694da8e.png)

huwb:
Merged #355

PR #355, opened 2019-10-13
#350 - Use coloured caustics texture

huwb:
As suggested by @americanbean

@americanbean are you ok with us integrating this coloured caustics texture into Crest?


americanbean:
@huwb yes, no problem. I can generate a couple of other options. Perhaps one that’s a happy medium.

huwb:
Thanks @americanbean ! If you do make a happy medium version let me know. For now i pushed the last one you provided.

Issue #351, opened 2019-10-10
Spectrum power interpolation doesn't respect disable option when calculating amplitude


daleeidd:
**Describe the bug**
I am not certain if this is a bug or intended behaviour. This [line in the source](https://github.com/crest-ocean/crest/blob/master/crest/Assets/Crest/Crest/Scripts/Shapes/OceanWaveSpectrum.cs#L107) interpolates between the current power and the next higher power, but it doesn't check if the next power is disabled or not.

**Additional context**
There are three options I can think of:
1. Leave as is. This is expected behaviour
2. Find the next highest power that is enabled
3. If the next power is disabled, skip interpolation

The current option (1) can be a little confusing when debugging or authoring. I was wondering why a disabled slider was still affecting the spectrum.

I was going to do a PR, but I was not sure which option you would prefer.

huwb:
Thanks for IDing this and for talking it through.

Is there an option 4 where the disabled flag forces the corresponding power down to the minimum value. So it will be as if the slider is set to the far left, regardless of its actual value. What do you think?

daleeidd:
Option 4 sounds good. There is a PR waiting.

PR #356, opened 2019-10-14
Spectrum power lerp should check disabled flag

daleeidd:
Resolves: #351

The aim is to reduce confusion when authoring or debugging since before
the fix the disabled slider would still affect the spectrum. Option 4
out of the following options was chosen:

1. Leave as is. This is expected behaviour
2. Find the next highest power that is enabled
3. Skip interpolation
4. Use the minimum power

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Shapes/OceanWaveSpectrum.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Shapes/OceanWaveSpectrum.cs
@@ -84,10 +84,8 @@ namespace Crest
                 return 0f;
             }
 
-            if (_powerDisabled[index])
-            {
-                return 0f;
-            }
+            // Get the first power for interpolation if available
+            var thisPower = !_powerDisabled[index] ? _powerLog[index] : MIN_POWER_LOG;
 
             // Get the next power for interpolation if available
             var nextIndex = index + 1;
@@ -109,7 +107,7 @@ namespace Crest
             var alpha = (wavelength - lower) / lower;
 
             // Power
-            var pow = hasNextIndex ? Mathf.Lerp(_powerLog[index], nextPower, alpha) : _powerLog[index];
+            var pow = hasNextIndex ? Mathf.Lerp(thisPower, nextPower, alpha) : _powerLog[index];
 
             var a_2 = 2f * Mathf.Pow(10f, pow) * domega;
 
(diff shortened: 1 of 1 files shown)


huwb:
@daleeidd while reading the code i realised the same problem exists on the lower side of the lerp (but is probably masked by the code early-outing and not including any wavelengths at all if the lower side is disabled). Does my additional commit look good to you? If so I'll get it merged. Thanks again!

crest/Assets/Crest/Crest/Scripts/Shapes/OceanWaveSpectrum.cs
```diff
@@ -104,7 +107,7 @@ public float GetAmplitude(float wavelength, float componentsPerOctave)
             var alpha = (wavelength - lower) / lower;
 
             // Power
-            var pow = Mathf.Lerp(_powerLog[index], _powerLog[Mathf.Min(index + 1, _powerLog.Length - 1)], alpha);
+            var pow = hasNextIndex ? Mathf.Lerp(thisPower, nextPower, alpha) : _powerLog[index];
```
huwb: After pushing my last change, I've just noticed the hastNextIndex check here. Is this required? I was thinking that the two powers should always be lerped, but one or both of them are pushed down to min if disabled. So disabling a slider forces it to min power, but nothing else changes. To get to the point, can we remove the ternary on hasNextIndex from this line?

daleeidd: It was to not interpolate anything for the last slider (512) since there is no next power. So you want interpolation with `MIN_POWER_LOG` for that case too? Before it was interpolating power 512 with power 512.

daleeidd:
@huwb Sorry. I don't understand the use of the min value for the current index since that doesn't cause a discrepancy between the wave shape and the inspector UI (as you stated by the problem being masked). I am guessing you are trying to solve a different problem or improvement. Other than that, the commit is good. I am happy to amend mine, and retest all changes, once feedback is given on my response so it can be merged :^)

crest/Assets/Crest/Crest/Scripts/Shapes/OceanWaveSpectrum.cs
huwb: You're right, it was confusion on my part. I mistakenly thought hasNextIndex included the check for the disabled flag, but its purely based on index, so i think we can disregard this

huwb:
Merged, nice one!

PR #360, opened 2019-11-01
Split Crest from Crest Examples using asmdef

moosichu:
Splits project into assemblies that have well-defined dependenices. This
will speed up the compile-time of things (only dependents have to be
recompiled) and will prevent us from accidentally calling-into example
code from the project.

Find more info here: https://docs.unity3d.com/Manual/ScriptCompilationAssemblyDefinitionFiles.html


huwb:
Thanks dude!

Could you talk about the granularity in the examples - im wondering if only examples/scripts needs an asmdef, but interested to hear your thoughts.

Can you give a view on the below - I guess the vast majority of our users wont have asmdefs set up. Does it matter?

> Note: Although it’s not required, whenever you use Assembly Definitions in a Project, you should do so for all of the code in your Project. Otherwise, when you change scripts in one of the predefined assemblies, Unity has to still recompile all the code in your Project, because the predefined assemblies automatically depend upon any assemblies you create using an Assembly Definition.

moosichu:
> Could you talk about the granularity in the examples - im wondering if only examples/scripts needs an asmdef, but interested to hear your thoughts.

As Crest is a fairly small project code-wise, I don't think much granularity was needed beyond "put Crest in it's own assembly, and put each of the examples in their own asssemblies". Our existing folder structure was very amenable to this, I didn't need to update any code at all beyond make the helper height sampling classes public.

> Can you give a view on the below - I guess the vast majority of our users wont have asmdefs set up. Does it matter? (Should there be a doc update somewhere to recommend people use them?)

Unity consider it best-practise for larger teams. But if they don't have asmdefs set-up this won't affect their workflow at all - the biggest change will be that when they make a change to their code, the crest code won't be recompiled which is a nice bonus.

People already using asmdefs will need to add crest as a reference in them, but considering they are already using them they should know how that works.

PR #362, opened 2019-11-06
Adaptor to dynamically batch one-at-a-time height queries

huwb:
Usage - create an adaptor in the code where one-at-a-time queries need to be serviced:

`DWPAdaptor _adaptor = new DWPAdaptor();`

Execute queries to get heights:

`float waterHeight = _adaptor.GetWaterHeight(position);`

When done writing queries for a frame, call the update function:

`_adaptor.Update();`

This could be called from Update() of a monobehaviour (if queries will come through during FixedUpdate), or from LateUpdate. It is probably not critical when exactly it's called, as long as it's called exactly once per frame and in a consistent way.

This is intended specifically to proxy non-batched queries into the system. We don't recommend this for general purpose use - for example if a single height needs to be queried, use `SampleHeightHelper` which is simpler, more direct and more efficient.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Collision/DWPAdaptor.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Collision/DWPAdaptor.cs
@@ -1,0 +1,57 @@
+﻿// Crest Ocean System
+
+// This file is subject to the MIT License as seen in the root of this folder structure (LICENSE)
+
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace Crest
+{
+    /// <summary>
+    /// Tries to dynamically batch height query points on the fly. 
+    /// </summary>
+    public class DWPAdaptor
+    {
+        class QueryToken { }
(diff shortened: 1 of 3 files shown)


crest/Assets/Crest/Crest/Scripts/Collision/DWPAdaptor.cs
```diff
@@ -0,0 +1,57 @@
...
+
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace Crest
+{
+    /// <summary>
+    /// Tries to dynamically batch height query points on the fly. 
+    /// </summary>
+    public class DWPAdaptor
```
moosichu: Out of curiousity, what does "DWP" stand for?

huwb:
In the (near) future the DWP2 authors will connect directly to the compute-height-queries. This adaptor is no longer useful or necessary. Closing.

PR #365, opened 2019-11-18
Fix per-frame GC from OceanPlanarReflection

huwb:


huwb:
Merging after testing by Jason

Issue #358, opened 2019-10-30
Invalid sampling data - LOD to sample from was unspecified.


controlaware:
**Describe the bug**
GPUReadbackDisps.cs :: ComputeUndisplacedPosition(...)

Using Dynamic Water Physics 2, the DWP collision provider queries the Crest Ocean collision provider for information. It seems like when the LOD changes, the `PerLodData` tag is not filled in. This means that the function mentioned above throws null-reference errors and the collision does not work anymore. DWP doesn't get the height collision and the object falls through the ocean.

**Log**
GPUReadbackDisps.cs, `NullReferenceException` thrown on line:

`for (int i = 0; i < 4 && lodData._resultData.InterpolateARGB16(ref guess, out disp); i++)`

Debug log shows the `Debug.Assert` text found at the beginning of the method. Call stack is:

DWP2.CrestWaterDataProvider:GetWaterHeight
  Crest.CollProviderCache:SampleHeight
    Crest.GPUReadbackDisps:SampleHeight
      Crest.GPUReadbackDisps:ComputeUndisplacedPosition

Adding a check under the assert removes the NullReferenceException error:

`            
            Debug.Assert(i_samplingData._tag != null, "Invalid sampling data - LOD to sample from was unspecified.");

            var lodData = i_samplingData._tag as PerLodData;

            if (lodData == null)
            {
                undisplacedWorldPos = i_worldPos;
                return false;
            }`

But obviously the `Debug.Assert` is still there, which spams the Debug Log window (999+ in a second or two). 

**Versions**
Built-in render pipeline, Unity 2019.2.9f1, Crest-8.0.0 unity package. 

**To Reproduce**
Create Crest ocean according to quick-start guide. From DWP2, drop in the "TorpedoBoat" prefab. Add a constant relative force of -320000 in the Z direction. Attach a movable camera script and move camera as boat moves panning out then back in. After the boat moves from one LOD to another, the boat will fall through the ocean.

**Platform**
 - Editor
 - Windows 10 Pro 1903

**Hardware**
AMD Ryzen 7 2700X, 32GB RAM, AMD Radeon RX580 8GB software version 19.9.2, not overclocked.

huwb:
Hi,

I've seen scattered reports of this, but never got a good repro.

On the latest version, we changed out this system for a new one which is faster and simpler and completely sidesteps this issue. In the future we will completely remove the displacement readback system.

[...]

PR #366, opened 2019-11-20
Fix DWP2 Invalid sampling data error

furkandinc:
Fix for #358 
DWP2 requires sampling data tag should not be null. But not every lod level data returns a valid Rect to compare the position.  So, returning the last lod level with a valid Rect data while containing the `lastCandidate` constraint is a tricky fix for the issue.


huwb:
Thanks for this. I added a commit where i ran format doc to make the whitespace comply with teh editorconfig.

As an aside, we hope to remove this collision provider - the readback one - from a future version. It is superceded by the compute-height-queries provider. The DWP2 author will connect to this provider.

Regardless this is still a good fix, merging.

PR #372, opened 2019-12-11
Added Unity PBS Reflection Probe Sampling

HelloKitty:
Not sure this is desireable, but if it is this may be helpful in reflections.

Commit Description:
This probe sampling code is based on the Unity PBS standard shader and interpolation code suggested by https://catlikecoding.com/unity/tutorials/rendering/part-8/.

It should blend probes, which I think was not taking place before, and also it should make reflection smoothness depend on the Ocean specular parameter.

### DIFF
--- a/crest/Assets/Crest/Crest/Shaders/Ocean.shader
+++ b/crest/Assets/Crest/Crest/Shaders/Ocean.shader
@@ -55,6 +55,8 @@ Shader "Crest/Ocean"
 		[Header(Reflection Environment)]
 		// Strength of specular lighting response
 		_Specular("Specular", Range(0.0, 1.0)) = 1.0
+		// Controls blurriness of reflection
+		_Roughness("Roughness", Range(0.0, 1.0)) = 0.0
 		// Controls harshness of Fresnel behaviour
 		_FresnelPower("Fresnel Power", Range(1.0, 20.0)) = 5.0
 		// Index of refraction of air. Can be increased to almost 1.333 to increase visibility up through water surface.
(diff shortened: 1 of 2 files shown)


HelloKitty:
For reference the standard shader source it's based on is in UnityGlobalIllumination UnityGI_IndirectSpecular function: https://github.com/TwoTailsGames/Unity-Built-in-Shaders/blob/bab22c6c04df30e674a47f225e946072c86222cb/CGIncludes/UnityGlobalIllumination.cginc#L164

huwb:
Thanks, this is useful!

I tried it and it seems to work for me. I added a commit to bring the whitespace in line with how its done in the rest of crest. actually i like the way you had it better, but most of all we want consistency. i'll contemplate moving to the style you had.

I'm not 100% sure about connecting the smoothness to Specular, but if you like it this way then I'm happy to go with it?

If you confirm that then I'll merge it.

HelloKitty:
@huwb Opps, I didn't notice the standard I apologize. Hmmm, I opted to use an existing input without adding one but I guess the metal/specular PBS shader in Unity3D separates specularity and smoothness of the reflection as separate inputs. It's up to you if you want to add that input. Maybe there are some analytical ways to compute ocean reflectivity/smoothness too but that kind of math is beyond me.

Also, this commit doesn't solve the discontinuity of reflection probes at ocean tile seams sadly. Even though it seemed like it should have. So there must be something else wrong with that particular thing. But hopefully it'll help it work once whatever else is causing that issue clears up.

Example:
https://i.gyazo.com/c7689f43d8dd00e5231d280656891409.jpg

Oh, and I also made a typo. I don't know what it affects but let me correct it.

huwb:
The discontinuity is usual caused when tiles are outside the probe extents? Is that the case here?

huwb:
Let me know if the probe extents was the issue.

I think it does make sense to expose roughness/smoothness separately. I've done so in my latest commit. Let me know what you think.

huwb:
Bumping question about the tile issue.

nkevin90:
basically what i was looking for for the completion of our project

skaughtx0r:
This PR seems to work great for me. Fixes some issues I was having with the reflections from probes being too intense.

huwb:
Thanks for the confirmations. Merging.

PR #375, opened 2019-12-17
Minor fixes

daleeidd:
Sorry if having a few small fixes in a single PR is an issue. It was just the most efficient way. I can redo as separate PRs if desired.

Incomplete refactor: `parms` was years ago
`IndexOutOfRangeException` was due to "slice index + 1" so adding an extra element is correct I believe
`_SubSurfaceCrestColour` was from the now removed height lerp subsurface colouring

### DIFF
--- a/crest/Assets/Crest/Crest-Examples/Whirlpool/Materials/Ocean-Flow.mat
+++ b/crest/Assets/Crest/Crest-Examples/Whirlpool/Materials/Ocean-Flow.mat
@@ -131,7 +131,6 @@ Material:
     - _Specular: 1
     - _StartLevel: 0.157
     - _SubSurfaceBase: 0.24
-    - _SubSurfaceCrestColour: 1
     - _SubSurfaceDepthMax: 9.4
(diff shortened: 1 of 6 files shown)


huwb:
Thanks. Looks good, but the index one is very curious.

We intentionally have an extra element in the array to simply the shaders and logic. The last element is duplicated. Are you getting an IndexOutOfRangeException? In what circumstances?

daleeidd:
When I set the LOD Count to 15, whilst experimenting, which is the maximum amount. Sorry, I should have provided these (where the exceptions are thrown):
https://github.com/crest-ocean/crest/blob/master/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgr.cs#L132
https://github.com/crest-ocean/crest/blob/master/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgr.cs#L133

huwb:
Ahh gotcha. Tested here and your fix is perfect. Merging!

huwb:
Thanks again!

PR #377, opened 2019-12-19
Ocean Debug GUI optimisations

### LINKED ISSUE #277: Create custom system for drawing debug texture arrays
In `OceanDebugGUI.cs` we currently use Unity's [`GUI.DrawTexture`](https://docs.unity3d.com/ScriptReference/GUI.DrawTexture.html) functionality to draw LODs to the debug overlay. However, this function doesn't support drawing texture arrays directly, meaning we have to copy each LOD slice to a seperate texture before drawinng it to the screen - an unnecessary overhead.

Ideally we would have a single shader which provides this functionality for us, either with the ability to specificy which LOD we want to draw to the GUI or just the ability to draw all of them in a single draw-call (or CS dispatch?).

If anyone takes this on as their first issue, I'm more than happy to guide them through the process.

daleeidd:
Resolves #277. Not my first issue, but I am very green when it comes to these things so I thought I would give it a try. Followed the following:
https://docs.unity3d.com/Manual/SL-TextureArrays.html

I went with the simplest approach which is a shader to render a specific depth of the texture array. I placed it under `Crest-Examples` since that is where the `OceanDebugGUI` lives.

I made some other optimisations and a minor design change. If you want to use the old layout, then I can easily change it back (but still have it optimised).

Before:
Before

After:
After

Tested on MacOS with 2018.4 and 2019.2

### DIFF
--- a/crest/Assets/Crest/Crest-Examples/Shared/Scripts/OceanDebugGUI.cs
+++ b/crest/Assets/Crest/Crest-Examples/Shared/Scripts/OceanDebugGUI.cs
@@ -11,6 +11,8 @@ public class OceanDebugGUI : MonoBehaviour
     [SerializeField] bool _showSimTargets = false;
     [SerializeField] bool _guiVisible = true;
     static float _leftPanelWidth = 180f;
+    static float _bottomPanelHeight = 25f;
+    static Color _guiColor = Color.black * 0.7f;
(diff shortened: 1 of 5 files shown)


huwb:
It all looks completely perfect to me, and works for me. Great job. Merging!

Issue #364, opened 2019-11-17
Baked depth cache


huwb:
Instead of always dynamically generated depth cache, could bake it offline, perhaps as an editor tool.

Pros: Simpler/more intuitive. Less prone to bugs. Gets around the instanced terrain issue.
Cons: Potentially large high precision texture (disk space concern). Needs rebake/regen when world changes.

I had an attempt at this but could not get it to load/save correctly: https://github.com/crest-ocean/crest/commit/742f62cfc0c42485083eb75b23406d4a05d47570

Discord user 'snacktime' also had a stab at this: https://gist.github.com/gamemachine/a83671661c867417c26820fe32e82562

This issue is to track this approach and see if we can get it shippable.

daleeidd:
What was the issue both of you encountered? I have tested this on 2019.2.16f and it works. I tried it on 2018.4.0f1 and the texture saved squished horizontally. This is on MacOS. I did a site search through Unity's release notes and found a few fixes for MacOS for ReadPixels, but nothing pertinent for Windows. Not sure what version this was fixed in.

Live depth cache:
![Depth Cache Live](https://user-images.githubusercontent.com/5249806/71170682-f13ab880-22af-11ea-85da-4151e5dd1ad8.png)

Saved depth cache:
![Depth Cache Saved](https://user-images.githubusercontent.com/5249806/71170687-f6980300-22af-11ea-86d2-a0dbb4c87a4b.png)

The saved one is a bit blurry because of resolution setting when saved.

huwb:
Interesting. I get this:

![image](https://user-images.githubusercontent.com/939901/71269853-2c88c600-2348-11ea-9b9f-54405c98ef9f.png)

One potential thing is that the depth cache was probably populating on starttup, even if you specified a saved cache. I've just pushed a change to that branch which makes it not render anything if a cache was specified.

Also, the "save" button was visible even when outside of play mode, when there was no data in the cache. It may not have actually saved any useful data in this case. I've made it so you can only save while in play mode as well.

daleeidd:
But that function also populates the cache from the texture. Here is more substantial proof that it works (no island):

Without texture:
![No cache](https://user-images.githubusercontent.com/5249806/71299322-ee8bb080-23e0-11ea-8e71-a97477104985.jpg)

With texture:
![Cache](https://user-images.githubusercontent.com/5249806/71299312-da47b380-23e0-11ea-9f00-d0b253f8e3a8.jpg)

[...]

PR #378, opened 2019-12-22
Baked depth cache improvements

daleeidd:
Continues on from issue #364

Overwriting is for quality of life. Otherwise, you would have to remove the texture property, bake, replace and finally delete the old texture. It can track path changes so users can rename GUID name if desired. GUIDs are used to avoid conflicts.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
@@ -4,6 +4,7 @@
 
 // This is the original version that uses an auxillary camera and works with Unity's GPU terrain - issue 152.
 
+using System;
 using UnityEditor;
 using UnityEngine;
 
@@ -310,7 +311,8 @@ namespace Crest
                 byte[] bytes;
                 bytes = tex.EncodeToEXR(Texture2D.EXRFlags.OutputAsFloat);
 
-                string path = dc._savedCache ? AssetDatabase.GetAssetPath(dc._savedCache) : $"Assets/{target.name}.exr";
+                string path = dc._savedCache ? 
+                    AssetDatabase.GetAssetPath(dc._savedCache) : $"Assets/OceanDepthCache_{Guid.NewGuid()}.exr";
                 System.IO.File.WriteAllBytes(path, bytes);
                 AssetDatabase.ImportAsset(path);
 
--- a/crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
@@ -16,9 +16,25 @@ namespace Crest
     /// </summary>
     public class OceanDepthCache : MonoBehaviour
     {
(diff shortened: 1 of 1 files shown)


huwb:
Thanks for this!

I'm not 100% sure the best process here - tbh i'd happily merge this back right now since its going into a dev branch - but ill comment for now.

I mulled over this for a bit. I think we're on the right track but I've come to realise that "Populate on startup" is not logically equivalent to "bake cache on the fly". There are situations where a user may want a dynamic-rendered cache, but they don't want it populated on startup. They may populate it later on once all the sublevels have streamed in, etc.

In my code so far, if a saved cache is provided, it uses that, otherwise it relies on populate on startup, or it relies on the user populating the cache at some point. I think this may work, but it may not be the clearest thing in the world as the behaviour is implicit and it requires this knowledge/context.

One idea might be to have a drop down box to select the 'mode' of the depth cache, similar to how a reflection probe has different update modes. Then the user could either set it to e.g. "populate at runtime" or "baked offline". This would make things a bit more explicit?

daleeidd:
I think I understand. An enum called mode:
- "Populate at runtime" would be same behaviour as populate at startup?
- "Baked offline" would use the saved cache data?
- "Scripted" or "Manual" so developers can call it themselves?

huwb:
Yeah that was my line of thinking.

Just took a look at reflection probe and it seems to have relevant modes, might be good to align with this wherever it makes sense:

image

daleeidd:
Oh right. I was looking at light probes. Oops. I will update the commit.

daleeidd:
Hopefully I have interpreted what you proposed correctly. I had to make some other changes so that it functioned as per selected options.

Also, feel free to go ahead and make any changes you want if it is more efficient than explaining what you want to me. I appreciate efficiency and I don't want to get in the way of that. I won't take it personally.

huwb:
Tried it out - seems to work well, great stuff!

I decided to try hiding the refresh mode if cache set to baked. It kind of escalated from there, it turns out almost every setting is for realtime and when set to baked it now collapses to this:

image

I dont love the complexity the new code adds but i think its probably ok. Let me know what you think? Feel free to experiment and change.

FYI EveryFrame is debug only because i dont know theres any point to update a cache each frame - if one wants a changing depth they can assign a "Register Ocean Depth Input" component to draw straight into the depth data and skip the middleman. So i dont think it should be added.

daleeidd:
Works great. Haha, yeah, I was trying to avoid going down that rabbit hole. Wish Unity would provide more attributes.

The way the quad was being initialised had it rendering a square depth into the ocean for on demand. Fixed.

I also added a button to populate the cache for on demand. Only a small bit of code but it is useful for testing etc.

huwb:
Super cool!

Made a couple of small additions myself.

So I think this is in good shape. The defaults should mean it continues working for existing users, so i'm fairly confident deploying this. Do you plan to make any other changes or finalisation?

daleeidd:
Neat. Nope, all good from me.

huwb:
All merged into master! Thanks this is a great contribution.

PR #380, opened 2019-12-24
Ocean Debug GUI housekeeping

daleeidd:
TextureArray shader
- Use code style similar to other shaders
- Use unity_ColorSpaceDouble to brighten textures
- Use newer UnityObjectToClipPos
- UV only needs to be float2

OceanDebugGUI
- Make bottom bar full width if sidebar hidden

### DIFF
--- a/crest/Assets/Crest/Crest-Examples/Shared/Scripts/OceanDebugGUI.cs
+++ b/crest/Assets/Crest/Crest-Examples/Shared/Scripts/OceanDebugGUI.cs
@@ -167,7 +167,8 @@ public class OceanDebugGUI : MonoBehaviour
 
         // Draw bottom panel for toggles
         GUI.color = _guiColor;
-        GUI.DrawTexture(new Rect(_leftPanelWidth, Screen.height - _bottomPanelHeight, Screen.width, _bottomPanelHeight), Texture2D.whiteTexture);
+        GUI.DrawTexture(new Rect(_guiVisible ? _leftPanelWidth : 0, Screen.height - _bottomPanelHeight, Screen.width,
+            _bottomPanelHeight), Texture2D.whiteTexture);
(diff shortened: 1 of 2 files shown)


huwb:
Wow I hadn't come across unity_ColorSpaceDouble, feel like i should really know about that..

So this should be used even with render textures that are not originating from sRGB images?

daleeidd:
I'm not 100% on the correct usage. I was following along to this [article](https://catlikecoding.com/unity/tutorials/rendering/part-3/). And I thought maybe this is what Unity does for GUI textures by default. The value is similar to what I was already doing (x 4), which I had to do since we aren't using a GUI texture anymore, and they were really dark.

But now that you point it out, I am doubting that I am using it correctly. Maybe I should be looking into UnityUI instead.

daleeidd:
It seems like it is mainly used for detail textures and skyboxes going by the Unity's shaders.

huwb:
Yeah i also ended up at that article. If i understood correctly that change will be doubling the image intensity when rendering in Linear? If so i dont think we want that..

daleeidd:
It does:
```
#define unity_ColorSpaceDouble fixed4(4.59479380, 4.59479380, 4.59479380, 2.0)
```

daleeidd:
This is what it would look like without any change to the intensity:
Dark

It's not too bad, TBH. And more details do come back

huwb:
Yeah.. i think some of the data is already blown out in the viz so I'd vote for not multiplying them up further.

daleeidd:
No worries. I will fix this up shortly

daleeidd:
Done. Thanks for the feedback

huwb:
Nice one! Had an offline thought - could you ping me on discord or mail me: huw.bowles@gmail.com

Issue #357, opened 2019-10-22
Receive illumination on the water surface of spot and point Lights


ftejada17:
Simply what I want is that the surface of the water receives the illumination of other sources of light, like for example the spotLight or pointLight.

Here I leave a video of the Ultimate Water System asset in which it looks like it receives the light of a SpotLight. Where it gives light, under the ocean there is NO land.
https://youtu.be/LUnDvCD0i2g
And in this video I show Crest in a similar situation. It looks like Crest simply doesn't seem to receive SpotLight light or pointLight as already mentioned.
https://youtu.be/3vT9wBNUZOk

I have no experience in Shaders programming and I find it impossible to implement something so that Crest has this feature. Anyway I think it is something quite important that you should value if you put natively in Crest.

Another thing that I have not been able to test properly, is if under water the light is absorbed by the water according to the depth ...
If not, it should be included as it is essential.

Regards

huwb:
Thanks for adding this, good to have it tracked.

I'm not sure if/when i'll have bandwidth to put time into the built-in RP sadly, but i've put the help wanted label on this, perhaps someone from the community can contribute the solution.

daleeidd:
I tried adding an extra pass as an experiment to add point and spot lights, but it seemed to be the wrong way to go about this since the light won't be applied across tiles.

I believe the way to accomplish this is to look at how shadows are done. It would require using a compute shader to write the light data into a texture array. Then the texture array can be sampled in the ocean shader. An RGB texture array could be used to capture the light's colour.

The shadow compute shader was written by looking and reimplementing code from `UnityCG.cginc` since we don't have geometry to work with. It might be the same case for lighting.

The directional light (sun) doesn't need all of this because it is everywhere and has no boundaries (those are shadows which is compute). Point lights and spot lights have a limited range.

The steps would be:
1. Architecture (LOD Manager etc) **Easy**
2. Compute shader (use OceanGUIDebug to view output) **Hard**
3. Sample data in shader **Medium**

@huwb would this be correct?

adamjm123:
I'd love to see this feature also!

huwb:
Thanks @daleeidd!

That is an odd artifact.. hmm.. it doesnt immediately look like its aligned ot the geometry tiles, although it might be. The ocean verts are displaced so i would expect the visible boundary to be not straight, and the shapes/edges in the 2 screenshots look more screenaligned to me rather than tile-aligned. But i'm not sure..

[...]

PR #382, opened 2019-12-27
Add vertex light support

magenta404:
This will help address #357
There is also pixel light support with #383

Experimental support for vertex lights. Vertex lights are point lights which are set to _Not Important_ or in some cases _Automatic_.

Night
Night Time

### Attenuation
The attenuation formula was taken from [here](https://forum.unity.com/threads/point-light-in-v-f-shader.499717). The `UNITY_LIGHT_ATTENUATION` doesn't work for vertex lights. The `Shade4PointLights`, as discussed in the thread, didn't give good results.

### Shader Parameters
I've added some shader parameters so others for flexibility for the time being. Not sure if they will be merged. They are mostly there to separate the parameters for the primary direction light.

### Notes
`ShadeVertexLightsFull` cannot be used since it is for _Legacy Vertex Lit_.

### Tiling Artefacts
Tiling

Tiling artefacts occur depending on the attenuation, lights overlapping and camera orientation. The frame debugger shows lighting values being filled. It could be the attenuation calculation or a limitation in vertex lighting. Further investigation required.

### Underwater
Underwater

I have come across this issue when trying to integrate a weather asset with cloud shadows. It is much worse at night. Not including lights might be the best solution. Or at least having it behind a toggle.

## Missing

- Directional Light style reflections


huwb:
Really solid work. Code looks all good to me.

I tried it out and seems to work well from initial experiments.

Could you push the tiling fail case? You could check in main.unity temporarily and we could revert it later, if that makes sense.

daleeidd:
Thanks. Done

huwb:
It may be because the light is set to Auto and unity is deciding to make it a pixel light. If i set the red light to be "Not Important" the issue goes away.. ?

daleeidd:
I thought that too. I am in the same test case now and I have to set them (or at least the red one) to _Important_ for the issue to go away.

ABROKE
ANOTBROKE

daleeidd:
I just went through the frame debugger and `VERTEXLIGHT_ON` was active for those tiles.

Metal is a bit strange with lighting. I will have to test on Windows tomorrow. Hopefully, I can get the same as yours.

daleeidd:
I get the same for Windows as for Mac. The issue is that the red colour is severely diminished:

Lights marked as _Important_
A Important

Lights marked as _Not Important_
A Not Important

Also, if I change one of the green lights to a blue colour, the red colour is passed through correctly.

This could be a bug in Unity.

huwb:
Thanks for the above, and trying to wrap my head around things.

I think that Unity will move lights between vertex and pixel depending on number of lights, importance, maybe intensity(?). And there is a limit of 4 lights per vertex (and the directional light seems to impact the result for me - when i turn it off the result changes - so it's counted in the 4?).

If all that is true the story that my brain has constructed is that it is moving lights between vertex and pixel based on these factors, but this branch only has vertex, not pixel, so lights disappear. But I'm not 100% sure if that is consistent with what we're seeing / with the above?

huwb:
Worth noting as well that since verts can be displaced quite far horizontally, the render bounds are quite a lot larger than the tile appears and it may be overlapping more lights than it would seem

image

daleeidd:
Thanks. Displacement tip is useful. After disabling attenuation:
Displacement Lighting

I have gone over the following again:
https://docs.unity3d.com/Manual/RenderTech-ForwardRendering.html

> but this branch only has vertex, not pixel, so lights disappear

It is hard to say. The only thing I think I can do to confirm is to merge the pixel and vertex branch together, and max out the lights to see what happens. Not too difficult.

> But I'm not 100% sure if that is consistent with what we're seeing / with the above?

From what I have been experiencing, it hasn't been consistent with the manual. I didn't think _Important_ lights would even render without pixel light support in the shader. I thought that Unity was populating the shader data with pixel light calculations. I have also seen the issue with lower amounts of lights. Lastly, when comparing _Important_ and _Not Important_ in the Frame  Debugger, the light positions, world light and spherical harmonics are the same; the only thing changed is the light colours. I would think if the light disappears, then the light positions shouldn't include the same set of lights:

Important
111 Bad
Not Important
111 Good

I might have to make some reproducible cases with planes and a trimmed down shader. I'll let this sit in my mind for a bit.

daleeidd:
I will make a forum post or file a bug since the issue is unrelated to Crest.

daleeidd:
Finally got around to it: https://forum.unity.com/threads/vertex-lights-behaving-strangely.822300/

daleeidd:
Update from Unity (12/03/2020):
> We successfully reproduced this issue and have sent it for resolution with our developers.

daleeidd:
Removing draft status. It could probably use more work which I can look into another time.

Unity has given a conclusion:
> This particular case has been investigated thoroughly and we have decided, to not address this fix for the time being. Our developers found out that using stock URP shader works as expected, so you could use that as a workaround.

We could implement it here and then recommend URP if anyone wants nice lighting.

PR #383, opened 2019-12-27
Add pixel light support

magenta404:
This will help address #357
There is also vertex light support with #382

Experimental support for pixel lights. Pixel lights can be point, spot or directional lights which are set to _Important_ or in some cases _Automatic_.

Red Pixel
Point Light

Green Pixel
Spot Light

## About

This is a very rough PR. It will need a lot of work.

### Pass and Blending
Adds another pass and uses `BlendOp Max`. This prevents the more egregious artefacts. Improving the extra pass is a better way to go. Distilling the pass to only perform the necessary work for the light itself is the ideal.

### Performance
Only the tiles which are affected by the light will require additional passes. Performance will depend on how many lights and their coverage. Performance can be improved by removing unnecessary work.

## Issues

Plenty. I haven't properly worked through everything so some features are broken or misused.

### Technical Debt
In its current form, technical debt is pretty high.

### Directional Light
Based on light direction rather than camera to light for spot lights.

### DIFF
--- a/crest/Assets/Crest/Crest/Shaders/Ocean.shader
+++ b/crest/Assets/Crest/Crest/Shaders/Ocean.shader
@@ -525,6 +525,7 @@ Shader "Crest/Ocean"
 			// Temporary. Prevents doubling of some effects leading to artefacts.
 			// But this pass should be distilled so that only the light is applied instead.
 			// Doesn't work very well with foam lighting
+			// Blend SrcAlpha One
 			BlendOp Max
 			// Culling user defined - can be inverted for under water
 			Cull [_CullMode]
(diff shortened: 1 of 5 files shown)


daleeidd:
Removing draft status. It could probably use more work which I can look into another time.

Unity has given a conclusion:
> This particular case has been investigated thoroughly and we have decided, to not address this fix for the time being. Our developers found out that using stock URP shader works as expected, so you could use that as a workaround.

We could implement it here and then recommend URP if anyone wants nice lighting.

daleeidd:
Oops... made changes to wrong PR

Issue #381, opened 2019-12-26
Question: How do I get the hands to create ripples on ocean surfase in VR


EcoEditor:
Hello Huw, @huwb  

I'd like to ask you for your help regarding the Object Interaction with water.
In the demo scene three boats, the boats interact with water and create a "path" of interaction (small wave and foam) behind and in front of it.
![image](https://user-images.githubusercontent.com/42212240/71473992-1b293900-27e2-11ea-8056-7c9e5532f13b.png)

I've been trying to get my VR hand to interact with the water. This are the steps:
1. First the scripts responsible for Object Interaction is: RegisterDynWavesInput.cs (inherits from RegisterLodDataInput.cs and relevant interfaces + OceanRenderer.cs (variables as scale). Also the comment says: "Attach this GameObjects that you want to influence the sim to add ripples etc."
Seems that the "ObjectWaterInteraction.cs" and the "BoatAlignNormals.cs"/"BoatProbes.cs" will not be relevant in my case - since the player will initiate the movement.

2. After attaching this script to the VR hand, I get a very frequent and large wave input, also the wave is spiking regardless of the y position of the hand. 

3. Reading further in RegisterLodDataInput.cs, showed me that the material on the interacting object need to have a variable of "_Weight", since unity's "default material" has no such variable, I've created a new material with the DynWavesObjectInteraction.shader

4. The spiking waves disappeared, and the hands didn't affect the water at all. 

5. I've played with the shader properties and couldn't set it properly to interact with water.

6. Also it occurred to me that placing the object on the ocean relates with water interaction registered. For example, in threeboats scene, if the medium boat is too far away from the original position (400,0,400), it will not create ripples. I wonder what variable on the Ocean gameObject is responsible for this behavior. 

My goal is to create ripple based on hand movement, could you please guide me to the relevant scripts, scriptable object, components, shaders, video tutorial?
Thanks
L.

huwb:
Hello, I've created an experimental feature to make this easier. I modified ObjectWaterInteraction.cs to, if it does not find a boat script, attach an 'adaptor' script which will feed it the required information based on the parent object's motion.

20fe01ed15871cf756d379f23fb2cccfc190d3db

I have placed an "AnimatedObject" gameobject that demo's this. There is nothing special about this GameObject - it could be animated from any source, including moving the object manually in the scene view. It has a sphere shaped interaction object as a child which drives the interaction. 

Would it work to use the finger objects etc as the parent (instead of AnimatedObject)?

[...]

PR #387, opened 2020-01-02
Sphere object water interaction

huwb:
Status: Code ready to merge. Scene needs updating with latest from master.

Following on from issue #381 from @EcoEditor, i wanted to clean up our object water interaction system and have a second stab at how the interaction is modelled.

The old system renders a sphere or capsule mesh and scales it based on the velocity, and then does some calculations with the normal to estimate object water interaction. It was complicated and messy and unnecessarily expensive. It also didnt really deal with fading out the interaction when the object is partially submerged, or fully submerged under the water.

This PR adds a interaction model for a sphere. It models the sphere implicitly and only renders 2 triangles per interaction. It estimates the strength of the interaction based on how submerged the sphere is and has exponential falloff when sphere is fully submerged.

Multiple spheres can be composed to make nontrivial shapes:

Video: https://drive.google.com/open?id=1RZUy-jLF9j7aU4n_5QsS_rb3DrDHKwaS

Spheres (green wireframes):

image

It's not the most efficient approach to have a bunch of spheres individually updating, but for a hero asset like a fingers/a hand I think it can work well. It would be possible to add other primitives like capsules or cylinders but would require a more sophisticated approach and I feel this sphere version is worth putting into commission and leave nonspherical primitives to future work.

As a final, incidental change, the RandomMotion script has been extended to add random rotation to drive the 'spinner' object. It shouldn't affect existing uses of this script.


EcoEditor:
It looks good in the editor. Going to test with VR next week.

huwb:
Thanks both. If you see any issues later let us know. Merging.

PR #388, opened 2020-01-03
Unify depth cache validation

huwb:
Status: ready for review by commit. @moosichu / @daleeidd let me know if you are up for taking a look.

The validation code on the ocean depth cache was split into two places.

This change unifies it into one place and has an option to perform the validation on Start() (in addition to the 'Validate' button in the OceanRenderer inspector) which defaults to true. This way people get the full validation by default and can disable it if desired.

Added a validation error for the depth cache Y scale being 0, which breaks the cache population.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
@@ -296,7 +296,7 @@ namespace Crest
                 Debug.LogWarning("Validation: Ocean depth cache transform scale is small and will capture a small area of the world. The scale sets the size of the area that will be cached, and this cache is set to render a very small area. Click this message to highlight the cache in question.", this);
             }
 
-            if (transform.lossyScale.y < 0.001 || transform.localScale.y < 0.01)
+            if (transform.lossyScale.y < 0.001f || transform.localScale.y < 0.01f)
             {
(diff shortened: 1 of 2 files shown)


daleeidd:
Neat. Validate on start option is nice. The 'Force Always Update Debug' check happens twice though.

crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
```diff
@@ -301,6 +296,11 @@ public void Validate(OceanRenderer ocean)
                 Debug.LogWarning("Validation: Ocean depth cache transform scale is small and will capture a small area of the world. The scale sets the size of the area that will be cached, and this cache is set to render a very small area. Click this message to highlight the cache in question.", this);
             }
 
+            if (transform.lossyScale.y < 0.001 || transform.localScale.y < 0.01)
```
daleeidd: Should these floats have `f`?

daleeidd:
Also, while you're on this file, the `_geometryToRenderIntoCache` needs `true` to work in 2019.2 like `_layerNames`.

crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
huwb: yes! how did this even compile?? weird

huwb:
Good points, thanks! I believe these are addressed, let me know if you feel this is good to merge?

daleeidd:
All good from me

huwb:
Thanks!

PR #391, opened 2020-01-08
Fix incorrect condition in wave data generation from spectrum

EliGould:
This is a tiny thing but the condition is reverse here.


huwb:
Hm that's an odd one! Thanks!

PR #393, opened 2020-01-10
Clip surface using LoD system. Remove from ocean depth

### LINKED ISSUE #77: Mask/Cutout Ocean

daleeidd:
Fixes #77 

Before merging, the main scene will need to be reverted.

Attempt at using the LoD system for clipping the surface. Clips the surface without taking Y into account. 

Not sure if having this data in the debug GUI is necessary. Let me know if you want it removed.

I wasn't sure if I needed to call `BindResultData` or `BindData` anywhere else similar to the sea floor depth so I left it out since it works without it.

**Experiments**
I have done a couple of experiments adding Y similar to `OceanDepths.shader` which only takes above the surface into account (worked, but needs the displaced Y for it to be nice enough for use).

I also tried an experiment in using the Y for both above and below the surface by adding above to the red channel and below to the green channel; then when sampling, clip only if the two values where populated. Getting the data in the texture worked, but sampling was giving me weirdness.


huwb:
Super cool. Would you be up for adding some kind of demo of it to boat.unity? I'm not completely sure what. I guess an 'empty tub' would work perhaps..?

Due to the way lods blend and etc, I believe the 0.5 bias in the clip() statement should be retained. For example lods blending is pointless as anything slightly over 0 will be cut. Also I think 0.5 averages the 'error' coming from the texel resolution of the lod:

image

image

crest/Assets/Crest/Crest/Scripts/LodData/RegisterClipSurfaceInput.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Crest Ocean System
+
+// This file is subject to the MIT License as seen in the root of this folder structure (LICENSE)
+
+using UnityEngine;
+
+namespace Crest
+{
+    /// <summary>
+    /// Tags this object as a clip surface provider. Renders depth every frame and should only be used for dynamic objects.
```
huwb: Super minor - comment could use update

crest/Assets/Crest/Crest/Scripts/LodData/RegisterClipSurfaceInput.cs
```diff
@@ -0,0 +1,31 @@
...
+    /// </summary>
+    public class RegisterClipSurfaceInput : RegisterLodDataInput<LodDataMgrClipSurface>
+    {
+        [SerializeField] bool _assignClipSurfaceMaterial = true;
+
+        public override float Wavelength => 0f;
+
+        protected override Color GizmoColor => new Color(1f, 0f, 0f, 0.5f);
```
huwb: I guess we should pick some different colour. I'm open to ideas here :). Perhaps black?

crest/Assets/Crest/Crest/Scripts/LodData/RegisterClipSurfaceInput.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        protected override void OnEnable()
+        {
+            base.OnEnable();
+
+            if (_assignClipSurfaceMaterial)
+            {
+                var rend = GetComponent<Renderer>();
+                rend.material = new Material(Shader.Find("Crest/Inputs/Clip Surface"));
```
huwb: I think the shader will need to reside in the Resources folder for this to work (similar to the depth shaders)

crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgrClipSurface.cs
```diff
@@ -0,0 +1,61 @@
...
+            var drawList = RegisterLodDataInputBase.GetRegistrar(GetType());
+            if (drawList.Count == 0 && _targetsClear)
+            {
+                return;
+            }
+
+            for (int lodIdx = OceanRenderer.Instance.CurrentLodCount - 1; lodIdx >= 0; lodIdx--)
+            {
+                buf.SetRenderTarget(_targets, 0, CubemapFace.Unknown, lodIdx);
+                buf.ClearRenderTarget(false, true, Color.black * 1000f);
```
huwb: Super minor but factor of 1000 can be removed

b1gry4n:
Here to show support. The images youve posted have me excited. Im looking forward to using this in urp :)

crest/Assets/Crest/Crest/Scripts/LodData/RegisterClipSurfaceInput.cs
daleeidd: Black is good

daleeidd:
@huwb Thanks! Sorry, I should've been more diligent about the clip bias. I wasn't sure what it was so I have added the information you've provided as a comment.

I added another shader so clipping can be done with a texture too.

I would like your feedback on f907c4e. I thought we were sampling from the non-displaced position already. But by using the displaced position to sample the clipping surface data, it keeps the clipped surfaced aligned with the geometry correctly. Is this offsetting to re-align? There might be a nicer approach to mine. I am a little confused.

I believe the only thing left to do is revert the main scene.

huwb:
Nice! 

Re f907c4e , I'm checking on my phone so might be missing a something but what you have looks good to me. It should definitely use the displaced position so that it respects the desired clipping area. 

I think we're close to merging, feel free to revert the main scene and I'll try this out later today.

daleeidd:
Done

Lastly, just a reminder that these changes still exist from your prototype commit for the depth LOD manager (not sure if you want this file reverted too):

https://github.com/crest-ocean/crest/pull/393/files#diff-40f16de4725a41e06815ee381e8c6e04R18

https://github.com/crest-ocean/crest/commit/20c24b1f727c4a8482013e1faedc1ead05244cf1#diff-40f16de4725a41e06815ee381e8c6e04R48

huwb:
Gonna merge this in to the dev branch so we can see the final set of changes

PR #394, opened 2020-01-12
Added ability to adjust planar reflection intensity.

skaughtx0r:
This exposes a material variable that multiplies with the reflection alpha, so you can reduce the intensity of the reflection.

reflection_intensity_002

### DIFF
--- a/crest/Assets/Crest/Crest/Shaders/Ocean.shader
+++ b/crest/Assets/Crest/Crest/Shaders/Ocean.shader
@@ -65,6 +65,8 @@ Shader "Crest/Ocean"
 		[Toggle] _PlanarReflections("Planar Reflections", Float) = 0
 		// How much the water normal affects the planar reflection
 		_PlanarReflectionNormalsStrength("Planar Reflections Distortion", Float) = 1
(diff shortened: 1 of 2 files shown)


huwb:
Thanks @skaughtx0r 

Currently planar reflection intensity is controlled using the Specular control. That will correspondingly reduce the sky reflection as well as the planar. Can this be used for this control?

skaughtx0r:
@huwb Changing the specularity isn't giving me the desired effect. It has a side effect of reducing the light specular highlights as well as reducing the sky reflection which I don't really want. Here's some comparison screenshots.

Most of my trackside cameras are very low fov 2 - 5 degrees when zoomed in all the way, and when that zoomed in the reflections look way too intense by default.

This is default specular at 1 and intensity at 1.
default-reflection

This is specular at 0.5 and intensity at 1.
specular-at-5

This is specular at 1 and intensity at 0.75
intensity-at-75

huwb:
Thanks for illustrating. I haven't got my head completely around this, but softening the planar reflection seems like a good control to have (as in the above) so lets merge it. Thanks for this!

PR #395, opened 2020-01-12
Performance optimization, disable shadows when rendering planar reflection.

skaughtx0r:
Disable rendering shadows into the planar reflection. This cut the planar reflection render time in half for my project, from about 2.2ms to 1.2ms.


huwb:
Hi there, could this be a user configurable setting - similar to `_disablePixelLights` .

I'm undecided if it should be off by default - perhaps yes?

skaughtx0r:
Done. I put it behind a _disableShadows flag and set it to false by default.

huwb:
Looks perfect to me and worked in my test. Merging. Thanks for this!

PR #396, opened 2020-01-12
Feature/clip surface data

huwb:
Creating a PR for this feature that wraps in @daleeidd 's changes into the initial prototype.


huwb:
I removed the depth data changes, I'll create a separate PR for them.

I made a few minor changes here and there.

One worth mentioning is the example in boat.unity (that looks cool btw) - I dropped the clipping plane to the center of the object as it seemed a bit more robust to rotations etc if it was closer to the water level. Does this change make sense @daleeidd ?

If so thumbs up from me for merging this in if you're happy with the changes!

daleeidd:
Makes sense. All good from me.

huwb:
Decided to deprecate the old depth mask approach, which lead to a bit of a chain reaction of changes, mainly switching to the new approach for the rowboat in main.unity. I decided to add a warning if the clipping option is not enabled on the material, similar to some of the other lod datas, which hopefully helps validation. Hope these make sense to you.

Merging!

PR #397, opened 2020-01-12
Reduce depth precision to half

huwb:
I think this should be ok. Half precision didn't work in the past because the depth used to represent 'height' so had bad precision at shallow depths, and the artifacts were obvious. Now that its properly water depth it seems to work fine.

Add a couple of minor extras to explicitly target the R channel.

I tested in main which has a fairly large variation in sea floor depts and it seems fine to me.

@daleeidd @moosichu let me know if you have thoughts!


moosichu:
I'll try this on Vulkan and on a Mac to make sure it all works.

daleeidd:
Seems to work as before on Mac for me. I have a really shallow scene to test on.

huwb:
I tried on Vulkan (editor and standalone) and it seems fine, so I'll merge. if you see issues let me know. thanks both

PR #402, opened 2020-01-19
Generate displacements from height tex

huwb:
**Status:** experimental for now

AnimWavesAddFromTex - generate displacements from heights

Uses forward differences to compute derivates and displace proportionally.
Similar to how the disps are generated from the dynamic wave sim.

@skaughtx0r - is this still of interest? i think you mentioned you're using a different approach to get the displacements but thought i'd ask.


skaughtx0r:
Yeah, I'm actually just using a custom shader now. I'll try and test this when I get a chance though.

PR #406, opened 2020-01-22
Clip surface using convex hull

daleeidd:
Changes for having only the volume of a convex hull clip the ocean surface. It is kind of working. Differs from the previous approach by only clipping the surface if there is geometry both above and below the ocean surface.

## TODO

Using the ocean surface height is a hack currently which does not work properly. If the convex hull is positioned above `_OceanCenterPosWorld.y` and overlaps the surface height due to a large wave, then it will not clip the surface as it should.

Having the actual surface height in the clip surface data provider shader will be a better approach. 

Another possible approach is to render the top and bottom of the convex hull into the clip surface texture as long as it is within an appropriate distance from `_OceanCenterPosWorld.y`, and finally clipping the surface using that data.

Any suggestions on how to solve this or a better approach will be appreciated.

## Issues

- Two convex hulls –one above and one below the surface– will trigger the clipping when they should not. If it was possible to clear only red and green values (not yellow) after each object is rendered into the clip surface texture, then that would be the optimal solution. It could be mitigated with scripting.


huwb:
Nice one

Thats an interesting problem when two colliders overlap vertically! Need to think about that Maybe its an edge case that is rare, or simply a constraint that needs to be taken into account. Disabling clipping when an object is not close to the surface may mitigate this issue.

I guess for it to take into account the actual surface height we need to find the anim wave data. I wonder if there would be a downside to using SetGlobalTexture so that any shader that wants the data can get it.

daleeidd:
Thanks! Got it all working well enough. From here, we just need to determine if we want to use two channels or not. Also, whether we want to keep the previous way (mesh silhouette rather than convex hull) and make this one optional.

> Disabling clipping when an object is not close to the surface may mitigate this issue

I was thinking that too. Just a matter of how accurate we want the check to be. We would need the closest point on the mesh to the closest point on the surface (could be just a height query) to be really accurate. We could also cull ones that are way off without height queries for performance.

huwb:
Cool. It worked well for me with the spheres.

I'm not sure if this is a polish thing but i seem to see disparities between the intersection and the generated hole in some cases, seems like its breaking down for me for a rotated cube:

image

Yeah I would be interested if it can be done with a single channel (R8). I am wondering if perhaps the shader can have two passes, the first draws back faces and the second draws front faces. Maybe the back could draw 1 if underwater, otherwise clip(), and the front could draw 0 if underwater, otherwise clip? I dont know if both passes will automatically render, it may require hacking the shader code (or submitting the mesh twice with two shaders i suppose..).

(As an aside, it occurred to me it could be interesting to support signed distance fields as another source of shape for clipping - writing a gradient into the texture based on distance to surface instead of just 0 or 1 values should allow very precise hole edges, for the same reason as https://steamcdn-a.akamaihd.net/apps/valve/2007/SIGGRAPH2007_AlphaTestedMagnification.pdf . The downside is that a signed distance field is needed - i guess an analytical primitive like box, sphere etc. It would be fun to implement and a cool feature to have (feel free to take a run at it if you're interested). I don't think it would replace or make redundant this convex hull approach though as this is more practical for most people.)

daleeidd:
> I'm not sure if this is a polish thing but i seem to see disparities between the intersection and the generated hole in some cases, seems like its breaking down for me for a rotated cube

Nice catch, thanks! That's a vertex issue. Moving the shader code to the pixel shader fixes it. I will fix this.

daleeidd:
The accuracy issue is fixed. Not sure if we should use the vertex shader (using varyings) or fragment shader. The fragment shader is noticeably more accurate. Is performance a concern when sampling the surface in the shaders? Should varyings be an option as a toggle?

huwb:
Thanks, i'll check it out, hopefully today. 

Its not clear to me right now that vertex is an option - the hull could be very large triangles whereas the surface is quite detailed. if i'm not misunderstanding i think i would just default to fragment the way you have it now as it would cover most cases.

The remaining question for me is if it can be reduced to one channel using some sort of approach with the clipping, potentially similar to the above. let me know if you think this could be an option.

daleeidd:
The difference appears to be small for most cases. The following is a pair of images: one for fragment and one varyings respectively. The last set has a mesh which is vastly larger than the island.

1 Frag
1 Varyings
2 Frag
2 Varyings
3 Frag
3 Varyings
4 Frag
4 Varyings
5 Frag
5 Varyings
6 Frag
6 Varyings

daleeidd:
Also, I tried your approach to get it working on a single channel and I works. I'll add a commit tomorrow.

daleeidd:
It now only uses the R channel. I have also added an option to disable the input if it is too far from the surface to prevent them from nullifying each other. It works well enough. There might be a nicer or more optimal approach.

The only issue I can still see is that we cannot support composite hulls since they will interfere with each other when overlapping. The developer will have to merge their hulls together.

If everything seems good, I can do a round of comments, add user documentation and finally rebase to remove scene changes etc.

huwb:
Awesome! Nicely done

Understood about multiple meshes. I was wondering if we could add some kind of validation for it but im not sure..

Re fragment height vs vertex height - this is the image in my head (apologies for the programmer art):

image

I would expect people will use lowest res mesh possible and stuff like the above feels like it will happen often - i think?

daleeidd:
I've also added clipping to the logo renderer. Forgot to do this for the initial feature launch.

> Understood about multiple meshes. I was wondering if we could add some kind of validation for it but im not sure..

Well, overlapping meshes kind of work. It appears to depend on render order and if the offending mesh will be drawing black or not. The only thing that comes to mind is checking overlapping mesh bounds but this could produce false positives. An example of a use case that will work without issue is a bunch of overlapping AABBs. I've made a note of it in the user documentation.

> Re fragment height vs vertex height - this is the image in my head (apologies for the programmer art):

I sort of understand. I am happy to leave it as fragment shader only if you are.

I'll also need to update the scenes with convex hulls before merging. Other than that, I think everything looks finished.

huwb:
Thanks for this.

My last comment didnt properly explain my query about clipping based on heights sampled on verts. I edited it and hopefully it makes more sense, let me know if i've misunderstood.

If I understood correctly this will replace the previous clipping shader you made. Would it be possible to retain the old one and call this ClipSurfaceConvexHull or such? A common use case is to simply kill the water in an area, which i guess could be done with a convex hull as well, but would be slightly cheaper using the original shader i suppose. does this sound possible and is it worth doing do you think?

I'm starting to wonder if the shader name 'AddFromGeometry' makes sense (i know thats what it was called before this PR). I think it would be worth renaming to something about convex hull, and if we do bring back the last shader maybe we should give it a name like 'ClilpSurfaceRemoveArea' or something?

On the c# side, i'd usually try to avoid adding per-frame callbacks like Update and LateUpdate for cpu perf reasons. However i get the reasons for it. Maybe its worth merging Update into LateUpdate to consolidate them into one function if thats possible.

Let me know what you think!

daleeidd:
Thank you for clarifying. It would depend on the object whether that scenario would be an issue. Something with thick walls that couldn't use the `ClilpSurfaceRemoveArea` shader might benefit? It's too hard to say without coming up with a scenario. I am happy to readdress later. I might even find a use case myself.

I have incorporated all of your other suggestions. And I have updated both the scenes and documentation too. Let me know if there is anything else.

daleeidd:
Also... I was thinking about disabling the underwater shader if the camera is in the convex hull. It would be simple enough to add using a trigger collider or mesh bounds (to avoid add adding a collider on the camera); or harder using the mesh. It wouldn't be perfect, but good enough for a lot of use cases since the camera is mostly well within the hull. But maybe it should wait for #299 ?

huwb:
Yeah i agree it would depend on scenario and it would work in some. It could be added as an option for advanced users. I'd be weary of support / bug reports with less advanced users, as it may fail in non-intuitive ways.

Good point about turning off in clip areas, this slipped my mind in the past. Ideally this mechanism would support any clipping inputs (like remove area) so that it works for general cases, so would be driven by the clipping data. This would seem to be quite doable by duplicating QueryFlow for the clipping case, which the UnderwaterEffect could then use. Would you be up for creating a task for this?

#299 - i'm not sure if/when this will merge in - i've added a status to the description with more detail. if you have thoughts about feel free to chime in

Besides that it looks great, except git is having a small fit over the obj file. when i check out that branch, it looks like git is taking the initative and removing the content from the file, and adding a link to LFS storage, which is a strange thing to see. 

image

I think for some reason when you committed the file it neglected to use LFS. do you know why that might be the case? is it a checkin option on your client perhaps?

huwb:
(i see its the only obj file, we could probably turn off lfs for this type as a backup plan)

daleeidd:
> I'd be weary of support / bug reports with less advanced users, as it may fail in non-intuitive ways.

Good point. We'll leave it be then.

> Would you be up for creating a task for this?

Sure thing. Added. 

>  if you have thoughts about feel free to chime in

I can have a look. The PP stack has a few issues in VR.

> I think for some reason when you committed the file it neglected to use LFS. do you know why that might be the case? is it a checkin option on your client perhaps?

I am using git on the terminal. Something must of went wrong. I will look out for it next time since I don't have much mileage with LFS yet. Fixed in the latest commit.

huwb:
Ah cool. It looks good for me too now.

If you're happy then lets get it merged!

daleeidd:
Great. Thank you. Merged.

PR #407, opened 2020-01-25
Fix sampling foam from previous frame

huwb:
Selection of source lod was pretty broken. It would fail to get data
from the smaller cascade all the time, and could sample outside param
arrays!

@moosichu @daleeidd giving you guys a chance to review.


crest/Assets/Crest/Crest/Shaders/Resources/UpdateFoam.compute
```diff
@@ -37,20 +39,17 @@ void UpdateFoam(uint3 id : SV_DispatchThreadID)
 	half foam = 0.0;
 	const half r_max = 0.5 - _LD_Params_Source[sliceIndexSource].w;
 
+	// Try to sample the source slice, if it the sample position lies within its footprint?
```
moosichu: `// Try to sample the source slice, if the sample position lies within its footprint`

(minor grammar fix and that is what is happening, so remove question mark)

huwb:
Thanks both, mergin'

PR #411, opened 2020-01-28
[Test] Experiment with using Native Containers in query API

moosichu:
Seeing what the impact of this would be, if we were to actually support these we would probably have to keep supporting the old API a while as well which would require effort. Another issue with this is the fact that NativeArray won't support ref returns until Unity updates the version of C# it uses, which will remove a lot of the annoying assining-out to a variable and assigning back again.

But this does seem to work! If we were to have jobified queries as well, it would probably be best to have it work with an API like this.

### DIFF
--- a/crest/Assets/Crest/Crest-Examples/Shared/Scripts/Collision/OceanSampleDisplacementDemo.cs
+++ b/crest/Assets/Crest/Crest-Examples/Shared/Scripts/Collision/OceanSampleDisplacementDemo.cs
@@ -1,6 +1,7 @@
 ﻿// This file is subject to the MIT License as seen in the root of this folder structure (LICENSE)
 
 using Crest;
+using Unity.Collections;
 using UnityEngine;
 
 /// <summary>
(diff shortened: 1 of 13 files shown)


NuclearCookie:
I'm querying the ocean height around 90 times per frame, using the ShapeGerstnerBatched script.
This took 2.5ms on a desktop PC.
I have also converted what I needed to use NativeArrays, unity mathematics and made a job for ComputeDisplacements, and it the cost dropped to 1ms.
Looks well worth the effort.

moosichu:
Thank you for the feedback! I'm personally keen to move over to Unity.Mathematics and NativeContainers by default for most of our API, but mainting backwards compatibility is important so it's something we will have to think carefully about, especialy as it introduces dependencies on Unity packages and we already have to support 3 versions of crest across many more versions of Unity already.

When the asset store supports it, we plan on making all versions of Crest unity packages which will allow us to better able declare dependencies appopriately as well.

PR #414, opened 2020-02-02
Refactor HLSL

huwb:
**Status**: WIP

This branch refactors shader code based on learnings from SRP versions.


daleeidd:
What are the chances of us doing the clip surface refactor as part of this? It was simple enough to get it working in this branch. But I don't know how far away from finished this refactor is.

huwb:
As for status, if memory serves, I think this might be mergeable now, it's done a good job of moving hlsl code.

The bigger thing was to actually change the functions like WorldToUV to match the srp versions. This would be great to do but is very error prone and will take some time to do.

As for your actual question specific to the clip thing, I'm not sure, but potentially yes.

Shall we get this branch merged? Could you try it on your side and of it looks good merge it? Maybe try Reimport the shaders folder when you do.

huwb:
What is the clip surface refactor, is it related to getting it ready to work in hdrp? Sorry I'm a bit sleep deprived!

daleeidd:
Sure. I can do that.

Sorry. I was a bit loose with the term. It was just having it work with this new structure which also HDRP uses.

huwb:
Got it. You could include the changes into this branch before merging of they're logically related enough.

daleeidd:
I brought over the clip surface stuff. The changes made to the clip surface shader in the other PR didn't work which is why it probably wasn't working in HDRP either. Got it working with the new files.

I also brought over some minor, but safe, changes not related to the clip surface from downstream.

I'll do another test tomorrow, but everything appears to be working fine. The only thing I can notice that we might want to address is the `CBUFFER`s have been mostly disabled in this refactor. I remember there being notes about Vulcan having issues with this stuff. Should this be addressed before merging?

huwb:
I think it was just that CBUFFER is not defined in BIRP and it was never clear to me that constant buffers are supported. Although googling now, it looks like it is, for e.g. https://catlikecoding.com/unity/tutorials/scriptable-render-pipeline/custom-shaders/

I'm not sure how to get them in, should we try to define the macro if it isnt already? maybe we can find the definition in SRP.core or elsewhere.

If you want to look at this feel free, otherwise i will later.

The vulkan thing was just that every cbuffer relevant to a shader must have a unique name, otherwise they conflict.

daleeidd:
Sorry. A little confused. That link is for SRP. Wrong link?

Constant Buffer macros are listed here:
https://docs.unity3d.com/Manual/SL-BuiltinMacros.html
But they don't work for compute shaders apparently. Is that what you're referring too?

I did find this comment about that (but not much else so far):
https://forum.unity.com/threads/cbuffer-and-compute-shader.164968/#post-3420530

huwb:
Ok i think i just overlooked them before, not sure why. So lets just enable them.

If they don't work in compute we can find a solution.

daleeidd:
Done. I have tested it on both Mac and Windows and they appear to function identically to master.

ffdce92 was interesting. I don't know why this affected shadows but not others.

I added some dummy macros for using constant buffers in compute to silence the errors. Let me know if you want a better filename/location.

huwb:
Hm curious about the shadow lods, they all seem to work find for me in win64 editor. Don't want to drag this out unnecessarily, just understanding. ill anyway continue reviewing.

crest/Assets/Crest/Crest/Shaders/Macros.hlsl
```diff
@@ -0,0 +1,15 @@
+// Crest Ocean System
+
+// This file is subject to the MIT License as seen in the root of this folder structure (LICENSE)
+
+// MACROS - Macros to help with cross pipeline development
+
+#ifndef CREST_MACROS_H
+#define CREST_MACROS
+
+#ifndef CBUFFER_START
```
huwb: So did you need to add a stub for this in the end - did the macro not work in 2018? this is a good solution and looks good, im just asking out of curiosity

huwb:
This looks great, nice work.

I'm curious if any errors will come up on any compilers from the change of the type of slice index. However it didnt generate any for me when i reimported shaders and any remaining can come out in the wash.

Everything seemed to work for me, tested the example scenes in win64 editor and a standalone win64 build.

I'm happy to see this merged, merge it when you're ready. Thanks!

daleeidd:
Here is what it looks like on Metal (also happens in Windows for me):

shadows issue

They look a little faint because the screenshot was taken at the first frame.

crest/Assets/Crest/Crest/Shaders/Macros.hlsl
daleeidd: Actually... I think I have done the wrong thing here. The macros aren't defined, but I didn't check if the stub is required. I've made an assumption. Also, there is a typo which I have fixed. I can copy the code from the PPv2 stack which defines these very macros for compute.

daleeidd: I have implemented the stub from PPv2. Tested working fine on Metal. Thanks!

daleeidd:
Might as well test the cbuffer macro Windows tomorrow. I will also merge if you have you haven't found anything else.

huwb:
> Here is what the issue looked like on Metal (also happened on Windows for me)

Wow ok well nice find

Yes merge it in when you're happy

Issue #399, opened 2020-01-12
IndexOutOfRangeException with DWP2 using new height queries


skaughtx0r:
Getting an IndexOutOfRangeException in Crest.QueryBase.UpdateQueryPoints().

I started to try and make a reproducible project with just the DWP2 demo scene and Crest and it just errors immediately, the water never renders, and all the water objects just fall.

It seems the issue is that there is a s_maxQueryCount set to 4096 and if that gets exceeded, exceptions start to get spammed.

**Log**

The following errors are spammed repeatedly in the log.

```
IndexOutOfRangeException: Index was outside the bounds of the array.
Crest.QueryBase.UpdateQueryPoints (System.Int32 i_ownerHash, Crest.SamplingData i_samplingData, UnityEngine.Vector3[] queryPoints, UnityEngine.Vector3[] queryNormals) (at Assets/Crest/Crest/Scripts/Collision/QueryBase.cs:245)
Crest.QueryDisplacements.Query (System.Int32 i_ownerHash, Crest.SamplingData i_samplingData, UnityEngine.Vector3[] i_queryPoints, System.Single[] o_resultHeights, UnityEngine.Vector3[] o_resultNorms, UnityEngine.Vector3[] o_resultVels) (at Assets/Crest/Crest/Scripts/Collision/QueryDisplacements.cs:61)
DWP2.CrestWaterDataProvider.GetWaterHeights (UnityEngine.Vector3[]& p0s, UnityEngine.Vector3[]& p1s, UnityEngine.Vector3[]& p2s, System.Single[]& waterHeights0, System.Single[]& waterHeights1, System.Single[]& waterHeights2, UnityEngine.Matrix4x4[]& localToWorldMatrices) (at Assets/DWP2/Dynamic Water Physics 2/Scripts/WaterObject/WaterDataProvider/CrestWaterDataProvider.cs:104)
DWP2.WaterObjectManager.Schedule () (at Assets/DWP2/Dynamic Water Physics 2/Scripts/WaterObject/WaterObjectManager.cs:475)
DWP2.WaterObjectManager.FixedUpdate () (at Assets/DWP2/Dynamic Water Physics 2/Scripts/WaterObject/WaterObjectManager.cs:228)
```
```
ArgumentOutOfRangeException: Specified argument was out of the range of valid values.
Parameter name: Bad indices/count arguments (managedBufferStartIndex:0 computeBufferStartIndex:0 count:5196)
[...]

PR #415, opened 2020-02-02
Configurable Max Query Count

skaughtx0r:
This PR is in reference to Issue #399


skaughtx0r:
@huwb I defaulted s_maxGuids to 1024 as suggested, however I'm not sure I understand why that property exists at all. It looks like it's only being used to spit out an error and return early.

huwb:
Thanks, looks good to me! I made a couple of very minor tweaks, mainly defining a symbol for the default max query count value. Merging.

Issue #373, opened 2019-12-11
Crest does not support Domain Reload option


huwb:
Disabling domain reload breaks Crest currently.

huwb:
Oh #417 fixed this, so closing.

PR #417, opened 2020-02-06
Support no domain reload

huwb:
## Status

Ready to merge.

One of our users needs this with a priority if possible. @daleeidd if you're working today could you please give this a review? If not it's fine. Thanks!

## What / why

The speedup entering/exiting playmode when domain reload is disabled in 2019.3 is incredible.

As noted in #373, crest did not support this option before now.

If the domain is not reloaded, any static variables will retain there value across the transition. There is a special callback that can be used to initialise these variables if needed. More info: https://docs.unity3d.com/2019.3/Documentation/Manual/DomainReloading.html

## How

Some of our statics don't need to be reset across reloads. Some statics didnt need to be statics at all (for example, some shader param ids where private, and members of a class that were basically a singleton, so there was no point making them static - e.g. `sp_LD_TexArray_Flow`).

The remaining statics received (re)initialisation in the callback. The callback isn't usable pre-2019.3, so i've left initialisation both in line in the variable declaration, and in the callback function.

I could have ifdef'd out the entire callback function (`InitStatics`) each time, but `TextureArrayHelpers` actually needs to call the function as it has non-trivial initialisation, so i decided to go with this.

Also, `TextureArrayHelpers` used to depend on OceanRenderer from static functions which was a weird dependency, because it created a texture array with the current lod count slices, and the current lod resolution. I realised i could just create it as (4, 4, MAX_LOD_COUNT), where 4 is a 'small number' (2 may have worked as well, or maybe 1, but i seem to recall 4 being a min size for some gpu related things like compression, and thought it may as well make sense to use this). This saves memory, probably improves perf, and removes the dependency.

(A final note, it would be good agree a coding standard for statics (perhaps here: https://github.com/crest-ocean/crest/blob/master/CONTRIBUTING.md ). I think they were originally `s_` but have evolved a bit organically and there is a range of things like no prefix, `_`, `s_`, `sp_`, and `krnl_`. I started setting them to `s_` but stopped as i think it would be best to agree something we're all happy with, and it probably deserves a new PR).

## Tests

I ran the example scenes in both 2018.4 and 2019.3 and all seems fine here. I ran a standalone build from 2019.3 as well and it seemed to work.

## Issues

None that i'm aware of, besides the pitfall of double initialisation, which i'm not sure can be easily resolved in a way that is elegant and works on versions pre-2019.3 and after.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Collision/QueryBase.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Collision/QueryBase.cs
@@ -38,7 +38,7 @@ namespace Crest
         const int s_computeGroupSize = 64;
         public static bool s_useComputeCollQueries = true;
 
-        int sp_queryPositions_minGridSizes = Shader.PropertyToID("_QueryPositions_MinGridSizes");
+        readonly int sp_queryPositions_minGridSizes = Shader.PropertyToID("_QueryPositions_MinGridSizes");
 
         const float s_finiteDiffDx = 0.1f;
 
--- a/crest/Assets/Crest/Crest/Scripts/Collision/QueryBase.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Collision/QueryBase.cs
@@ -38,7 +38,7 @@ namespace Crest
         const int s_computeGroupSize = 64;
         public static bool s_useComputeCollQueries = true;
 
-        readonly static int sp_queryPositions_minGridSizes = Shader.PropertyToID("_QueryPositions_MinGridSizes");
+        int sp_queryPositions_minGridSizes = Shader.PropertyToID("_QueryPositions_MinGridSizes");
 
         const float s_finiteDiffDx = 0.1f;
 
--- a/crest/Assets/Crest/Crest/Scripts/Collision/VisualiseCollisionArea.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Collision/VisualiseCollisionArea.cs
@@ -12,7 +12,7 @@ namespace Crest
 
         float[] _resultHeights = new float[s_steps * s_steps];
 
-        static float s_radius = 5f;
+        readonly float s_radius = 5f;
         static readonly int s_steps = 10;
 
         Vector3[] _samplePositions = new Vector3[s_steps * s_steps];
--- a/crest/Assets/Crest/Crest/Scripts/Collision/VisualiseCollisionArea.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Collision/VisualiseCollisionArea.cs
@@ -12,7 +12,7 @@ namespace Crest
 
         float[] _resultHeights = new float[s_steps * s_steps];
 
-        readonly float s_radius = 5f;
+        static readonly float s_radius = 5f;
         static readonly int s_steps = 10;
 
         Vector3[] _samplePositions = new Vector3[s_steps * s_steps];
--- a/crest/Assets/Crest/Crest/Scripts/Helpers/OceanDebugGUI.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Helpers/OceanDebugGUI.cs
@@ -17,8 +17,8 @@ namespace Crest
         readonly static Color _guiColor = Color.black * 0.7f;
         ShapeGerstnerBatched[] _gerstners;
 
-        static Dictionary<System.Type, bool> s_drawTargets = new Dictionary<System.Type, bool>();
-        static Dictionary<System.Type, string> s_simNames = new Dictionary<System.Type, string>();
+        static readonly Dictionary<System.Type, bool> s_drawTargets = new Dictionary<System.Type, bool>();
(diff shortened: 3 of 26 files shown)


daleeidd:
Looks ok to me. It tested fine too. Hopefully I didn't miss something.

Couldn't collections like Dictionary be kept as readonly and just use `dictionary.Clear()` in `InitStatics`?

crest/Assets/Crest/Crest/Scripts/Collision/QueryBase.cs
```diff
@@ -38,7 +38,7 @@ public abstract class QueryBase : MonoBehaviour
         const int s_computeGroupSize = 64;
         public static bool s_useComputeCollQueries = true;
 
-        readonly static int sp_queryPositions_minGridSizes = Shader.PropertyToID("_QueryPositions_MinGridSizes");
+        int sp_queryPositions_minGridSizes = Shader.PropertyToID("_QueryPositions_MinGridSizes");
```
moosichu: Don't we still want to keep these as readonly?

crest/Assets/Crest/Crest/Scripts/Collision/VisualiseCollisionArea.cs
```diff
@@ -12,7 +12,7 @@ public class VisualiseCollisionArea : MonoBehaviour
 
         float[] _resultHeights = new float[s_steps * s_steps];
 
-        static float s_radius = 5f;
+        readonly float s_radius = 5f;
```
moosichu: Readonly values which are initialised with constant values can also remain static.

crest/Assets/Crest/Crest/Scripts/Helpers/OceanDebugGUI.cs
```diff
@@ -216,5 +216,16 @@ void ToggleGUI()
...
             _guiVisible = !_guiVisible;
         }
+
+#if UNITY_2019_3_OR_NEWER
+        [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.SubsystemRegistration)]
+#endif
+        static void InitStatics()
+        {
+            // Init here from 2019.3 onwards
+            s_drawTargets = new Dictionary<System.Type, bool>();
```
moosichu: I would keep the dictionaries as readonly and just clear them.

crest/Assets/Crest/Crest/Scripts/Collision/QueryBase.cs
huwb: I assumed that we need to reinit them each time. I did google it but didnt find anything. so i made them not readonly so they can be re-init'd each time. I think this is the safest route.

crest/Assets/Crest/Crest/Scripts/Collision/VisualiseCollisionArea.cs
huwb: Oops not sure why i did that. fixed now. thanks.

crest/Assets/Crest/Crest/Scripts/Helpers/OceanDebugGUI.cs
huwb: Ok ill do it that way

crest/Assets/Crest/Crest/Scripts/Collision/QueryBase.cs
huwb: sorry just realised this was one i removed the static flag from. yeah they could be readonly, good point. ill take a look

huwb:
Thanks both!

I responded to comments directly, thanks for those

@daleeidd good catch on the static events. looking at that one revealed something pretty funky in the code - currently every ocean chunk registers the begin camera rendering event, only to set the same static reference. i.e. `_camera` is set to the same value dozens of times. I've made a change which first deregisters before registering, which means only one chunk will eventually register and event (and i've added that code to deregister as instructed by the documentation). maybe there is a more sane way to do this, i need to think about it if theres a better way as this is a bit odd.

thanks for thoughts on coding standards too, we can discuss on discord if that makes sense

crest/Assets/Crest/Crest/Scripts/Collision/QueryBase.cs
huwb: done on latest. good catch

huwb:
Ok the latest should reflect all the feedback, does it look ok to you guys?

I retested on 2018.4 and 2019.3, all example scenes, with and without domain reload, and it seems to work well. And a standalone build too.

huwb:
Merging with Tom's blessing

daleeidd:
I just realised that the static event handler wasn't necessary this time round since `OnEnable` and `OnDisable` works like normal. Sorry about that. Unless you use `ExecuteInEditMode` which then "There is no extra OnDisable/OnEnable calls for ExecuteInEditMode scripts".

It is probably still not a bad idea to have it since it doesn't hurt.

PR #418, opened 2020-02-09
Reduce tiling artifacts with normal map and foam

skaughtx0r:
This adds stochastic sampling to ocean normal and foam shaders to reduce tiling. Shader functions are from: https://www.reddit.com/r/Unity3D/comments/dhr5g2/i_made_a_stochastic_texture_sampling_shader/

This is mostly useful for calmer water and lower tiling values on the normal map.

Stochastic sampling disabled
stochastic-disabled

Stochastic sampling enabled
stochastic-enabled


huwb:
Thanks for this @skaughtx0r and apologies for the long delay getting back to you.

Very useful code!

I'm not sure about merging it in mainly from a perf perspective - i think with flow enabled we already take 8 samples of each texture, and this would make it 24 which may be a bit excessive to ship to everyone. There's probably smart strategies to reduce this a bunch. We take 2 samples to have panning normals, and then we do this at 2 different scales (to break up patterns in normal maps and to make normals work in the distance), then we do this twice to animate the texture when flow is enabled. Perhaps these different techniques coudl be combined/share samples, rather than exploding combinatorially!

skaughtx0r:
Thanks for taking a look at this, yeah I was a little concerned with the amount of texture samples, but I didn't notice too much of a performance hit. Not sure if I'd be able to figure out how to optimize it myself.

huwb:
Sure. I'm happy to have this open and at some point in the future my curiosity will get the better of me and i'll come back and do some tests :). Til then!

skaughtx0r:
I'm ok if you want to just close this PR (to clean things up). I found that this is way too expensive on lower-end hardware, I ended up just using a larger higher resolution normal map instead. Also, the tiling isn't super noticeable in the foam anyway except around the shorelines.

huwb:
I made a task on our internal list in the Long Term bucket to revisit this. I will close for now. I'll report back if/when we do experiment with this.

Thanks again for sharing this and for the info.

PR #419, opened 2020-02-10
Fix bug where Gerstner Batches weren't getting removed when changing scenes

skaughtx0r:
Not sure how to reproduce in Crest project, but in my game I was able to consistently get this to happen. Every time I'd leave a race course and go back to the main menu, the Gerstner Batches were not getting removed from the `RegisterLodDataInputBase` registrar. This was causing some weird artifacts making the water look very twitchy.

[Video of the bug](https://youtu.be/Dyh8h2tWgT4)

Removing the check for `OceanRenderer.Instance != null` seems to fix it. I'm not sure why the OceanRenderer is already null in the OnDisable() callback, but it doesn't appear to be used.


daleeidd:
Thanks! The video is private though

skaughtx0r:
Whoops, fixed.

huwb:
Thanks @skaughtx0r , that looks like a good change to me, merging

Issue #318, opened 2019-08-13
Curtain fog density seems to disappear at lower depths.


holdingjason:
This entire sequence was recreated in the latest master, standard demo.

1.  Density of fog effect on objects seems to clear up the deeper you go.
2.  This effect can also be noticed on terrain etc as well as objects.
3.  Darkening effect as you go deeper does not appear to effect things like terrain etc near camera.

Near surface see the nice blue haze you have on the crab, fish and especially the terrain.
https://cdn.discordapp.com/attachments/568134291540803595/610571334275170326/unknown.png

Still some of that blue haze but now it seems to have less impact on the crab and terrain.  Also no real change in the darkness of the terrain and crab.
https://cdn.discordapp.com/attachments/568134291540803595/610571462235127809/unknown.png

Deeper still no haze at all on the crab or terrain.  Lighting in the distance is now totally dark as it should be.  Lighting however on the terrain and crab is wrong since we should be dark here as well, even without the haze.  In this situation however we do need haze since I would want to bring down a light source so we could see, I would want it to then look like the first shot with the blue haze.
https://media.discordapp.net/attachments/568134291540803595/610571704153931785/unknown.png?width=1076&height=527

holdingjason:
Just screwing around with this.  I found this code in the oceanemmisive shader.  This is where we check the camera depth and end up going dark at somepoint.  However this also completely eliminates the haze effect since the color drops out completely.  Ambient Light and light from the directional source are still lighting the scene so that is why we get light on the terrain and objects etc.  I can replicate what I want by limiting camDepth to say no more then 10 or so and then adjusting the ambient/directional light source in code to slowly darken them.  This works for the most part however if I shine a spot light at that point (or I am guessing a point light) it does not pick up any of the scattering effect but looks crystal clear.

	if (i_outscatterLight)
	{
		half camDepth = max(_OceanCenterPosWorld.y - _WorldSpaceCameraPos.y, 0.0);
		
		if (i_underwater)
		{
			col *= exp(-_DepthFogDensity.xyz * camDepth * DEPTH_OUTSCATTER_CONSTANT);
		}
	}

holdingjason:
Further thinking on this.  I don't think we want the curtain to go to far into dark etc.  This won't work with other lighting techniques ala spot light etc without adding that lighting to the curtain as well so it mixes correctly and keep it hazy at a minimum with a min color scattering.  Just tossing around random thoughts after playing with the shader.

[...]

PR #428, opened 2020-02-29
When underwater, attenuate scene colour to model outscattering of light

huwb:
## Status

Ready to merge

## What & why

Right now when camera is deep underwater and close to the seabed, the seabed still appears bright and it looks weird. I believe this is the primary issue reported in #318 . This change multiplies the scene colour by a term based on camera depth to model light getting scattered out and diminishing with depth.

Without this change - cam deep underwater near terrain:

image

With this change - light diminishes naturally and will go to black if deep enough:

image

## How

Multiplies scene colour by term based on camera depth to model light getting scattered out and diminishing with depth.

## Testing

Ran in win64 editor. Ran win64 standalone build.

## Issues and limitations

None that i'm aware of. I suspected a few potential issues but it seems to work very well.

### DIFF
--- a/crest/Assets/Crest/Crest-Examples/Shared/Scripts/UnderwaterEnvironmentalLighting.cs
+++ b/crest/Assets/Crest/Crest-Examples/Shared/Scripts/UnderwaterEnvironmentalLighting.cs
@@ -13,36 +13,44 @@ namespace Crest
     /// </summary>
     public class UnderwaterEnvironmentalLighting : MonoBehaviour
     {
-        float lightIntensity;
-        float ambientIntensity;
-        float reflectionIntensity;
-        float fogDensity;
+        float _lightIntensity;
+        float _ambientIntensity;
+        float _reflectionIntensity;
+        float _fogDensity;
 
-        void OnEnable()
+        float _averageDensity = 0f;
+
+        void Start()
         {
+            Color density = OceanRenderer.Instance.OceanMaterial.GetColor("_DepthFogDensity");
+            _averageDensity = (density.r + density.g + density.b) / 3f;
+
             // Store lighting settings
-            lightIntensity = OceanRenderer.Instance._primaryLight.intensity;
-            ambientIntensity = RenderSettings.ambientIntensity;
-            reflectionIntensity = RenderSettings.reflectionIntensity;
-            fogDensity = RenderSettings.fogDensity;
+            _lightIntensity = OceanRenderer.Instance._primaryLight.intensity;
+            _ambientIntensity = RenderSettings.ambientIntensity;
+            _reflectionIntensity = RenderSettings.reflectionIntensity;
+            _fogDensity = RenderSettings.fogDensity;
         }
 
         void OnDisable()
         {
             // Restore lighting settings
-            OceanRenderer.Instance._primaryLight.intensity = lightIntensity;
-            RenderSettings.ambientIntensity = ambientIntensity;
-            RenderSettings.reflectionIntensity = reflectionIntensity;
-            RenderSettings.fogDensity = fogDensity;
+            OceanRenderer.Instance._primaryLight.intensity = _lightIntensity;
+            RenderSettings.ambientIntensity = _ambientIntensity;
+            RenderSettings.reflectionIntensity = _reflectionIntensity;
+            RenderSettings.fogDensity = _fogDensity;
         }
 
         void LateUpdate()
         {
-            // Darken light when viewer underwater
-            OceanRenderer.Instance._primaryLight.intensity = Mathf.Lerp(0, lightIntensity, OceanRenderer.Instance.DepthMultiplier);
-            RenderSettings.ambientIntensity = Mathf.Lerp(0, ambientIntensity, OceanRenderer.Instance.DepthMultiplier);
-            RenderSettings.reflectionIntensity = Mathf.Lerp(0, reflectionIntensity, OceanRenderer.Instance.DepthMultiplier);
-            RenderSettings.fogDensity = Mathf.Lerp(0, fogDensity, OceanRenderer.Instance.DepthMultiplier);
+            float depthMultiplier = Mathf.Exp(_averageDensity * 
+                Mathf.Min(OceanRenderer.Instance.ViewerHeightAboveWater, 0f));
+
+            // Darken environmental lighting when viewer underwater
+            OceanRenderer.Instance._primaryLight.intensity = Mathf.Lerp(0, _lightIntensity, depthMultiplier);
+            RenderSettings.ambientIntensity = Mathf.Lerp(0, _ambientIntensity, depthMultiplier);
+            RenderSettings.reflectionIntensity = Mathf.Lerp(0, _reflectionIntensity, depthMultiplier);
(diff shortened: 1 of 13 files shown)


daleeidd:
Although it does work well for the sun, it also affects other lights. It wouldn't be possible to do a deep underwater scene with a torch for example.

huwb:
Good point! Hadn't thought about that.

I'm not sure if there's a solution to that, as all the lighting is baked in at this point. 

I guess an alternative would be to use a script that actually scales down the directional light intensity. Might be just crazy enough to work. I'll post back when I've tried it.

moosichu:
Would the idea be to have a script which reduces the intensity of directional lights based on their distance from the camera based on the depth-fog-density?

huwb:
That would be an interesting test. Maybe just the sunlight (?) as assigned to the oceanrenderer. If you want to try it out feel free, otherwise i will later.

huwb:
Just reread your question, no it would just be reducing the directional light intensity based on camera depth compare to sea level. So moving the changes in this PR from the hlsl into a c# script:

```
				// Darken the background to model light outscattered as it passes from sun/env through water to the background surface
				half camDepth = max(_OceanCenterPosWorld.y - _WorldSpaceCameraPos.y, 0.0);
				sceneColour *= exp(-_DepthFogDensity.xyz * camDepth * DEPTH_OUTSCATTER_CONSTANT);
```

huwb:
Well think i found a pretty horrible unity bug. the tangent vector was using a transform node with the space set to Absolute World. i didnt think this mattered as its set to Direction so position should not influence it.

image

I guess what it is doing is a translation to convert from absolute world to world, and then treating the result as a direction, which meant it would shift around depending on camera location.

Setting the space to World seemed to be enough for me, but ill try it again at SRP level now.

huwb:
Not sure if you decided to pick this up @moosichu , apologies if so, but i had a bit more time and had a stab at this.

I multiply down the primary light intensity, and also the ambient light intensity. Both are required to ramp down the light.

I'm not sure how i feel about changing the light in the scene, but it seems to work and i cant put my finger on why its not a good idea, so maybe it will work.

The one thing i dont love is having to read the fog values off the material, which means if the value on the material is changed it wont be picked up. I also don't love that it runs per frame but i guess thats ok.

The last thing is that if underwater is not enabled then we wouldnt want this to run, potentially adding yet another user configurable option. This can probably be eliminated by moving the code to UnderwaterEffect.cs. I'll look at this if we decide to proceed with this approach.

I'll test in the 'RPs next.

daleeidd:
Having it as an option is a good idea since most weather assets override the sun values and they will have their own way of managing it. Is it possible to expose the multiplier as a property? Then we could have our own component in crest examples which applies the values to sun and ambient. This might work better for downstream too due to lighting differences.

huwb:
Ah good idea, I'll try

huwb:
Ok the mechanism i was using to scale down the ambient lighting - *RenderSettings.ambientIntensity* - does nothing at all in URP. I looked through the lighting code and didnt spot any reference to a scalar on the lighting.

In this case i'm inclined to stick with the original solution - just scale the background colour - and call not supporting local lights a limitation. Its a shame but i dont see a way around it. Let me know what you think.

daleeidd:
It works for me but only when a _Skybox Material_ is set and _Environmental Lighting > Source_ is set to _Skybox_. You should see the _Intensity Multiplier_ slider when the settings are correct.

Also, we will have to disable the fog when underwater. Either by disabling it, or countering it in the underwater curtain. HDRP (and SRP once it supports it) can simply use the volume framework to blend out the fog once deep enough underwater. Fog would affect the underwater effects regardless of our solution for outer scattering.

The last issue is to do with baked lighting. If the lighting is baked when the skybox exposure is set to one, the ambient probe will have some colour even when the intensity multiplier is set to 0. There is nothing we can do about that it seems until Unity supports realtime GI in SRP.

Also, if we expose the depth fog colour, as well as the multiplier, then users will be able to add scripts to their lights to get the right colour.

I'm happy to work on this if you want to work on other things?

huwb:
Thanks for the dilligence

Interesting that we're using *Gradient* as the source in that scene, im doubtful whether this should be the case. i'm guessing we should probably just switch it to skybox.

> For HDRP we would have to blend out (or blend in 0 exposure sky cos of default settings or disable those default settings) the atmosphere volume in addition to the directional light.

Ok makes sense

I think this would be really cool. Lets chat on discord about short-term priorities.

daleeidd:
I made it a seperate component. I have confirmed that it works with URP. 

I added a spotlight to the end of the camera. Let me know if you want it kept or not.

As for HDRP, I have worked out a solution which works. It just requires an additional global volume which we will blend in which overrides indirect lighting (and anything else we need) as zero.

daleeidd:
I was thinking one thing we could add is a light intensity parameter to the ocean material which we could increase with depth to counter the unnatural change in lighting. It didn't make much of a difference though.

huwb:
Tested on birp, looks really cool.

I'm not sure we should leave a spotlight on the camera by default, for perf and misc. I'll think about it. We can probably demo it in other ways, like little glowing red mines?

huwb:
I didnt understand the parameter note. What is the unnatural change? In any case sounds like it didnt make an impact.

huwb:
Would it be worth moving the changes in OceanRenderer into the underwater script? I think i'd vote for this - if we're going to add a new script might as well put all the changes there, perhaps?

There could conceivably be something else that cares about the depth multiplier in the future, but i dont know what that would be, and we could always move it back.

crest/Assets/Crest/Crest-Examples/Shared/Scripts/UnderwaterEnvironmentalLighting.cs
```diff
@@ -0,0 +1,48 @@
+﻿// Crest Ocean System
+
+// This file is subject to the MIT License as seen in the root of this folder structure (LICENSE)
+
+using UnityEngine;
+
+namespace Crest
+{
+    /// <summary>
+    /// Provides outer scattering based on the camera's underwater depth. It scales down environmental lighting
```
huwb: Minor typo - outer scattering should be 'outscattering'

daleeidd:
> I'm not sure we should leave a spotlight on the camera by default, for perf and misc. I'll think about it. We can probably demo it in other ways, like little glowing red mines?

Sounds good. Done.

> Would it be worth moving the changes in OceanRenderer into the underwater script?

Sure thing. Done.

> I didnt understand the parameter note. What is the unnatural change? In any case sounds like it didnt make an impact.

We are reducing the lighting which affects the ocean surface and depth fog. I thought it would make the surface darken sooner than it should. But the impact was minor (I was sticking between 1 and 2 multiplier).

I also ported the out-scattering constant over.

crest/Assets/Crest/Crest-Examples/Shared/Scripts/UnderwaterEnvironmentalLighting.cs
daleeidd: Thanks for the correction. For some reason it has been outer scattering in my mind this whole time lol

huwb:
Happy for this to merge when you are. Nice one

daleeidd:
Your additions looks great. Merged

PR #429, opened 2020-03-01
Fix ps4 shader compile errors

### LINKED ISSUE #424: Shader compile errors on PS4
When I try to make a PS4 build I get these compile errors: 

image

Tested with built-in RP and Unity 2019.3.1f1 with a RTX 2060 Super on Win10.

moosichu:
Fixes #424 

Basically, the PS4 shader compiler doesn't implicitly down/up cast out params in shader procedures.

This fixes-up ones which were caught in that issue.

I have no real idea about how we would check for this issue in the code ourselves, beyond just being careful about this going forward. Could add it to the coding guide?

This is ready to merge. :)


huwb:
Just browsing PRs, when you consider this done could you write it up and we'll get it reviewed. Thanks!

huwb:
Ah interesting!

I guess there was a pretty inconsistent use of halfs and floats that drifted a bit over time, good to make it deliberate/explicit now.

I wonder if depth and SSS could be halfs? I think they're both stored natively as halfs and don't need full float precision.

moosichu:
Will jump on this at the weekend. ☺

moosichu:
@huwb I've done the requested for sss variables, but have not done so for Ocean depth. The reason for this is the fact that ocean depth is fed into `lodAlpha_worldXZUndisplaced_oceanDepth` which is a float4 (as opposed to a half4) variable, so I was wondering what you think the best way forward would be for that.

PR #431, opened 2020-03-02
Simple first version of dashboard

huwb:
Status: Ready to merge

This is a general purpose window to add dev stuff, but would not be intended to be included in published packages. Hence it is put outside the main crest folder.

So far it has scene loading buttons.

We could add buttons to resave SGs or to do other automation.

Thoughts @daleeidd / @moosichu ?

Thanks!

### DIFF
--- a/crest/Assets/Development.meta
+++ b/crest/Assets/Development.meta
@@ -1,0 +1,8 @@
+fileFormatVersion: 2
+guid: d9fc318dcee127f43a02f59eee10c7c9
+folderAsset: yes
+DefaultImporter:
+  externalObjects: {}
+  userData: 
+  assetBundleName: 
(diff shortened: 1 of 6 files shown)


moosichu:
Nice one! Only thing I can add at this moment is that we should add an Editor and Runtime asmdef for this dashboard stuff. But looks good otherwise. :)

daleeidd:
Perfect. This will be really handy

huwb:
@moosichu do you think you could add the asmdefs for me? my available time is pockets of around 5min on my phone at the moment and i'm a bit cold when it comes to this stuff

huwb:
Thanks. Assume we're happy with this, merging.

PR #432, opened 2020-03-03
Fix sea floor depths being at maxed value when not enabled

daleeidd:
Zero state for sea floor depths is red. Before this, it was black which makes the ocean completely flat and maxes foam values.

Is using a white texture like so suitable?

Also, should I go through and `BindNull` for depths everywhere else to be safe? The one I added wasn't necessary to fix this issue, but might be a good idea to be safe.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgrSeaFloorDepth.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgrSeaFloorDepth.cs
@@ -55,7 +55,7 @@ namespace Crest
         }
         public static void BindNull(IPropertyWrapper properties, bool sourceLod = false)
         {
-            properties.SetTexture(ParamIdSampler(sourceLod), TextureArrayHelpers.BlackTextureArray);
+            properties.SetTexture(ParamIdSampler(sourceLod), TextureArrayHelpers.WhiteTextureArray);
         }
 
 #if UNITY_2019_3_OR_NEWER
--- a/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgrSeaFloorDepth.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgrSeaFloorDepth.cs
@@ -70,8 +70,8 @@ namespace Crest
 
         static void InitNullTexture()
         {
-            var texture = Texture2D.whiteTexture;
-            // Null texture needs to be white with a 1000 intensity. 
+            var texture = Instantiate<Texture2D>(Texture2D.whiteTexture);
+            // Null texture needs to be white (uses R channel) with a 1000 intensity. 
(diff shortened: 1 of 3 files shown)


huwb:
Default for sea depths is "very deep" - 1000m. This is why it appears red.

I'm familiar with the flat ocean effect but it hasn't normally appeared. Are you getting it in some scenario?

I agree it would be best to have a null setter in theory. But it would need to be a texture that has the value 1000 if we need this.

daleeidd:
I get it whenever I disable the sea floor depth sim in the ocean renderer. Tested in Windows and Mac

> But it would need to be a texture that has the value 1000 if we need this.
Sure thing

huwb:
Ah gotcha. Thanks yeah, guess i havent tried disabling it recently.

daleeidd:
I have changed it to a 1000 white texture. Let me know if you are happy with where it is being initialised. I couldn't think of a better way.

daleeidd:
Red

So if I change the texture value to red (to test something), it changes all of the albedo tints to red. Is this something we need to worry about even with a white texture?

huwb:
Wow what.. you make a red texture and it makes the scene red? I've never seen something like that before. Can you tell me how to repro it - just use red instead of white?

huwb:
Oh you may have actually set unity's white texture to red instead of creating a new red texture :)

daleeidd:
Oh... yes... I should be cloning it first. Thanks!

daleeidd:
@huwb ok to merge with latest commits?

huwb:
Looks good to me!

daleeidd:
Thanks

crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgrSeaFloorDepth.cs
```diff
@@ -2,6 +2,7 @@
 
 // This file is subject to the MIT License as seen in the root of this folder structure (LICENSE)
 
+using System.Linq;
```
moosichu: Hi, just looking at this briefly, I can't see any reason why System.Linq would be needed in this file? Just going to test locally and will remove it if it isn't needed. Just giving a heads up here.

daleeidd: For `Enumerable.Repeat`. I should replace it with something that doesn't use Linq since I think Linq can produce garbage.

moosichu: I've made a PR for this #451, yeah, we are making an array anyway, so the garbage isn't so much of a concern (as it's already being generated). It's a preference thing more than anything, but the little research I have done suggest that the Enumerable.Repeat is a tad slower as well anyway.

PR #434, opened 2020-03-04
Refactor HLSL

daleeidd:
Continues on from #414

The main changes:
- Uses the constant buffer macros from `HLSLSupport.cginc` in compute shaders. Removes the ones I defined.
- Adds a function to sample the clip surface shader for SG
- Parity a few more constant buffer with downstream

I fixed a few warnings/errors that came up when checking against different graphics APIs. The Vulkan one is "Unknown Type Error".

I cannot tell from compiled code if our cbuffers are being used in compute or not. They might be ignored. It is not as obvious as other shaders.


huwb:
That's great that you could remove the macros. The hlsl support sounds great too.

Not sure if this is ready to merge but thumbs up if so.

daleeidd:
Thanks. Merged

PR #435, opened 2020-03-07
Use displacement for clip surface convex hull

daleeidd:
Before, the CSCH (clip surface convex hull) appeared to work pretty well. But once there was enough waveform displacement, due to chop and possibly other sources, the clipped area would become less accurate to the point where it would no longer appear.

This samples the displacement using the same code as the query displacement compute shader. Four iterations works well when testing which is what the other shader used too. I am not sure how we should set the minimum slice. I found that two was a good value. But it should probably be influenced by `_TexelsPerWave`?

I will need to add some documentation. Should I delete the old function (the first `SampleOceanDataAtWorldPosition`)?

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
@@ -10,7 +10,7 @@ namespace Crest
 {
     public interface ILodDataInput
     {
-        void Draw(CommandBuffer buf, float weight, int isTransition);
+        void Draw(CommandBuffer buf, float weight, int isTransition, int lodIdx);
         float Wavelength { get; }
         bool Enabled { get; }
     }
@@ -53,11 +53,14 @@ namespace Crest
             }
         }
 
-        public void Draw(CommandBuffer buf, float weight, int isTransition)
+        public void Draw(CommandBuffer buf, float weight, int isTransition, int lodIdx)
         {
             if (_renderer && weight > 0f)
             {
                 _materials[isTransition].SetFloat(sp_Weight, weight);
+
+                _materials[0].SetInt(LodDataMgr.sp_LD_SliceIndex, lodIdx);
+                _materials[1].SetInt(LodDataMgr.sp_LD_SliceIndex, lodIdx);
 
                 buf.DrawRenderer(_renderer, _materials[isTransition]);
             }
--- a/crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
@@ -58,9 +58,7 @@ namespace Crest
             if (_renderer && weight > 0f)
             {
                 _materials[isTransition].SetFloat(sp_Weight, weight);
-
-                _materials[0].SetInt(LodDataMgr.sp_LD_SliceIndex, lodIdx);
-                _materials[1].SetInt(LodDataMgr.sp_LD_SliceIndex, lodIdx);
+                _materials[isTransition].SetInt(LodDataMgr.sp_LD_SliceIndex, lodIdx);
 
                 buf.DrawRenderer(_renderer, _materials[isTransition]);
             }
--- a/crest/Assets/Crest/Crest/Shaders/OceanHelpersDriven.hlsl
+++ b/crest/Assets/Crest/Crest/Shaders/OceanHelpersDriven.hlsl
@@ -36,7 +36,7 @@ half4 SampleOceanDataAtWorldPosition(in Texture2DArray i_oceanData, in const flo
 }
 
 // Used to get the world position of the ocean surface from the world position by using fixed-point iteration
-float3 SampleOceanDataAtWorldPosition(in const Texture2DArray i_oceanData, in const float3 i_positionWS, in const uint i_iterations)
+float3 SampleOceanDataDisplacedToWorldPosition(in const Texture2DArray i_oceanData, in const float3 i_positionWS, in const uint i_iterations)
 {
 	float3 undisplacedPosition = InvertDisplacement(i_oceanData, _LD_Pos_Scale[_LD_SliceIndex], _LD_Params[_LD_SliceIndex], _LD_SliceIndex, i_positionWS, i_iterations);
 	return SampleOceanDataAtWorldPosition(i_oceanData, undisplacedPosition);
--- a/crest/Assets/Crest/Crest/Shaders/OceanHelpersDriven.hlsl
+++ b/crest/Assets/Crest/Crest/Shaders/OceanHelpersDriven.hlsl
@@ -35,4 +35,59 @@ half4 SampleOceanDataAtWorldPosition(in Texture2DArray i_oceanData, in const flo
 	return result;
 }
 
+float3 SampleOceanDataAtWorldPosition
+(
+	in const Texture2DArray i_oceanData,
+	in const float2 i_positionXZWS,
+	in const float i_minSlice
+)
+{
+	uint slice0, slice1;
+	float lodAlpha;
+	PosToSliceIndices(i_positionXZWS, _MeshScaleLerp, i_minSlice, slice0, slice1, lodAlpha);
+
+	const float wt_smallerLod = (1. - lodAlpha) * _LD_Params[slice0].z;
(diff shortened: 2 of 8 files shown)


huwb:
This clip shader+geo will be rendered into a particular slice, so there shouldn't be a need for specifying min slice or the code to compute the slice index - the slice index should be known at draw time.

I looked around for existing input shaders that use the slice index but the only one was the gerstner, and the gerstner script sets it (ShapeGerstnerBatched.cs:388).

I think we could add it for generic ocean inputs. I think that means perhaps adding an argument LodIdx to RegisterLodDataInputBase.Draw , and then setting the lod index on the material similar to how its set in the gerstner script above.

Then you should be able to get the lod index directly similar to how AnimWavesGerstnerBatch gets it.

daleeidd:
Thank you. That has improved everything significantly.

crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
```diff
@@ -53,12 +53,15 @@ protected virtual void Start()
...
         }
 
-        public void Draw(CommandBuffer buf, float weight, int isTransition)
+        public void Draw(CommandBuffer buf, float weight, int isTransition, int lodIdx)
         {
             if (_renderer && weight > 0f)
             {
                 _materials[isTransition].SetFloat(sp_Weight, weight);
 
+                _materials[0].SetInt(LodDataMgr.sp_LD_SliceIndex, lodIdx);
```
huwb: would only this suffice:
`_materials[isTransition].SetInt(LodDataMgr.sp_LD_SliceIndex, lodIdx);`

daleeidd: Thanks

daleeidd:
I think it is ready to merge if you do. Checked against DirectX/Metal/Vulcan and tested on Windows standalone. Feel free to do it if you want it part of the upcoming release and I don't do it in time.

crest/Assets/Crest/Crest/Shaders/OceanHelpersDriven.hlsl
```diff
@@ -35,4 +35,20 @@ half4 SampleOceanDataAtWorldPosition(in Texture2DArray i_oceanData, in const flo
...
 
+// Used to get the world position of the ocean surface from the world position by using fixed-point iteration
+float3 SampleOceanDataAtWorldPosition(in const Texture2DArray i_oceanData, in const float3 i_positionWS, in const uint i_iterations)
+{
+	const float3 queryPositionXZ = i_positionWS.xyz;
+
+	float3 undisplacedPosition = queryPositionXZ;
+	for (uint i = 0; i < i_iterations; i++)
+	{
+		const float3 displacement = SampleOceanDataAtWorldPosition(i_oceanData, undisplacedPosition);
```
huwb: I think this may still do unnecessary work = a lod alpha will be computed (in SampleOceanDataAtWorldPosition), and lods being interpolated. I was hoping instead of calling this function, it would simply sample the lod slice directly, so this line would become:

`const float3 displacement = i_oceanData.SampleLevel( ... , WorldToUV(...));`

thinking about it, should we separate the iteration out, for example make a helper function called `InvertDisplacement(float3 displacedPosition)` or such which is specific to displacements and returns the undisplaced position given a displaced position? just wondering as this operation might be reusable elsewhere.

sorry to send you around the houses instead of giving you more specific instruction in the first place!

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/ClipSurfaceConvexHull.shader
```diff
@@ -25,6 +25,10 @@ Shader "Crest/Inputs/Clip Surface/Convex Hull"
 			#include "../../OceanHelpersNew.hlsl"
 			#include "../../OceanHelpersDriven.hlsl"
 
+			CBUFFER_START(CrestPerOceanInput)
+			uint _DisplacementSamplingIterations;
```
huwb: Side note - this will stop the loop getting unrolled and potentially make the iteration more expensive. However, how much more expensive depends on hardware and i don't think the clip data is going to render tons of pixels (the lods are modest in resolution and number), and if you feel this is a useful control then lets keep it.

crest/Assets/Crest/Crest/Shaders/OceanHelpersDriven.hlsl
huwb: if the above does work out, i'll be curious if it can be moved to *OceanHelpers.hlsl* , rather than the *Driven* version, by passing in the slice index and any other driven data as arguments. Only because then its dependencies are straightforward and clear and it can be used outside of 'driven' cases.

daleeidd: So if I change to that line I won't have to sample two scales/textures?

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/ClipSurfaceConvexHull.shader
daleeidd: I see. Thanks. I am happy to have it reverted. The `uint` on the C# property is enough.

crest/Assets/Crest/Crest/Shaders/OceanHelpersDriven.hlsl
huwb: Yeah it should in theory just stay within one lod.

i'm waiting to see what happens at the boundaries, which could displace off-texture. i *think* it won't cause issues but we'll see how it pans out. in the worst case we'll need to revert it back to how you have it now, but i think its worth trying this simplification to avoid sampling two textures for every iteration.

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/ClipSurfaceConvexHull.shader
daleeidd: I completely misinterpreted that. Would be good to keep it. Sometimes a value of 10 greatly improves the hole.

daleeidd: I completely misinterpreted that. Would be good to keep it. Sometimes a value of 10 greatly improves the hole.

huwb: Ah fair let's keep it then

daleeidd: On further testing, ten is only for very extreme waves which I don't think anyone would use since they look broken.

We could replace it with an enum (DisplacementSamplingQuality: Low = 0, Medium = 4, High = 8)? Would duplicating the loop and selecting the right one with an if-else be a better alternative?

crest/Assets/Crest/Crest/Shaders/OceanHelpersDriven.hlsl
daleeidd: I pushed the change. Let me know if I interpreted you correctly.

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/ClipSurfaceConvexHull.shader
huwb: Yeah maybe that is a good shout. Its hard to know if it will be faster but it seems like a reasonable approach.

crest/Assets/Crest/Crest/Shaders/OceanHelpersDriven.hlsl
huwb: Yeah looks great! My only remaining doubt is the two functions with the same name `SampleOceanDataAtWorldPosition` differing only in args. Trying to think if we can differentiate them. `SampleOceanDataDisplacedToWorldPosition` maybe?

daleeidd: Sure. Done (hopefully renamed, the right one)

huwb: looks good nice one

huwb:
Thanks dale!

PR #437, opened 2020-03-10
Remove duplicated weight/strength input on object-water interaction

huwb:
**Status:** Ready to merge

**Overview:**

Weight input on script and Strength input on the material are multiplied
together. This causes confusion as people arent sure what the difference
between the two is.

This change fixes the weight at 1, leaving only the strength param. Some
weights were set > 1, this change bakes that into the materials.

**Tests:**

Ran boat scenes in editor and verified wakes and other interactions look unchanged.

**Issues:**

If people have weights that aren't 1, this change will suddenly increase the strength of the interaction. I suggest to mitigate this with release notes.

### DIFF
--- a/crest/Assets/Crest/Crest-Examples/BoatDev/Data/BoatAlignNormal.prefab
+++ b/crest/Assets/Crest/Crest-Examples/BoatDev/Data/BoatAlignNormal.prefab
@@ -322,7 +322,6 @@ MonoBehaviour:
   _localOffset: {x: 0, y: 0, z: 0}
   _noiseFreq: 6
   _noiseAmp: 0
-  _weight: 5
   _weightUpDownMul: 0.3
   _teleportSpeed: 500
   _warnOnTeleport: 0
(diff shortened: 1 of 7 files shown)


daleeidd:
Feel free to disregard this if you strongly believe this is the best way to solving this.

Materials aren't very flexible compared to scripts. Couldn't a tooltip help people with this confusion? Something like: Weight is applied against strength to add flexibility. It sounds like it could be a useful but I am not really sure.

Speaking of [tooltips, we could even add them to materials](https://gist.github.com/JohannesDeml/9b04bd62a975f3f2c8ffcdf0d7400762) if we are willing to take on the extra code. It might go a bit of way to helping people use other material properties too.

Otherwise, another way to mitigate the removal of this would be to use the _System.Obsolete_ attribute. The warning might be a nuisance but I can imagine it will be worse without it for those that used this property.

huwb:
Thanks for the thoughts @daleeidd

Yeah I think I do strongly feel the weight should be removed, this will always be a source of confusion I feel, a tooltip would help but in my opinion be a bandaid for the core problem - duplicated factors.

The strength param on the material can be set from code so I don't feel it's a high barrier.

Besides, I've never found myself wanting to change this param at runtime as it scales with velocity etc. So it's a triple whammy of confusing authoring/being error prone/not actually useful.

Does this makes sense to you?

daleeidd:
Ok. Makes sense. Thanks.

huwb:
Interesting about material property tooltips though.. tempting to look into adding them for our materials. Would not work for SG material I guess, but would be good for other materials.

PR #440, opened 2020-03-11
Local gerstner waves

huwb:
**Status:** Ready to review code (documentation not done yet). Main will be reverted.

**What & why:**

Our gerstner waves are currently global (apply everywhere to the water). This change adds a *Mode* option similar to Volumes in HDRP where one can either have it *Global* which keeps the behaviour how it is now, or set it to *Geometry* which will use the geometry on the GameObject to render the waves.

This means different wave conditions can be added to different parts of the world, and overlapped/combined as desired. Also this is a good step towards 'local' water bodies which is requested frequently, as it limits the computation to a local area. Additionally, this feature enables a crude but simple way to oriented waves towards shoreline - overlap multiple patches with different wind directions.

This includes a simple, optional way to blend the waves out - it assumes UVs range from 0 to 1 and blends out the waves near these boundaries. This will work for some meshes, i anticipate for more general cases we'll want to add another version that weights the waves using vertex colours, this would be simple to add.

**How:**

Instead if rendering a 'full screen quad' into the displacement textures, renders arbitrary geometry. There is plenty of precedent for this so this feature was fairly simple to add once the input end was figured out.

**Testing:**

Ran win64 editor & standalone.

To test run main and there should be two patches of waves.

**Issues:**

None that come to mind. Default behaviour should not change.


crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/AnimWavesGerstnerBatch.shader
```diff
@@ -2,13 +2,11 @@
 
 // This file is subject to the MIT License as seen in the root of this folder structure (LICENSE)
 
-// A batch of Gerstner components
-Shader "Crest/Inputs/Animated Waves/Gerstner Batch"
+// Adds Gerstner waves everywhere. Must be given batch prepared by ShapeGerstnerBatched.cs.
+Shader "Crest/Inputs/Animated Waves/Gerstner Batch Global"
```
daleeidd: Will anyone use this shader directly? It could be made hidden if not.

daleeidd:
This is great. I've wanted this feature for a while now myself. Works really well.

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/AnimWavesGerstnerBatch.shader
huwb: No they won't - good shout

huwb:
Good point about demo. I tried to set something up on main but couldn't figure out how to put it to good use. I wanted to use it for shoreline waves but it requires lots of gerstner objects to surround an island like that, and I think they're will be a better way to achieve this (there's a task for this on our list).

So instead I just such a patch of waves into boat.unity as a lazy way to demo it for now.

Local water bodies would give better prospects for demoing it. Was also thinking about a big fan that generates waves, may have a stab at this some time.

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/AnimWavesGerstnerBatch.shader
huwb: Done

huwb:
I made a few more changes - mostly added doc. I assume you're still happy for this to be merged but feel free to comment on this and ill make further changes.

PR #442, opened 2020-03-13
Improve clip surface usability

daleeidd:
Improvements based on feedback from @huwb


huwb:
Thanks @daleeidd

I have doubts about assigning the shader. I mentioned this but probably want clear my bad. Excerpt from other conversation (I'm on my phone so being lazy!):

> But the user could also assign the material them self and I feel there are redundant authoring paths (specify on script vs assign material) which can conflict and create confusing behaviour. Also, some inputs have parameters on the material, so automatically creating the material doesnt work for those cases because material defaults often wont be correct. I think i'd vote for either not assigning a material at all and depending on the user, in line with most of the other register scripts.

If the material is automatically created I'm not sure there's a way for users to tune the params on the material, which is currently done at authoring time.

What do you think?

daleeidd:
Sorry. I misread the issue. I will replace the assign feature for manually settings materials instead.

daleeidd:
@huwb  Fixed. Ready for review

PR #443, opened 2020-03-13
Don't rely on auto-generated shader for remove area

huwb:
It can return values less than 1 which is not intended.

### DIFF
--- a/crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/ClipSurfaceRemoveArea.shader
+++ b/crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/ClipSurfaceRemoveArea.shader
@@ -8,11 +8,42 @@ Shader "Crest/Inputs/Clip Surface/Remove Area"
 {
 	SubShader
 	{
+		Tags { "Queue" = "Geometry" }
+
 		Pass
 		{
 			Blend Off
 			ZWrite Off
 			ColorMask R
+
+			CGPROGRAM
+			#pragma vertex Vert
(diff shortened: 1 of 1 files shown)


crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/ClipSurfaceRemoveArea.shader
```diff
@@ -8,11 +8,42 @@ Shader "Crest/Inputs/Clip Surface/Remove Area"
...
+			};
+
+			Varyings Vert(Attributes input)
+			{
+				Varyings o;
+				o.positionCS = UnityObjectToClipPos(input.positionOS);
+				return o;
+			}
+
+			half Frag(Varyings input) : SV_Target
```
daleeidd:
```
Metal: Error creating pipeline state (Crest/Inputs/Clip Surface/Remove Area): fragment shader color output does not have enough components for the pixel format (MTLPixelFormatBGRA8Unorm_sRGB)
(null)
```
Changing it to `half4` silences this error. It still works with the error though. You can still use `1.0` as return, it's only the return type it cares about.

huwb: Thanks will change it

huwb:
@daleeidd let me know if thats good to go now

daleeidd:
Sorry. This slipped through my fingers. Tested. Looks good

huwb:
Thanks!

PR #444, opened 2020-03-13
Sort ocean inputs based on render queue

huwb:
**Status:** Ready to merge

This will allow ordering between clip shaders (include before remove). This change is taken from the `feature/local-water-bodies` branch where it seems to work.

Merging this into master should have no impact on existing content/functionality whatsoever as there are no order dependencies that I'm aware of between current inputs, so it should be inconsequential if the order the inputs are submitted changes.

It could conceivably add a bit of extra cost to adding/removing ocean inputs due to the sorting, but the number of inputs will rarely be large so it seems unlikely it will ever be an issue.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
@@ -2,12 +2,29 @@
 
 // This file is subject to the MIT License as seen in the root of this folder structure (LICENSE)
 
+using System;
 using System.Collections.Generic;
 using UnityEngine;
 using UnityEngine.Rendering;
 
 namespace Crest
 {
+    using OceanInput = SortedList<int, ILodDataInput>;
+
+    /// <summary>
+    /// Comparer that always returns less or greater, never equal, to get work around unique key constraint
+    /// </summary>
+    public class DuplicateKeyComparer<TKey> : IComparer<TKey> where TKey : IComparable
+    {
+        public int Compare(TKey x, TKey y)
+        {
+            int result = x.CompareTo(y);
+
+            // If non-zero, use result, otherwise return greater (never equal)
+            return result != 0 ? result : 1;
+        }
+    }
+
     public interface ILodDataInput
     {
         void Draw(CommandBuffer buf, float weight, int isTransition, int lodIdx);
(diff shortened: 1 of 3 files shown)


crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
```diff
@@ -73,7 +91,7 @@ public void Draw(CommandBuffer buf, float weight, int isTransition, int lodIdx)
         static void InitStatics()
         {
             // Init here from 2019.3 onwards
-            _registrar = new Dictionary<System.Type, List<ILodDataInput>>();
+            _registrar = new Dictionary<System.Type, OceanInput>();
```
daleeidd: You should be able to replace this with `.Clear()`. Even without static domain reload it won't skip static initialisation for the very first time.

huwb: Thanks will do

huwb:
Squeezed in a rename as well, hope thats ok. Mergin'.

moosichu:
This generates a fair chunk of garbage (2K per frame). I'm going to write an automated test for that this weekend as well.

moosichu:
Indexing a C# dictionary generates garbage for example :(

crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
```diff
@@ -26,15 +43,16 @@ public abstract class RegisterLodDataInputBase : MonoBehaviour, ILodDataInput
 
         public static int sp_Weight = Shader.PropertyToID("_Weight");
 
-        static Dictionary<System.Type, List<ILodDataInput>> _registrar = new Dictionary<System.Type, List<ILodDataInput>>();
+        static DuplicateKeyComparer<int> s_comparer = new DuplicateKeyComparer<int>();
+        static Dictionary<Type, OceanInput> s_registrar = new Dictionary<Type, OceanInput>();
 
-        public static List<ILodDataInput> GetRegistrar(System.Type lodDataMgrType)
+        public static OceanInput GetRegistrar(Type lodDataMgrType)
```
moosichu: Iterating over a sorted list in a foreach loop generates a lot of garbage.

moosichu:
```
 SortedListCachedArrays<TKey, TValue> : SortedList<TKey, TValue>
```

There is already a solution to this!

crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
huwb: Thanks good catch

PR #445, opened 2020-03-14
Remove garbage generation due to usage of sorted list

moosichu:
The builtin C# SortedList type generates garbage when iterating over it in a `foreach` loop.

We started using one as a result of PR #444, which was causing about 1.9K of garbage to be generated each frame.

This change introduces our own SortedList type, which is backed by a regular C# list. It isn't self sorting and has to be explicitly sorted after modification. I think this is fine as the lists tend to be quite small, and aren't modified often.

I've profiled the game with this list, and garbage generation is now avoided.


huwb:
Thanks Tom, could you add pr documentation please. What have you changed and why, etc. Thanks!

huwb:
Thanks. Did you rename the data structure? Has it changed - it's not showing me a line by line diff, just that offers removed and ocean added.

moosichu:
Yes, the datastructure has completely changed. Using the existing cached sorted list had some weird issues when I tried to use it (for some reason depth stopped rendering). Debugging wasn't super fruitful so I just implemented what I thought we might need from first principles.

It now is just a datastructure backed by a regular list, but with APIs to make it act mostly like a `SortedList` in that you can provide a comparer for the keys, and it will handle that all for you. It's not super efficient, but it does seem to do the job.

huwb:
Thanks dude. I stared at it for a while.

I dont think it needs sorting after removing an item? The order should not be broken by a removal i guess.

I think i'd vote for always sorting on add. Basically my philosophy would be that it should keep itself sorted at all times, otherwise 'sorted' is no longer an invariant, and it might be error prone for users.

One last question - do you think we could shave off the enumeration code - i'm not personally attached to foreach loops and i feel a `for(int i = 0; i < container.Count(); i++) container.GetItem(i);` style setup would work perfectly well. let me know what your thoughts are?

Thanks! Good catch IDing this issue.

huwb:
Actually the enumeration code is not much at all, so i'd be up for keeping it. i for some reason though it needed more.

moosichu:
> 
> 
> Thanks dude. I stared at it for a while.
> 
> I dont think it needs sorting after removing an item? The order should not be broken by a removal i guess.
> 
> I think i'd vote for always sorting on add. Basically my philosophy would be that it should keep itself sorted at all times, otherwise 'sorted' is no longer an invariant, and it might be error prone for users.
> 
> One last question - do you think we could shave off the enumeration code - i'm not personally attached to foreach loops and i feel a `for(int i = 0; i < container.Count(); i++) container.GetItem(i);` style setup would work perfectly well. let me know what your thoughts are? **EDIT:** maybe discard this thought as i just rereviewed and it doesnt add much code to have it
> 
> Thanks! Good catch IDing this issue.

You are correct about removing elements, so I've removed that code!

As the only way of accessing this array is through the enumerator, and we always resort it (if needed) when the enumerator is accessed, it's completely safe to only resort the array at that point. Therefore I've kept things as-is, but have left it clearer in the comments that is what the intention is.

huwb:
Ah gotcha, cool.

It all looks good to me, just one more query - could the backing list be private? I'm on my phone so might be missing an obvious reason why it must be public, if so apologies.

moosichu:
> Ah gotcha, cool.
> 
> It all looks good to me, just one more query - could the backing list be private? I'm on my phone so might be missing an obvious reason why it must be public, if so apologies.

That was literally just me being dumb! It was meant to be private, doing that now. :)

PR #446, opened 2020-03-17
Make Crest handle failed compute shader loads more gracefully

moosichu:
A common issues users have is that compute shaders won't load correctly through no fault of their own. However, previously Crest would just null-reference exception all over the place without a clear indication to the user as to why.

This change aims to ensure two things:

- When a compute shader fails to load, we give a good error message explaining that and provide a common solution.

- Crest no longer spams the log with new null reference exceptions every frame if the compute shaders fail to load.

The first point is achived by replacing the calls to `Resources.Load` to a call into a wrapper which asserts if the load fails.

The second point is achived by finding all the bits of code which only run under the assumption that compute shaders will successfully load, and ensure that they now simply stop updating or just *do nothing* if that assumption fails.

Below is a screenshot showing what I get if I have manually deleted all the compute shaders myself:
image


huwb:
Nice one!

PR #451, opened 2020-03-21
Make array initalisation use loop instead of LINQ

moosichu:
Motivation:

LINQ functions, whilst convenient can be slow and generate more gargbage than neccessary. This change was motivated by #432 an is incredibly minor.

The main reason I'm making a PR for this is because I noticed that Texture2D has a new NativeArray API: https://docs.unity3d.com/2018.4/Documentation/ScriptReference/Texture2D.GetRawTextureData.html

I looked into using it, but in the end I went for the safe incremental improvement. But I thought it would be worth highlighting this as an option we could start using in the future.


huwb:
Fine with this change. Thanks for the info.

PR #452, opened 2020-03-21
Make _LightShadowData knockout apply to Metal as well

### LINKED ISSUE #449: Update Shadow Compute Shader won't compile on Mac
**Describe the bug**

When opening the latest version of master of Crest on a Mac, the following error can be found in the log:

```
Shader error in 'UpdateShadow.compute': Unknown parameter type (54) for _LightShadowData at kernel UpdateShadow
```

**Screenshots / video**

image

**Log**

See above

**Versions**

Macbook Air, Builtin, Unity 2018.4, latest commit from master. 

(eb9efeb76084fe3617bf94b6010e8f1112a21076)

As the bug is so simple, the other information isn't needed for the bug report.

moosichu:
Fixes #449.

We already had the problem from the issue (UpdateShadow compute shader not compiling) on Vulkan, so applying the same fix to Metal was a straightforward thing to do.

Honestly though, I think part of the issue does come from the fact that we are trying to do this shadow stuff in a compute shader, and unfortunately Unity doesn't support having standard shaders target a specific slice in a texture array until 2019.1 (which is why we don't do this with a compute shader in URP), so we have to stick with the compute shader in BIRP for now.

### DIFF
--- a/crest/Assets/Crest/Crest/Shaders/Resources/UpdateShadow.compute
+++ b/crest/Assets/Crest/Crest/Shaders/Resources/UpdateShadow.compute
@@ -14,14 +14,14 @@
 //     _LightShadowData at kernel UpdateShadow
 // We skirt around this by defining our own _LightShadowData as a float4, and
 // then making sure that the *real* _LightShadowData variabel is knocked-out.
-#if SHADER_API_VULKAN
+#if SHADER_API_VULKAN || SHADER_API_METAL
 uniform float4 _LightShadowData;
 #define _LightShadowData _LightShadowData_KNOCKED_OUT
 #endif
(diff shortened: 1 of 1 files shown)


huwb:
Nice one!

Understood about the compute shader. I'm not super against dropping 2018, but there'll inevitably be people using 2018 so I wouldn't do it unless pushed I guess. Let me know if that makes sense.

moosichu:
Yeah, not saying we should either. Even if we did drop 2018.4 support I would only tho k it would be worth to begin considering that once 2019LTS has been released and is stable. We started supporting 2018.4 as our baseline when it was at that point as well.

PR #460, opened 2020-04-02
Feature/shadowing

nsalahragab:
Hello,
May I know how to add shadow and reflection here
image
I am trying to find out how but I can't. Thank you

Settings
image


huwb:
Hello! It looks like you have created a pull request, to make code changes to Crest. I'm guessing this is not what you intended :).

If you dont mind could you please post this as an issue in the issue section? https://github.com/crest-ocean/crest/issues

There is a template for issues that has some questions including which version of crest you're using? We have 3 fairly different versions so it will depend on which version you're on, so the answer will depend.

PR #463, opened 2020-04-04
Fix ocean depth disabled state properly

daleeidd:
Previous attempt: #432 

I wasn't diligent with this one. It fixed the foam problem but nothing else. The issue that remained is when the ocean depth is disabled, the value would be clamped to 0-1 due to not being an HDR texture.

I tested it on Metal, DirectX and Vulkan. I am using the [RGB9e5Float](https://docs.unity3d.com/ScriptReference/TextureFormat.RGB9e5Float.html) format. It appears to be well supported.

The alternative is to map the depth values to 0-1 instead.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Helpers/TextureArrayHelpers.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Helpers/TextureArrayHelpers.cs
@@ -41,6 +41,19 @@ namespace Crest
             );
         }
 
+        public static Texture2D CreateTexture2D(Color color, TextureFormat format)
+        {
+            var texture = new Texture2D(SMALL_TEXTURE_DIM, SMALL_TEXTURE_DIM, format, false, false);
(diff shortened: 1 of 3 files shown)


huwb:
Thanks, looks good, i only have a query about the format. Why the move to this one? Just wondering how it will compare with halfs.

daleeidd:
We need an HDR format since the depth is 1000. Otherwise, the value maxes out at 1. This format seems well supported. It's only for the null texture.

I'm not sure if there is another approach to getting 1000 R value using textures. Normally we are dealing with render textures which appear to behave a little differently with regards to storing HDR values.

huwb:
Sorry this fell off my radar. Thats fair. I'm vaguely curious if decimal 1000 can be represented by a 9e5 float exactly, but it probably does not matter. Belated thumbs up from me..

daleeidd:
I checked RenderDoc and the value is 1000.

Thanks. Will merge.

PR #464, opened 2020-04-04
Ouch - fix a ton of unprotected logs, all of which may produce NaNs

huwb:
Status: Ready to merge.

The all time silliest issue in crest. Theres a bunch of cases where we can compute log(0). This is a bad time as it gives -INF on hlsl.

This might be causing particular badness on metal - some instrumentation i added for someone reported a NaN coming from the height queries, which led to finding this. Hopefully this change helps or resolves the issue, but in any case it is necessary.


huwb:
@daleeidd merged

PR #468, opened 2020-04-05
Further Shader HLSL Refactor

daleeidd:
Continues on from #414 #434 

- Remove duplicate functions so both `OceanLODData` and `OceanHelpersNew` can be included together without conflicts.
- Use the new `PosToSliceIndices` with fixes.
- Brought over nearly all of the whitespace formatting changes from HDRP branch.
- Using the `CREST_` prefix for the internal shader keywords (I could use the `_INTERNAL` postfix?)


huwb:
Thanks @daleeidd! Great that you went so deep on unifying things.

I think postfix might be a good shout, yes.

With that could you please give an overview of the approach with keywords - which ones we have (not every single keyword, maybe just the foam ones as an example) and which are switched on by code vs user on material. Would be good as a refresher and also for longer term documentation.

daleeidd:
I had a look and there is only two keywords we enable using scripts which are the `CREST_FLOW_ON_INTERNAL` and the `CREST_DYNAMIC_WAVE_SIM_ON_INTERNAL` in the `ShapeCombine.shader`.

The other keywords are either enabled on the materials or are enabled by the kernel selection which is done in C# (the kernel selection that is).

I have redone the commit to only rename the ones in the `ShapeCombine.shader`.

huwb:
Ok, just to confirm theres no collision between user-enabled and code-enabled keywords? Like the case where flow was being forced on by code, but didnt work in standalone builds? 

I don't there is but just 100% confirming.

With that then thumbs up from me for merging!

daleeidd:
I just double checked. No conflicts anymore.

I also changed `_DIRECT_TOWARDS_POINT` to `CREST_DIRECT_TOWARDS_POINT_INTERNAL` since it is enabled with scripting.

daleeidd:
Thanks! Merged

PR #472, opened 2020-04-10
Fix Shallow Subsurface Scattering popping

daleeidd:
I was noticing popping in VR. Narrowed it down to the shallow subsurface scattering.

Popping

Pops exhibited with camera transform X and Z values between 39.201 and
39.205. 0.0001 seems to eliminate it entirely.

It probably wasn't that noticeable in 2D with certain ocean material settings. Specular etc can mask the problem enough. And 2D type of movement would probably skip past it.

It affects both BIRP and URP. Not sure about SG.

### DIFF
--- a/crest/Assets/Crest/Crest/Shaders/Ocean.shader
+++ b/crest/Assets/Crest/Crest/Shaders/Ocean.shader
@@ -329,7 +329,7 @@ Shader "Crest/Ocean"
 				}
 
 				// Data that needs to be sampled at the displaced position
-				if (wt_smallerLod > 0.001)
+				if (wt_smallerLod > 0.0001)
 				{
 					const float3 uv_slice_smallerLodDisp = WorldToUV(o.worldPos.xz);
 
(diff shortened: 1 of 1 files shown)


huwb:
Which type of VR is this? we currently only support Single Pass

daleeidd:
Singe Pass

daleeidd:
It's not a VR issue BTW

huwb:
Oooooh ok. I'll try this later on once i have my urp project upgraded

huwb:
Ha i saw it! good find :). I dont think this should block the next release if you agree, but great to have this one pinned down! Probably some imprecision with the vertex morphing stuff.. I'll reply when if i find more on it.

daleeidd:
Sure thing. I'm fine with leaving it out

daleeidd:
@huwb I am pretty certain I noticed this in HDRP too. Not sure if it is high enough priority for 4.2.

huwb:
Ok noted. I ran this test on the ocean in edit mode branch (thought debugging would be easier there) and couldn't seem to repro the issue so a retest would be useful

daleeidd:
I can still reproduce it. I just tried on BIRP master. It is more prominent on URP because the shallow SSS is much stronger. I recommend disabled shadowing, specular and transparency to make it more noticeable.

huwb:
Could I get data pushed so that it's set up with the fail case?

daleeidd:
Done. For some reason the old camera position I gave no longer reproduced it. I had to go into URP to find a new reproducible case, and copied that into BIRP. I tried to reproduce in HDRP but couldn't so it might be ok. You should see the following:

Screen Shot 2020-05-13 at 9 11 12 pm

There is an _Enable Fix_ option on the material (at the top). It is currently disabled.

daleeidd:
I am 100% certain it is in HDRP since I noticed it the other day. A user has reported too. From HDRP:

unknown

huwb:
Thanks for this. I messed around with it for a while. I'm guessing its some kind of unfortunate interaction between meshing and the boundaries of the cascades or positions of the texels. Changing the lod data resolution changes the behaviour.

I don't see any issue with dropping the threshold to 0.0001, so i would propose we do that. Does this sound good to you? If so let me know if you want to finalise this PR or if you want me to do it.

daleeidd:
Thanks. Sounds good. I can do it.

huwb:
Perfect!

PR #474, opened 2020-04-15
Bugfix: Remove collider that gets added by ShapeGerstnerBatched

skaughtx0r:
ShapeGerstnerBatched is now creating a Quad Primitive to render to. When this quad primitive gets created it has a Collider on it. Unfortunately, on some of my race courses 0,0,0 happens to be in the middle of the course and this collider is in the way.

Here's a video one of my users sent me:
https://gyazo.com/2c4e33bb7186f6270e604f157278c809

This PR removes the collider component from the primitive game object that gets created.


moosichu:
Thanks for this one! looked at the documentation and it looks like this is the case!

https://docs.unity3d.com/ScriptReference/GameObject.CreatePrimitive.html

I think we will merge this, but it might just be worth creating a quad primitive ourselves using CreateMesh? What do you you think @daleeidd @huwb?

huwb:
Thanks @moosichu , its a good point but the pattern posted in this PR is fairly standard when creating primitives in unity, i'd merge in the change as is rather than going deeper if thats fine.

moosichu:
Sorry, I just realised my comment was confusing. I meant "let's get this merged, but later, after that, we could come back and cleanup if we really felt the need to do so".

skaughtx0r:
@moosichu I thought about just creating the mesh myself, but it would have been a lot of extra code, since you'd need vertex, triangle, and uv arrays. I figured a one-liner would be easier to get merged.

PR #477, opened 2020-04-21
Show a proxy plane at the ocean sea level

huwb:
**Status:** ready to merge

Creates a transparent plane at the ocean height in the scene at the sea level, to use as reference.

image

Creates a hidden, nonsaved transparent plane gameobject. Done this way because it doesnt seem to be possible to have gizmo drawing intersect with the scene.

Only creates if in edit mode and proxy is enabled.

Tests: in win64 editor, and made win64 standalone build.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
+++ b/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
@@ -2,6 +2,7 @@
 
 // This file is subject to the MIT License as seen in the root of this folder structure (LICENSE)
 
+using UnityEditor;
 using UnityEngine;
 
 namespace Crest
@@ -103,6 +104,9 @@ namespace Crest
 
         [Header("Debug Params")]
 
+        [SerializeField]
+        bool _showProxyPlane = true;
+
         [Tooltip("Attach debug gui that adds some controls and allows to visualise the ocean data."), SerializeField]
         bool _attachDebugGUI = false;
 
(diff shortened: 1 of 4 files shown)


daleeidd:
This is very useful. I have done this myself for my own project albeit not as nice as this. I also have additional planes to show each depth where attenuation affects each wavelength. Been useful for designing terrain.

huwb:
> I also have additional planes to show each depth where attenuation affects each wavelength

Sounds interesting, could you speak more to this? I didnt follow

daleeidd:
planes

The game object names are the wave length and the Y value is the depth attenuation begins for that wavelength. Hopefully I haven't gotten how that works wrong.

When designing terrain, it is can be useful to visualise how the terrain will affect the ocean. You should see blue layers in the scene view.

huwb:
Ha interesting.

I think waves start getting affected by sea floor when the depth of the water is less than half of the wavelength. I think that means that the "64" should be 32m from sea level. I'm not sure if that is still a useful guide at that point.

I think i'd propose to merge the above as is for now. Let me know what you think.

daleeidd:
> I think that means that the "64" should be 32m from sea level.  I'm not sure if that is still a useful guide at that point.

Yes. That's what I have. I might just be obsessive about how the waves will be affected.

> I think i'd propose to merge the above as is for now. Let me know what you think.

By all means. I didn't mean for this to hold anything up.

huwb:
Oh i think i just fully understood what your screenshot shows @daleeidd - the lower planes are below. That's really interesting.

I'll merge this as is anyway. Thanks

Issue #456, opened 2020-03-31
[HDRP] Horizon line issue


huwb:
**Describe the bug**
Sometimes a visible line appears at the horizon.

In the screenshot below there is a horizon line both above and below water. We may find these are two separate issues and break them out accordingly.

**Screenshots / video**
![image](https://user-images.githubusercontent.com/939901/78024916-0fa15d00-7351-11ea-87ab-1d7daab9c19b.png)

**Versions**
2019.3 HDRP

**To Reproduce**
*How can the developers reproduce this bug? Are you able to reproduce the issue in the example content/scenes?
Please include steps to reproduce the behaviour, with the example content ideally so we can follow the steps on our side.*

**Platform**
*Editor or standalone? Platform (win/mac/...)*

**Hardware**
*What hardware are you seeing the issue on? GPU model is particularly useful.*

**Additional context**
*Add any other context about the problem here.*

daleeidd:
It is very difficult to reproduce in the main scene with the provide ocean renderer settings. But if you reduce the quality (like reducing the LOD resolution to 256) it is much easier. So this could be a seam where the tiles end?

daleeidd:
![Screenshot (138)](https://user-images.githubusercontent.com/5249806/78218820-24752080-750a-11ea-8b9c-0f4f16044317.jpg)
Game View

![Screenshot (137)](https://user-images.githubusercontent.com/5249806/78218826-25a64d80-750a-11ea-969d-40d587aaed70.jpg)
Scene View

I set the LOD to 256. That is also the default value. But the main scene is set to 384.

huwb:
Oh so there are legit cracks in the water surface?? Ok.

I've had it on projects that precision errors generate cracks like this. Sometimes just scaling the tiles by 0.5% more effectively covers this stuff. However i think id like to take a closer look at hte above is it could be abug, or something that could be improved in the vert shader.

skaughtx0r:
This bug is not limited to HDRP, I've been noticing this in the Standard pipeline version as well. One of my game's users said the turning off MSAA fixed it...

moosichu:
Isn't this caused by the underwater post-process effect? It was one of the reasons why it hadn't been merged-in for BIRP, but we had never seen it in HDRP ourselves.

skaughtx0r:
Oh, yeah, I'm using the underwater post-process branch, so that is likely the issue I'm seeing.

banan5555:
[...]

PR #481, opened 2020-04-23
Compute LOD black point fade

daleeidd:
Computes the LOD black point fade from the mesh density. By computing the value we reduce gaps between tiles for lower mesh densities and reduce popping for higher mesh densities.

First discussed in #456. It helps solve one cause of that problem.

The formula is based on hand crafted values that produced no gaps when ocean
is flat. It works pretty well when displaced too.

The approach to getting these values is to have a range parameter which sets the black point fade value for easy testing. The ocean needs to be flat. Then in play mode, switch to the scene view to show the camera and surrounding ocean tiles. Moving the camera around will expose gaps (even vertically). Change the black point value and repeat until no gaps appear.

It is still not 100% perfect as displacement can still produce gaps which can be noticed underwater if looking out for them (they appear as a few pixels here any there). Flying underwater with the simulation frozen is a good way to reproduce these gaps. I couldn't find a value prevent them all. I got to a point where higher or lower value would produces gaps sadly.

I found that a mesh density below 16 produces too many gaps too often to be useable.

Good values (it was hardcoded as 0.15):
- 008: 0.376
- 012: 0.250
- 016: 0.188
- 024: 0.126
- 032: 0.094
- 064: 0.047
- 128: 0.023
- 160: 0.018

The formula gets it pretty close to these values.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
+++ b/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
@@ -158,6 +158,10 @@ namespace Crest
         readonly int sp_oceanCenterPosWorld = Shader.PropertyToID("_OceanCenterPosWorld");
         readonly int sp_meshScaleLerp = Shader.PropertyToID("_MeshScaleLerp");
         readonly int sp_sliceCount = Shader.PropertyToID("_SliceCount");
+        readonly int sp_blackPointFade = Shader.PropertyToID("_CrestBlackPointFade");
+
+        float _baseMeshDensity;
+        float _blackPointFade;
 
         void Awake()
         {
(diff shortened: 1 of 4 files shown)


huwb:
Thanks, think im a thumbs up for this.

Its a shame the cracks could not be removed completely and i dont really get why. However it sounds like this addresses the worst of it so lets get it merged.

As a tiny optimisation i wondered if passing BlackPointWhitePoint instead of compting one of them was worth it. Only saves a single subtract but would be simple and clean to implement so wondered if it was worth considering. Thumbs up for me in any case.

daleeidd:
> BlackPointWhitePoint

Good idea. Done

I investigated the problem underwater gaps further. They are only noticeable underwater. They appear as folds.

Crack Game View Underwater
From the game view. They can appear slightly bigger than this

Crack Scene View Under
Up close with the scene view

Crack Scene View Above
From above with the scene view

Crack Scene View Side
From the side with the scene view

Changing the fade value only pushes the problem somewhere else. There might be a perfect value but I couldn't find it.

I was thinking that maybe there is a way to handle underwater differently. With possible changes to tile culling it could solve itself.

PR #484, opened 2020-04-25
Move far plane calculations to the CPU

moosichu:
@huwb did the legwork for this one in another branch that was then back-ported to BIRP. The plan here is to ensure that the flickering horizon line no longer appears in the distance by moving the calculations for that to the CPU.

It's less of a problem than it was before, but it's still present. Unfortunatley single-pass rendering doesn't work with this either, even though I believe that I have made the requisite changes to ensure that it should work.

The horizon line can still appear incredibly rarely unfortunately, the next step is going to be trying to move some of the calculations to doubles of possible.


moosichu:
Example of poor horizon lines with VR single-pass enabled.
image

image

moosichu:
Merging, even with VR issues as I plan to tackle all of that in a seperate pass once other underwater refactorings are complete.

Issue #208, opened 2019-04-01
Execute in Edit Mode


SilverbackKarthik:
Can we have the Ocean be generated in Edit Mode similar to Unity Particle System?

![image](https://user-images.githubusercontent.com/42980551/55348584-e031e900-548d-11e9-9c19-39a4521db8f5.png)

The reason behind this is that the project we are using this for is based on Unity Timeline and we would like to scrub the timeline in edit mode and see it with the waves. Even a low poly mesh based render would help visualize it.

I am making using of the Time Provider with the Timeline time to scrub in runtime mode now.

huwb:
Yeah, probably. I did some experiments where i slap [ExecuteInEditMode] and set HideFlags to DontSave on the ocean components/generated GOs and it gets a bunch of the way there, but there might be a fair bit of fixup required to get all the state figured out to work in edit mode and be robust etc. I'll post back if i make progress.

SilverbackKarthik:
Sounds great! I have been trying to work it out. This has been a most requested feature by our team do I have been trying to figure outs ways to get it working. Also, I have been trying to get Bow Wakes without having the objects move. Let me know if you have any pointers for that as well.

huwb:
I put a bit more time into it and was surprised to see something kinda working. There are resource lifetime issues and usually some errors creep into the log (at which point i'm currently reopening the scene) but the basics seem to be there. I'd be very interested if it already works with the Timeline and what kinds of functional tradeoffs it currently has, if you'd be in a position to grab it and try it with your project? It's in the branch feature/ocean-in-edit-mode

huwb:
Perhaps open a separate issue for bow wakes if you dont mind so we dont mix issues together. If you could outline what you're doing/what you want a bit more that would be useful.

SilverbackKarthik:
That seems to be working better than what I thought it would.
We could make a few improvements to it and have it working in a more optimized manner.

Yes. I will create a separate Issue for the Bow Wake feature,

moosichu:
@SilverbackKarthik are you working on this at the moment, I'm considering taking this on unless you are happy completing this. 🙂 If you've done any further work so far I am happy to take things on from there.

SilverbackKarthik:
@moosichu Yes, I am using it in a project and trying to fix few optimization bugs. But, have not had the time to test them through. So, If you are starting to work on this, It's all yours. I would like to get updates on your work as well.

So, I guess thread could be a place to share the info.

[...]

PR #485, opened 2020-04-26
Ocean in edit mode

huwb:
**Status:** Ready to merge

This started in #208 and has finally become shippable.

Changes required to support ocean in edit mode, at a high level:

* LodData - not monobehaviours anymore. Built up/torn down on demand.
  * Default settings created under the hood and used if none are specified
* Ocean update - in edit mode driven by editor update function. Custom time, deltaTime and frameCount implemented/emulated
  * Ocean update driven from a central place (OceanRenderer) rather than being distributed. This added some code but its nice that its explicit now and does not rely on execution order.
  * Update is only realtime in scene view if "Animated Materials" option is enabled, this is expected/normal I believe
  * Gerstner update done on demand instead of driven by LateUpdate
  * BuildCommandBuffer un-monobehaviour'd
* TimeProvider has default under the hood, but can be overriden with a manually specified component
* Dont change data. Don't add components (but if we do, make them nonsaved). Dont change data on components. These things are bad when they run in edit mode and mess with scene data.
* Stop requiring two material instances for each ocean input, which is great and simplifies things a bunch
* Remove the ocean proxy plane we recently added as its not needed

These changes generally make the system much more dynamic - since construction is manually handled and dynamic cases or detected, things can be built up/torn down at runtime which open the door to optimisation opportunities.

What is not great right now is that its hardcoded to follow a particular viewpoint. One day it would be nice if crest followed how postprocessing works - adding a PP effect to a generic list creates default data that can be overridden. maybe one day everything is applied to post processing volumes..?).

**Issues / todo**

- [ ] ~~QueryBase - un-monobehaviour? Move to anim waves?~~ **UPDATE:** taking a lazy approach and leaving as is - for now
- [x] Error related to flow in compute shaders
- [x] Rebuild code - loses ocean inputs on assembly reloads?
- [x] Ocean input duplicates the material, so tweaking input material settings has no effect in edit mode
- [x] Deal with cameras - follow scene view?
- [x] Shadow data?
- [ ] Gerstner component writes wave data onto serialised properties which is naughty and will generate data changes. Couldn't figure out a low impact way to fix this so i propose we revisit later.
- [ ] Doesnt yet support visibility/pickability toggles in 2019

**Testing**

2018.4, Win64 editor all scenes, win64 standalone.


moosichu:
Having few calls to MonoBehaviour Update() is good for performance as well! Really looking forward to trying this out when it's ready! :)

huwb:
Hi @daleeidd  @moosichu , at your leisure could i please get testing on this? It's close enough now to being ready for prime time, so testing should give meaningful results.

There is one error about texture dimensions not matching on Flow texture array or stuff like that. I know what the issue is, but dont quite know the best solution. I'll probably raise a discussion about it later, its safe to ignore for now.

huwb:
Beleive it or not i think this is ready to merge. Thoughts/tests welcome. And/or would be good to get this into mainline to get some mileage with it before the next releases.

daleeidd:
Overall works great. I have to enter and exit play mode for LOD resolution changes to take effect. Not sure about other detail parameters.

For the camera, have you tried adding a temporary camera which follows the scene camera?

**Windows**
I had to upgrade to 2019 to get this to run without hitching and freezing the editor. Has that been the same for you?

**Mac**
Same as Windows. But hitching is still an issue after upgrading to 2019. Not as severe as 2018 though. It only happens after I end scene view navigation (so letting go of trackpad or right mouse).

huwb:
Thanks!

Good point about detail params, that seems fixable.

Would be possible to keep proxy plane, but was hoping this would be stable and solid enough to render it redundant.. doesn't sound like we're there yet!

I could attach it to the scene view but there were some artifacts, some 1 frame glitches and the geomorph seemed to break and I couldn't figure out why, so far..

I ran everything on 2018.4, no hitches or freezes :/. Did they start occurring immediately? Would you mind staying fresh in 2018.4 - maybe delete library folder to reimport everything, and see if this changes anything?

daleeidd:
> Would be possible to keep proxy plane, but was hoping this would be stable and solid enough to render it redundant.. doesn't sound like we're there yet!

Just from my own situation, I would have disable this feature by editing Crest's source just to use it because of those hitches I mentioned a while ago for MacOS specifically.

But it needs to be considered also with less powerful hardware generally. Terrain Tools can slow down the editor quite a lot. I'm sure there is X asset on the asset store which would suffer and hinder someone's workflow somehow on some cheap hardware.

Another asset I was using was OverCloud. It runs in the editor too by default. If I didn't fire up the eGPU before opening Unity it would crash the OS when I opened my project. So I could no longer do some casual work on my project. That asset was far more resource hungry than Crest though.

Then again, running in the editor is pretty standard. If I was smart (or not so partial to MacOS) I would use the better hardware available to me with Windows. But I was evaluating Unreal and I could run some heavy stuff in the editor on MacOS without issue. It felt safe doing so. But they provide ways to not render anything in either edit or play mode.

If you do bring it back, maybe a scriptable object which stores some global settings about Crest?

> I could attach it to the scene view but there were some artifacts, some 1 frame glitches and the geomorph seemed to break and I couldn't figure out why, so far..

Ah, yes. I see it too. It also happens if I move the demo camera around using the handles.

Worked it out. It's the depth cache. Works fine if I disable it or set it to Baked. My guess is switching between the cameras to render the cache causes the issues?

> I ran everything on 2018.4, no hitches or freezes :/. Did they start occurring immediately? Would you mind starting fresh in 2018.4 - maybe delete library folder to reimport everything, and see if this changes anything?

It does. It is constant. It looks like it pauses every other moment even without interacting with the editor. Are you still using 2018.4.0f1? On 2019 it goes away unless I look at the game view which exhibits the freezing behaviour. It tried cleaning and re-importing as suggested but it still occurs.

---

I tried it with the terrain tools and it was really nice to see it respond well to edits. I wasn't able to select the terrain through the ocean though. The tiles shouldn't be selectable.

I think it should also respect the visibility toggle in the hierarchy view (left).

The following might be useful:
[Hierarchy](https://docs.unity3d.com/Manual/Hierarchy.html) (near the bottom)
[Scene Visibility](https://docs.unity3d.com/Manual/SceneVisibility.html)
[Scene Pickability](https://docs.unity3d.com/Manual/ScenePicking.html)

huwb:
Great notes thanks, will respond to all soon, but just quickly:

Re freezing - it's intended to not update every frame for power reasons. Maybe that's what you're seeing? If you fly around with the camera or staff the mouse on the game view, does it update smoothly? 

Alternatively if you turn on animated materials in the drop down list at the top of the scene view, is it smooth then?

There's also an FPS control on the ocean component to throttle the update rate as well.

huwb:
> Worked it out. It's the depth cache. Works fine if I disable it.

Wow yes! Disabling the depth cache make it follow the scene view perfectly and its great.

Just pushed a change where it should manage to switch between following game and scene cameras :). Loving this.

I havent looked at the depth cache issue yet so it will glitch if its enabled for now

daleeidd:
> Great notes thanks, will respond to all soon, but just quickly:
> 
> Re freezing - it's intended to not update every frame for power reasons. Maybe that's what you're seeing? If you fly around with the camera or drag the mouse on the game view, does it update smoothly?
> 
> Alternatively if you turn on animated materials in the drop down list at the top of the scene view, is it smooth then?
> 
> There's also an FPS control on the ocean component to throttle the update rate as well.

Yes. Sorry. I edited the comment probably as you were replying. I didn't have animated materials on. Enabling it solved it. It is on by default for 2019. I think we should add validation for it. [SceneViewState](https://docs.unity3d.com/ScriptReference/SceneView.SceneViewState.html)

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
```diff
@@ -11,17 +12,54 @@ namespace Crest
...
         [Tooltip("The viewpoint which drives the ocean detail. Defaults to main camera."), SerializeField]
         Transform _viewpoint;
-        public Transform Viewpoint { get { return _viewpoint; } set { _viewpoint = value; } }
+        public Transform Viewpoint
+        {
+            get
+            {
+#if UNITY_EDITOR
+                // If scene view is focused, use its camera. This code is slightly ropey but seems to work ok enough.
+                if (EditorWindow.focusedWindow.titleContent.text == "Scene")
```
daleeidd, line 26: `EditorWindow.focusedWindow && ` need to check that it exists

daleeidd: Also, this will lose focus the moment I touch an inspector. We might need to store state.

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
```diff
@@ -21,15 +21,17 @@ public Transform Viewpoint
...
-//                }
-//#endif
+#if UNITY_EDITOR
+                // If scene view is focused, use its camera. This code is slightly ropey but seems to work ok enough.
+                if (EditorWindow.focusedWindow.titleContent.text == "Scene")
+                {
+                    var sv = SceneView.lastActiveSceneView;
+                    if (sv != null && !EditorApplication.isPlaying && sv.camera != null)
+                    {
+                        return sv.camera.transform;
```
moosichu, line 31: I think making this feature optional would be a good idea, because I would want to be able to inspect the lods from the scene view in some situations based on where the attached viewpoit is located.

moosichu:
image

Got these on first startup, I'm also getting an incredibly hitchy experience. (Although it's awesome that it works!)

I have to go now for a bit but definetely plan on putting this through it's paces :) Really excited about this one!

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
huwb: Good point! Attempt pushed now, seems to work now?

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
huwb: Yeah that makes sense, made a proposal on latest ,let me know if this works

huwb:
Hitchy is by design/a unity feature to save power, see conversation above. @daleeidd on this we could make a validation message appear on the ocean renderer component once the help box stuff is in? I would be hesitant to print something in the log as i'll be running with the animation disabled when using my laptop, so it will be the default for me.

Re errors, thats odd! Line 85 is this for me. Which line is excepting for you?

image

moosichu:
When I mean hitchy, when I try to move the camera around the scene view it's no longer smooth. It jumps around quite a lot. One sec, making a recording.

moosichu:
OK, I think I know what the problem was. I had a very-very old version of this branch. and it wasn't updating. I think it might have even been from the previous attempt at implementing this? Please ignore everythign that I have said so far.

moosichu:
Comprehensive thoughts and feedback:
- Whilst having the ocean "play" in editor is nice, I think for this production ready, we need the following features:
   - The ability to turn this off completely (and have the plane instead?)
   - The ability (and I think this should be the default) to have time remain static in edit-mode: I want to default to seeing the scene in it's initial state. (Or this is what I would want if I used Crest in my own project). 
   - Default to having the LODs follow the provided view-point, not the scene camera.
- This change results in a lot of tiles being injected into the scene hierarchy that then get saved, this introduces the potential for scene merge conflicts which shouldn't really be there. Is this a problem that is solvable, can we generate the tiles in a way so that they don't get serialized?
- I get seams when moving the scene camera around
image (The ocean also jerks-around a lot)

Loving this change though and really excited for it! :)

huwb:
@daleeidd i think i fixed the glitchiness around the depth cache - it was another call to set the _OceanCenterPosWorld from the cache code. This was only there to fix a case at runtime where it the sea level was wrong, and might be BIRP only. Disabling this call in edit mode seems to have sorted it.

@moosichu your notes:

Disabling ocean in edit mode, yeah dale requested this too. can look into this. I'm not sure other ocean solutions would do this but if we're sure we want it then i can add the code back.

Not sure about freezing time by default (it might be possible using a custom time provider). could you explain why you would want this?

I think i disagree about the default to the Viewpoint transform - i think it should generate a representative result for whatever the current view is by default. It always confuses people when they see the ocean from a non-view-centric angle, if they dont know whats happening. as discussed above i've added a "follow scene camera" option which enables this kind of behaviour, hopefully that ticks this box enough if you agree?

Saving tiles in scene - are they? they shouldnt..

Seams and jitter - should be fixed by my latest push (its related to the depth cache issues that dale found)

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
daleeidd: Works great

moosichu:
Seems and jitter seem fixed. I already was on latest but had to restart Unity.

Regarding time freezing, it's a personal thing, and whilst I could make a time provider to solve the problem - just seeing anything move in edit-mode can be quite distracting. For me when editing levels I would want the state of the scene in edit-mode to represent the state of that scene at startup - so with the time values that the game would have at startup.

Now that the simple follow-camera toggle is-in, I'm happy to leave the default behaviour as-is.

It's also a bit odd that in our demo scene the default behaviour is now that the ocean moves, but only when you seem to interact with the editor which is quite jarring.

daleeidd:
> on this we could make a validation message appear on the ocean renderer component once the help box stuff is in?

Sounds good. I'll add something so we have only some validation run for the help boxes. It makes sense to do so anyway since some things only makes sense for GUIs only.

> i think i fixed the glitchiness around the depth cache

That works well now. Great.

moosichu:
Regarding tiles being serialized, if I change the LodCount, hit play and stop and then save the scene the tiles seem to all be stored in the scene file (I get a humongeous diff).

crest/Assets/Crest/Crest/Scripts/Collision/QueryDisplacements.cs
```diff
@@ -21,15 +21,15 @@ public class QueryDisplacements : QueryBase, ICollProvider
...
             Instance = this;
 
             base.OnEnable();
         }
 
         protected override void OnDisable()
         {
-            Instance = null;
+            // If we set this to null, it breaks exiting play mode, as OnDisable is called but no matching call to OnEnable :/
+            //Instance = null;
```
moosichu, line 32: Should we just put a comment here saying "we don't sent Instance to null because...", and delete the commented-out line of code?

crest/Assets/Crest/Crest/Scripts/Collision/QueryFlow.cs
```diff
@@ -21,23 +21,26 @@ public class QueryFlow : QueryBase
...
             Instance = this;
 
             base.OnEnable();
         }
 
         protected override void OnDisable()
         {
-            Instance = null;
+            // If we set this to null, it breaks exiting play mode, as OnDisable is called but no matching call to OnEnable :/
+            //Instance = null;
```
moosichu, line 32: Same as previous comment. Might be worth tagging these if it happens in multiple place.

huwb:
Thanks both for all the notes so far!

Yep theres a few things saving into the scene. D'oh! thanks i'll hunt them down..

@moosichu another option regarding updating in scene view is to set Edit Mode FPS on the ocean component to 0.1 which more or less stops it updating (i could let this go to 0 to truly stop it updating as well?)

> So the only remaining issue with animating is that the game view doesn't run correctly in realtime when not in play mode unless I hold the mouse down on it. Low severity.

I did look into this a bit but propose to leave it as is. Unity doesnt want it updating smoothly as the simulation is not running in the game view when in edit mode. it might be possible to stop it updating at all but if we're ok with it i'd leave it as is.

@daleeidd i'll keep working through your notes in a subsequent post

crest/Assets/Crest/Crest/Scripts/Helpers/TextureArrayHelpers.cs
```diff
@@ -80,14 +92,17 @@ static void InitStatics()
             // Init here from 2019.3 onwards
             sp_LD_TexArray_Target = Shader.PropertyToID("_LD_TexArray_Target");
 
-            if (BlackTextureArray == null)
+            if (_blackTextureArray == null)
             {
-                BlackTextureArray = CreateTexture2DArray(Texture2D.blackTexture);
-                BlackTextureArray.name = "Black Texture2DArray";
+                _blackTextureArray = CreateTexture2DArray(Texture2D.blackTexture);
```
moosichu, line 97: We shuld probably just access the public getter or collapse this code into a function that is called from two places.

moosichu:
I was actually looking at the threeboats scene, and hopefully this might explain why a frozen ocean could be good:
- Placement of boats on specific waves for when the scene starts. If you are authoring a scene you might want to place bouyant objects in very specific locations relative to the ocean surface (one of the big practical benefits of having the ocean in edit mode that you couldn't do before!). If you have fairly large waves, you will need to be able to see the ocean in the initial state in order to be able to do this.

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
```diff
@@ -290,7 +290,7 @@ static void EditorUpdate()
 
             if (!EditorApplication.isPlaying)
             {
-                if (EditorApplication.timeSinceStartup - _lastUpdateEditorTime > 1f / Mathf.Clamp(Instance._editModeFPS, 0.1f, 60f))
+                if (EditorApplication.timeSinceStartup - _lastUpdateEditorTime > 1f / Mathf.Clamp(Instance._editModeFPS, 0.01f, 60f))
```
moosichu, line 293: Could pre-pend:
-> `Instance._editModeFPS > 0`?

Also, 60 FPS can be quite a low upper-limit, people who like using the 144Hz monitors at 144Hz are *really* passionate about it. (I know of at least one instance where a game received one really bad review just because they assumed 60FPS would be the max). Maybe 240Hz would be a good upper-bound? (or at least don't clamp it here, so that people with know-how can set the _editModeFPS value themselves without it being auto-clamped).

huwb: The way the code is now, it ensures that one update will happen, so that the ocean is in a more or less presentable state. I could add more state to ensure this but was tempted to take this lazy route.

good point about higher freqs, but I'd need to check that the editor runs at >60hz, which i'm not sure i can do on this laptop.

huwb:
I think that level of synchronisation would be done with a time provider driven by a timeline or such. A typically game would have an arbitrary game time when entering a scene (depending on load time, time spent in menus, etc), so its not clear to me that locking it to 0 or some other time would help. If precise choreography is required one must use a timeprovider. This workflow with the timeline will work in edit mode, which is awesome.

I've made the fps slider go down to 0 so you can effectively stop updates that way, if convenient. Added a tooltip as well.

I stopped a bunch of things saving - guess i didnt check carefully before! Unfortunately, since i'm moving the ocean GO around, this changes scene data. Anyone have ideas on this? :/ I suppose having the option to disable this feature is one (not good) solution.

moosichu:
Cheers, thanks Huw. That's a very good point and I'm happy with where things are at. :) Yeah, I've never had to deal with not serializing stuff in the scene in edit-mode before. So nothing comes to mind.

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
moosichu: image

I don't know if this helps.

moosichu: https://forum.unity.com/threads/how-to-setup-the-editor-to-run-at-144hz.797001/

It seems to be supported at least. It's a niche use-case for sure though. Although with the growth of variable-frame-rate monitors I imagine that more-and-more people might start caring.

daleeidd:
> Unfortunately, since i'm moving the ocean GO around, this changes scene data. Anyone have ideas on this? :/ I suppose having the proxy plane option available again is one solution, although its a bit indirect..

The only two things I can think of at the moment is to either revert the position after rendering (not sure how that would work with Crest) or we attach a child object to the ocean renderer serve as the transform.

huwb:
Changing the number of lods and other construction params - do we need to respond to these on the fly? Feels like we might do. D'oh :).

I guess an alternative would be to display a message that the settings are unapplied and ask them to enter play mode? too clunky?

crest/Assets/Crest/Crest/Scripts/Collision/QueryDisplacements.cs
huwb: sure

crest/Assets/Crest/Crest/Scripts/Helpers/TextureArrayHelpers.cs
huwb: good call, factored it out, thx

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
huwb: I feel like people are going to see the slider go up to 144 and then complain to me that the editor is only running at 60hz. would it be ok to leave it at 60 and wait and see if anyone requests the higher rate? i like to keep slider ranges tight where possible so its quick and easy to set it to 60 which is currently the standard.

huwb:
Thanks @daleeidd i did the latter - added a 'root' under the Ocean gameobject which serves as the transform. I searched for all cases of OceanRenderer.Instance.transform and similar and think i got them all.

I added a 'Rebuild Ocean' button which can be used to apply ocean construction settings (and updated tooltips on the relevant settings).

I wasn't able to super cleanly reintroduce the proxy plane. I could hide the ocean tiles and early out of running the ocean sim i guess, when the proxy plane is enabled. I'm not super keen to add this as the complexity of my current code in this change is already exceeds my comfort threshold a bit.. do any of the following controls mitigate the mac issues @daleeidd :

* Animated Materials option to disable real-time update
* FPS setting to throttle/disable simulation update, while keeping surface visible
* Water layer visibility toggle under the Layers dropdown, to make the surface invisible

If not then i'll try implementing what i outline above.

huwb:
> The ocean debug GUI doesn't remove itself after exiting play mode

Think this was because 'Attach Debug GUI' was enabled. If so this is what i intended, let me know if it makes sense.

Scene pickability and visibility - good shouts. i dont even have these in 2018. tempted to revist this when we drop 2018. I did set the 'selection base' to be the ocean object, so you dont end up selecting a random tile. Are you saying you could not select the terrain at all?

Think i've responded to all the existing points now. Let me know if not!

daleeidd:
> I wasn't able to super cleanly reintroduce the proxy plane

That's fine. We can always introduce it if there ends up being demand.

> Animated Materials option to disable real-time update

I don't think I would rely on this myself since the ocean runs intermittently which is jarring.

> FPS setting to throttle/disable simulation update, while keeping surface visible

This works the best I think.

> Water layer visibility toggle under the Layers dropdown, to make the surface invisible

I didn't realise this exists. Solves the problem with selecting too.

> Think this was because 'Attach Debug GUI' was enabled. If so this is what i intended, it allows inspection of the textures while in edit mode. Let me know if it makes sense.

I would expect if I didn't have the component before going into play mode, I shouldn't have it going out of play mode. An improvement would be to add/remove the component immediately based on the toggle.

I wish there was a way to make it clear a component is not being serialised. I might be able to do that with a help box as a work around.

> Scene pickability and visibility - good shouts. i dont even have these in 2018. tempted to revist this when we drop 2018. I did set the 'selection base' to be the ocean object, so you dont end up selecting a random tile. But are you saying you could not select the terrain at all?

I see. I didn't realise this was a new feature. The layer drop down solves the problem. It has picking toggles. This was a terrain I made which positioned itself below the ocean so it was out of reach.

moosichu:
Actually, I just noticed that if the ocean is inside a prefab, you can apply the newly generated root + chunks as a prefab override :/

crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
```diff
@@ -177,6 +181,7 @@ public void PopulateCache()
             if (_camDepthCache == null)
             {
                 _camDepthCache = new GameObject("DepthCacheCam").AddComponent<Camera>();
+                _camDepthCache.gameObject.hideFlags = HideFlags.DontSave;
```
daleeidd, line 184: I wonder if we should use `HideAndDontSave` instead. It is useful for us to see the game objects though. There is also the `NotEditable` flag which can also make it clear that it is generated. Or perhaps even just add something to the name like "(Instance)".

huwb:
I had a stab at bringing back the proxy plane and it wasn't too bad (if you guys agree..) so i pushed it. It hides the ocean geo and stops the sim updating.

crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
huwb: I made it hidden by default, but with an option to show it. Hopefully that ticks this box

crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
```diff
@@ -124,6 +121,7 @@ public bool Validate(OceanRenderer ocean)
     /// <summary>
     /// Registers input to a particular LOD data.
     /// </summary>
+    [ExecuteAlways]
```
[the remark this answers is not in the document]
huwb: Thanks.. so accessing rend.material makes it create one? Any thoughts what i can do to resolve this?

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
```diff
@@ -175,7 +235,16 @@ public class OceanRenderer : MonoBehaviour
...
+#if UNITY_EDITOR
+        static float _lastUpdateEditorTime = -1f;
+        public static float LastUpdateEditorTime => _lastUpdateEditorTime;
+        static int _editorFrames = 0;
+#endif
+
+        BuildCommandBuffer _commandbufferBuilder;
+
+        // Drive state from OnEnable and OnDisable? OnEnable on RegisterLodDataInput seems to get called on script reload
+        void OnEnable()
```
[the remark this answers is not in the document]
huwb: I will do this. should we never update in prefab scenes? i didnt think about it too deeply before now..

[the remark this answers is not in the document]
huwb: Hm just bailing out puts the ocean in a weird state, i had to add a bunch of null ref guards, and after that the ocean doesnt recover fully when going into prefab editing and back out. I maybe have to detect this case and trigger a rebuild of the system :/.

What were you seeing that triggered this comment? Maybe its easier to try to get the ocean running in prefab scenes.

huwb:
@moosichu do you get hitching, or is it low update rate? i understand these might look and feel similar from a users experience, but low update rate can be fixed by turning on animated materials, whereas actual hitches are a problem we need to fix. or is it related to mac issues that dale also mentioned above?

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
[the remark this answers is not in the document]
daleeidd: That's strange because that's what I get without this change. If I go into prefab mode the ocean is has no lighting (expected) and doesn't recover when leaving. I get a bunch of null exceptions.

I'll check this out with latest changes to see if something has changed..

moosichu:
Where do I find this option?

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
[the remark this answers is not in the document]
daleeidd: I get this error when exiting MainScene prefab:

```
NullReferenceException: Object reference not set to an instance of an object
Crest.OceanChunkRenderer.ExpandBoundsForDisplacements (UnityEngine.Transform transform, UnityEngine.Bounds& bounds) (at Assets/Crest/Crest/Scripts/OceanChunkRenderer.cs:169)
Crest.OceanChunkRenderer.UpdateMeshBounds () (at Assets/Crest/Crest/Scripts/OceanChunkRenderer.cs:58)
Crest.OceanChunkRenderer.Update () (at Assets/Crest/Crest/Scripts/OceanChunkRenderer.cs:52)
```

If I go into MainSceneCore from MainScene the scene camera gets mess up. Its position is far away from the island and the near clip plane is really large.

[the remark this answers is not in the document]
huwb: Oh I didn't try prefab scene before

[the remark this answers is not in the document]
daleeidd: I was bailing out right at the top of both OnEnable and OnDisable. Having it work in prefab scenes would be better, but I thought this would be a quicker solution.

crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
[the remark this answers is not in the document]
daleeidd: Right now registration requires the user to have the mesh renderer setup correctly before adding the register component. And it is very difficult to recover from there.

We would need an isValid flag just to make things simpler...

The most robust solution would be to handle all this ourselves (we add both mesh and material fields and create mesh renderer). Then we could use OnValidate to register if it fails in OnEnable. But that is a pretty significant change. The extra benefit is we can setup the mesh renderer how we know it should be and users would get updates from bug fixes. Right now we have no control.

Or we always check these values if isValid is false when it is running in the editor before using them and do some healing. This one might be a bit flakey.

Last is an update function which runs only in the editor and does a delayed registration once everything is valid. But we still need to keep check if user changes the material or something and de-reregister and set invalid appropriately.

It seems this is already a problem but it was a safe assumption not to handle it because of enter/exist play mode purging bad state.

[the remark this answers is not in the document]
daleeidd: To add to the "robust solution" point, we could try something like `Graphics.DrawMesh`, `Graphics.DrawMeshNow` or `Graphics.DrawMeshInstanced` to avoid an extra component.

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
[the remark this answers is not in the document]
huwb: Bailing out at top of both seems to work perfectly!? I cant say I have a full grasp why though. I guess it just stops the code from messing with the state when entering a prefab scene, and when it comes back it just picks up where it left off?

I'll push this as a proposal.

crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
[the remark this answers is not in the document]
huwb: I think maybe polling the data off the object each editor update instead of relying on cached data might be the way forward? I had to do this in a few other places as well. You could easily break the gameobject apart and put it back together in arbitrary ways as a user and expect the system to keep up and handle it gracefully.

Your notes about setting up the object from code probably warrants further discussion. it could be an option, but i think the user could still break the system easily if it doesnt constantly poll the data, so i might look at that for now.

huwb:
@moosichu Ah cool, sorry was out for a few hours.

If you do get a hitch that you can reproduce enough to look at in the profiler then it would be interesting to know the source.

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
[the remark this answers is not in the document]
daleeidd: Pretty much. From what I gathered, the actual scene appears to continue to run in the background (or at least the registered editor update function). So before we were creating a new ocean in the prefab scene, and I think the new one wasn't being destroyed. I remember seeing a duplicate at one point.

And yeah, when we come back OnEnable or OnDisable on the original ocean are not called so that supports your statement.

moosichu:
I've looked in the profiler previously, just to confirm, it is the animated materials thing, but I don't know why that would prevent unity from updating the scene view whilst you are flying the camera around. It's definetley a unity bug but it's not a graphical hitch, something is just making it refuse to redraw the scene view :/

huwb:
> Annoyingly it goes away if OBS is open, so I can't record it.

Haha. I think you just found a workaround.

crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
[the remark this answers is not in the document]
huwb: Alright i think i resolved this - i poll for renderer/material/queue so this stuff stays live, and i tried different combinations of setting up ocean inputs and it seemed to handle all cases that i tested.

It adds a Update() function which i dont love adding, but it does very little work in standalone builds.

huwb:
Some more changes

* I found that dynamic waves were broken, fixed now.
* The depth cache also broke on code compiles, this should be fixed now too.
* Setting up ocean inputs should be graceful
* Setting up ocean from scratch should work
* Added outstanding issues above

Alright i think its ready now. What do you guys think?

huwb:
Yeah whirlpool hasnt been updated to make it run in edit mode. Definitely possible, and something we should probably look at. I'll leave that for future work for now.

Merging. Thanks both!

PR #486, opened 2020-04-27
Boat control abstraction

magenta404:
Moves boat control code out to its own component. This makes the boats more flexible to accepting different inputs like the new input system or nav mesh agents without having to modify the boat source.

The boats read from a Vector3 to get its desired input. It could be changed to a Vector2. But maybe a submersible could use the Y value?

There is an abstract class which components must inherit from. They need to write to the __Input_ property which is what the boats read.

I haven't added copyright headers or documentation yet. I'll wait for feedback on the proposal.


crest/Assets/Crest/Crest-Examples/BoatDev/Scripts/BoatControlPlayer.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    [Tooltip("The input axis name for steering."), SerializeField]
+    string _steerInputAxisName = "Horizontal";
+
+    [Tooltip("The input axis name for throttle."), SerializeField]
+    string _throttleInputAxisName = "Vertical";
+
+    void Update()
+    {
+        var throttle = UnityEngine.Input.GetAxis(_throttleInputAxisName);
```
moosichu, line 14: The problem with this approach (it's a minor one, as this is demo-code and not meant to be a production-ready control systrem) is that [FixedUpdate runs before Update](https://docs.unity3d.com/Manual/ExecutionOrder.html) meaning that this system has an effective frame of latency.  I *don't* think this needs solving, but it could be worth adding a comment explaining this?

crest/Assets/Crest/Crest-Examples/BoatDev/Scripts/BoatControlPlayer.cs
```diff
@@ -0,0 +1,19 @@
+﻿using UnityEngine;
+using Crest;
+
+public class BoatControlPlayer : BoatControl
```
moosichu, line 4: Worth making this a sealed class, minor thing but allows for de-virtualisation.

crest/Assets/Crest/Crest-Examples/BoatDev/Scripts/BoatControlBiased.cs
```diff
@@ -0,0 +1,13 @@
+﻿using UnityEngine;
+using Crest;
+
+public class BoatControlBiased : BoatControl
```
moosichu, line 4: Worth making this a sealed class, minor thing but allows for de-virtualisation.

crest/Assets/Crest/Crest-Examples/BoatDev/Scripts/BoatAlignNormal.cs
```diff
@@ -72,6 +68,11 @@ void Start()
             enabled = false;
             return;
         }
+
+        if (!_boatControl)
+        {
+            _boatControl = GetComponent<BoatControl>();
```
moosichu, line 74: Do we want this? Maybe it would be better to require the field to be assigned-to explicitly and then print an error and disable this script if this problem is encountered? (I personally like to keep things simple, and having one way of initialising as opposed to two reflects that, but please feel free to disagree on this point and don't let this be a merge-blocker).

moosichu (changes requested):
Really like this change overall! Only minor feedback. Thanks @daleeidd !

crest/Assets/Crest/Crest-Examples/BoatDev/Scripts/BoatControlPlayer.cs
daleeidd: Thanks. I've added a comment.

crest/Assets/Crest/Crest-Examples/BoatDev/Scripts/BoatAlignNormal.cs
daleeidd: Good point. I think allowing null here will be useful. For example, perhaps using boat probes for a broken vessel. A null component could be created for that use case though. So maybe we could warn on validation instead?

moosichu (approved):
Changes look good to me (once prefab cleanup and documentation is done)! I will let you decide what to do about a null `_boatControl` value. :)

crest/Assets/Crest/Crest-Examples/BoatDev/Scripts/BoatAlignNormal.cs
huwb: Or should we automatically attach a player control if no control is present?

huwb:
I think have a vector3 is a good shout.

"Bias" referred to adding a fixed value to the players input. I don't really remember the exact case, maybe that functionality is not needed anymore. But i guess what is now the 'bias' input could just be called BoatInputFixed or BoatInputConstant?

Should we be accomodating of extending to the new input system in the future? Should BoatControlPlayer be called BoatControlPlayerLegacy, or something similar to leave space? Removing/renaming files can cause issues with asset store updates so i guess we want to be solid-ish with file/class names.

crest/Assets/Crest/Crest-Examples/BoatDev/Scripts/BoatAlignNormal.cs
huwb: Maybe its a sensible default that can head off support requests.

daleeidd: We removed that functionality for the register input components (auto assign material). Also, the control classes live in _Crest-Examples_ which users might opt not to import. I created a proposal with PR #497 that might help address this problem better.

daleeidd:
> "Bias" referred to adding a fixed value to the players input. I don't really remember the exact case, maybe that functionality is not needed anymore. But i guess what is now the 'bias' input could just be called BoatInputFixed or BoatInputConstant?

Oh right. That makes sense. I have renamed it to BoatControlFixed.

> Should we be accomodating of extending to the new input system in the future? Should BoatControlPlayer be called BoatControlPlayerLegacy, or something similar to leave space?

That's a good idea. I thought we might be able to use the defines to accommodate both. It is possible, but the new system is different enough that we should create a seperate component for it. I've renamed it to BoatControlPlayerLegacy.

daleeidd:
Looks like we can just support the new input system in a similar way to how we are doing it already. I have reverted the rename. I'll just add support directly to this component so it will support both. I'll do it in a new PR since we need to also support it wherever we read from the old input system (DemoCamera, OceanDebugGUI etc) or it will error.

crest/Assets/Crest/Crest-Examples/BoatDev/Scripts/BoatAlignNormal.cs
huwb: True. thinking it through, i think it wasnt obvious to me that theres a good default choice of the ocean input. i guess it comes down to how we feel about the default behaviour being to apply player controls. i'll defer to your judgement

huwb:
Ok if it works directly into that component using defines (?) then that sound good

huwb (approved):

crest/Assets/Crest/Crest-Examples/BoatDev/Scripts/BoatAlignNormal.cs
daleeidd: I have changed this to use validation instead of getting a component.

daleeidd:
It has been a while but I resolved the last bit of feedback by using validation instead of getting a component when the boat control not provided by the user.

I added documentation. And I redid the prefabs since it was easier to bring master in. There is some noise in the prefab diff but it is minor. I cleaned most of the big stuff out.

I managed to get it working with Unity navigation. I will do a separate PR for that.

Havyx-Studio:
In my opinion unless the change is ready or it's a heavily requested addition - it seems like a low priority issue that could be re-examined in the future unless the lack of abstraction has a noticeable negative impact, or abstracting input would have a noticeable positive impact.

The exception being support for the input system package [ #808 ].

huwb:
It's been a while since I've looked at this but reviewing again my feeling is that we should auto assign a player control component if no control is present, otherwise it's going to trip people up and generate support traffic. Maybe an option is to auto assign the control if null and generate a warning when it does it, in addition to the validation warning you added. Would that work?

Havyx-Studio:
> It's been a while since I've looked at this but reviewing again my feeling is that we should auto assign a player control component if no control is present, otherwise it's going to trip people up and generate support traffic. Maybe an option is to auto assign the control if null and generate a warning when it does it, in addition to the validation warning you added. Would that work?

What happens in the case of a user who removes the player control component expecting their boat to have physics but is not controllable. 

Should an example script exist that can be assigned by the player that is basically blank/doesn't allow control of a boat, or are users who don't want a player controller component automatically assigned expected to remove the code that performs that check?

I can think of a few use-cases where you might want to delay adding a player control script - rather than adding it as soon as the game starts (something I do myself for my game where the world is generated, then the player is built through spawning different body parts, and only then are controls added).

huwb:
Fair point. Hm bit of a dilemma. Guess its fine as is then.

huwb:
Reset can be called at any time by the user so i'm not certain if that makes sense.

I reckon we should merge this as is. What is outstanding (besides the conflicts)?

daleeidd:
> Reset can be called at any time by the user so i'm not certain if that makes sense.

True.

> I reckon we should merge this as is. What is outstanding (besides the conflicts)?

Sounds good. Documentation. And I will add a fix function.

PR #487, opened 2020-04-27
Warn on reuse of a SampleHeightHelper multiple times in a frame

huwb:
This is a big pitfall that is too easy to get wrong - the intuitive
thing is to reuse SampleHeightHelper multiple times, but it will
simply overwrite the query position before the result is sent
to the GPU (if the compute height queries system is used).

This change throws a warning if the position is set multiple times
in a frame.

Unfortunately there is a case where this is valid - FixedUpdate can
be called multiple times per frame. I added an arg to indicate this
which will suppress the warning. Some collision providers will be
happy to receive multiple queries per frame. The compute height
query system will just overwrite the query pos and send off the
final (last) fixedupdate position to the GPU. This is imperfect but
I don't see how this can work better.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Collision/SamplingHelpers.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Collision/SamplingHelpers.cs
@@ -3,7 +3,7 @@
 namespace Crest
 {
     /// <summary>
-    /// Helper to obtain the ocean surface height at a single location. This is not particularly efficient to sample a single height,
+    /// Helper to obtain the ocean surface height at a single location per frame. This is not particularly efficient to sample a single height,
     /// but is a fairly common case.
     /// </summary>
     public class SampleHeightHelper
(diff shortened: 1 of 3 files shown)


crest/Assets/Crest/Crest/Scripts/Collision/SamplingHelpers.cs
```diff
@@ -15,16 +15,30 @@ public class SampleHeightHelper
...
         /// pass in the boats width. Larger objects will ignore small wavelengths.</param>
-        public void Init(Vector3 i_queryPos, float i_minLength)
+        /// <param name="fromFixedUpdate">Pass true if calling from FixedUpdate(). This will omit a warning when there on multipled-FixedUpdate frames.</param>
+        public void Init(Vector3 i_queryPos, float i_minLength = 0f, bool fromFixedUpdate = false)
         {
             _queryPos[0] = i_queryPos;
             _minLength = i_minLength;
+
+#if UNITY_EDITOR
+            if (_lastFrame >= Time.frameCount && !fromFixedUpdate)
```
daleeidd, line 36: This is minor but it might be more optimal to shortcut on the boolean check?

huwb: What do you mean - do you mean switch the order so the !fromFixedUpdate is evaluated first?

daleeidd: Sorry, yes.

huwb: Done

huwb:
Thanks both! @daleeidd i think everything is addressed now, shall i merge?

huwb:
Thanks

PR #489, opened 2020-04-28
Fix Material Preview error on Metal: not enough components in fragment shader

### LINKED ISSUE #488: [URP] Shader Inputs/Flow/Fixed Direction doesn't work on Metal
**Describe the bug**
Using the shader *Inputs/Flow/Fixed Direction* doesn't work on Metal. Throws this error:

> Metal: Error creating pipeline state (Crest/Inputs/Flow/Fixed Direction): fragment shader color output does not have enough component for the pixel format (MTLPixelFormatBGRA8Unorm_sRGB)

I don't understand why the flow target is a BGRA8 when it should be RGHalf?

**Screenshots / video**
*These are very helpful to understand and diagnose issues.*

**Versions**
Crest URP I think? 4.1?
Unity 2019.3.10f1

**To Reproduce**
Use fixed flow direction input on Metal.

**Platform**
 - Editor / OSX

**Hardware**
MacBook Pro 2014, osx 10.13.6

**Additional context**
*Add any other context about the problem here.*

daleeidd:
Fixes #488 

The error produced from this is because of the Material Preview in the Editor. This error does not stop the shaders from working correctly.

I've submitted a bug report to Unity. We could ignore this or revert it once fixed.

### DIFF
--- a/crest/Assets/Crest/Crest-Examples/Whirlpool/Shaders/Resources/FlowWhirlpool.shader
+++ b/crest/Assets/Crest/Crest-Examples/Whirlpool/Shaders/Resources/FlowWhirlpool.shader
@@ -39,7 +39,7 @@ Shader "Crest/Inputs/Flow/Whirlpool"
 				return o;
 			}
 
-			float2 Frag(Varyings input) : SV_Target
+			float4 Frag(Varyings input) : SV_Target
 			{
 				float2 flow = float2(0.0, 0.0);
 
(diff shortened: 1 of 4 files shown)


moosichu:
I hate how "vague?" type systems in shader languages feel. Don't know if it would be worth type-defing our own fragment shader return types (eg. frag_float4, frag_float, frag_half etc.) and then defining it differently based on the platform? What do you think @huwb

huwb:
I think I would go with the changes as is, even though its weird we have had to do this to shaders in the past so its the kind of weirdness I'm used to.

huwb:
Thanks @daleeidd

daleeidd:
From Unity:
> It's now fixed in Unity 2021.1.0a8 and above.

Issue #93, opened 2018-10-09
Triggering Particle Effects on Wave Peaks.


holdingjason:
I there a good way to apply a particle effect at the peak of a certain wave spectrum? My guess it would be similar to when foam is created,.  Do not need all of them just an occasional effect. 

(posted this in a new thread since I just realized where I have posted before was closed).

huwb:
Ah sorry i did actually get an email and i was planning to dig it up but i got buried under other stuff.

As a matter of fact there is something like this! The approach was suggested by Tessendorf - it samples random points near the viewer and detects choppy waves and emits particles, which is basically what you suggested.

Dig out the branch fx_test https://github.com/huwb/crest-oceanrender/tree/fx_test . It's from March and is utterly unrecognizable as Crest which is amusing in itself, but if you run the scene and wait a few moments you should start to see PFX being triggered in choppy parts of water (they look really bad and can take a while to spot).

I think this could be pretty awesome with some FX love, I found it really hard to author fx because im not a vfx artist at all. I'd be very interested if you wanted to mess around with this and see if you can get good results out of it..

holdingjason:
Sounds great will dig that up.  Yep we have an artist on staff that can put some things together and make something look nice that you can use.  Right now we are ultra focused on ship movement, fx etc to make a fun visual experience for our players so things like this, spray coming up from the boat as it impacts the water etc.

huwb:
Ok! Even just seeing your result/sharing techniques for making it look good would be helpful, but let's see how we go.

This stuff will probably need a bit of love to bring it forward - it needs to be adjusted to use the collision providers which didn't exist back then. It should not be a huge amount of work. I can assist or make a pr for it sometime soon, maybe on weekend.

holdingjason:
We are in no big rush so whenever you get to it is fine.

holdingjason:
@huwb Hola.  Just starting to finally get back and look at this.  

Ya that old branch is a mess and does not even really run anymore.  Will see if I can make any sense of how it computed the chop. 

[...]

PR #494, opened 2020-05-02
Ocean input option to correct for displacement

huwb:
**Status**: Ready for review. Might take some work to merge into SRPs so probably worth releasing 4.3 before merging this.

Adds 'displacement correction' option on ocean inputs. Stops input 'drifting' around with the horizontal displacement of the ocean. This is in response to someone posting the question on discord as to why their foam input drifted away from the character.

This was previously addressed using code like this https://github.com/crest-ocean/crest/blob/master/crest/Assets/Crest/Crest/Scripts/Interaction/ObjectWaterInteraction.cs#L116 . That requires nontrivial, bespoke C# code and requires a parent transform. This PR introduces a systemic fix that is easy to use and simpler.

TODO

- [x] Correction only works for wave particle input. Extend to rest of inputs.. DONE
- [x] Move object-water interaction to use this? YESSS this is a big win
- [x] We're currently failing to flush GUIDs from the water height query system and eventually hit the max gui count. Adding this additional sampler will likely exacerbate the issue. https://github.com/crest-ocean/crest/issues/93#issuecomment-616573908 RESOLVED


huwb:
@daleeidd @moosichu do you have thoughts on this

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/ClipSurfaceConvexHull.shader
```diff
@@ -44,8 +45,12 @@ Shader "Crest/Inputs/Clip Surface/Convex Hull"
 			Varyings Vert(Attributes input)
 			{
 				Varyings o;
-				o.positionCS = UnityObjectToClipPos(input.positionOS);
-				o.positionWS = mul(unity_ObjectToWorld, float4(input.positionOS, 1.0));
+
+				o.positionWS = mul(unity_ObjectToWorld, float4(input.positionOS, 1.0)).xyz;
+				// Correct for displacement
+				o.positionWS.xz -= _DisplacementAtInputPosition.xz;
```
daleeidd, line 51: This doesn't work as expected for clip surface since it already has displacement correction in the ocean renderer.

crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
```diff
@@ -52,6 +52,10 @@ public abstract partial class RegisterLodDataInputBase : MonoBehaviour, ILodData
         public abstract bool Enabled { get; }
 
         public static int sp_Weight = Shader.PropertyToID("_Weight");
+        public static int sp_DisplacementAtInputPosition = Shader.PropertyToID("_DisplacementAtInputPosition");
+
+        [SerializeField, Tooltip("Correct for horizontal displacement so that this input does not move from side to side with the waves.")]
+        bool _applyDisplacementCorrection = false;
```
daleeidd, line 58: Curious if it would be more intuitive if this option is the opposite and by default inputs don't receive displacement. So this option could be named "Apply Displacement" or "Move With Ocean Surface" or something. I think a lot of the time inputs are added to objects like boats or static objects where this _applyDisplacementCorrection will be wanted. Thoughts?

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/ClipSurfaceConvexHull.shader
huwb: Oh good catch. It already samples the clip data at the displaced position. So you mean we could make it sample at the undisplaced position to make it consistent with how the other ocean data is sampled? I could try that.

crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
huwb: Hmmm yeah I guess so! Could call it "follow horizontal motion"?

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/ClipSurfaceConvexHull.shader
daleeidd: Yes. I think it makes most sense to go that route. Easier to keep track of.

crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
daleeidd: That sounds good

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/ClipSurfaceConvexHull.shader
huwb: This has not really worked out. The correction this branch applies is only to the center of the input, not over the whole area. This means the clipping is inaccurate if we use this correction.

Think i will leave the ocean shader how it is, and remove this correction option for clipping inputs.

huwb:
I changed things up a bit to rename the option as suggested, and hardcode it for some input types:

* Sea floor depth - disable the correction, its not needed due to how the data is sampled
* Clip data - disable correction, its not needed due to how the data is sampled
* Dyn waves input - enable the correction because i cant think of a case where one would want it disabled

We could expose the option on the dyn waves if it is ever requested.

Thoughts @daleeidd?

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/ClipSurfaceConvexHull.shader
daleeidd: No worries. Sounds good.

daleeidd:
Sounds like a good idea.

Everything looks good. The prefabs can probably be reverted too if it isn't too much trouble as I think the defaults handle everything now.

huwb:
Reverted everything. Merging - let me know if either of you spot other issues. Thanks!

PR #496, opened 2020-05-02
Flush query guids every frame

huwb:
**Status:** Ready to merge

In the compute height query system, queries are made with an identifying GUID. This registers the query and the GUID is used to track it across frames.

Right now we are never flushing GUIDs which is bad. The code is manually populating the registrar with last frame's guids every time its acquired. The number of query GUIDs grows montonically, eventually throwing the "max guids" error.

I believe queries are uploaded and processed if they are present, even if they're no longer being actually used.

This code avoids repopulating newly acquired query registrars with previous frames' GUIDs. This seems to work and it seems like a better solution.

I also added a line to the gui overlay to report how many guids there are in each frame's results.


huwb:
Thanks merging

PR #499, opened 2020-05-08
Manually cull ocean mask chunks

### LINKED ISSUE #479: [HDRP] Underwater draws through buildings
**Describe the bug**

Discord user Puppetmaster reported an issue where the underwater effect appears through a building:

ezgif com-video-to-gif

He suggests it may have been introduced in Crest HDRP 4.1.

It might be that the surface is being culled by unity using dynamic occlusion culling. These tiles wont be added to the visible list, and therefore won't be registered to draw into the ocean mask. However I could not reproduce this effect locally.

Actually, maybe this is due to multiple water levels. I've enquired.

**Versions**
Unity 2019.3.9
Crest HDRP 4.1

**To Reproduce**
How can the developers reproduce this bug? Are you able to reproduce the issue in the example content/scenes?
Please include steps to reproduce the behaviour, with the example content ideally so we can follow the steps on our side.

**Platform**
 - Editor or standalone?
 - Platform (win/mac/...)

**Hardware**
What hardware are you seeing the issue on? GPU model is particularly useful.

**Additional context**

moosichu:
**Status:** Ready for review

We also render them before AlphaForwardRender pass to that mask data can
be available to transparent objects. This will be useful for things such
as underwater windows.

Should fix #479 (when ported to HDRP) and provide useful features for #462.

The mask texture is available from any transparent shader as `_CrestOceanMaskTexture`.


huwb:
I scanned through the changes a couple of times on my phone and it all looked good to me, nice one

Should we disable tiles which fail the culling test? This would save cost as unity will otherwise run culling on them again later. I did this kind of thing for the underwater optimisation where I turned off tiles which are too far from the camera and it worked well. Perhaps later that test I did could be combined with this code.

moosichu:
I think it might be worth creating a seperate PR for that, will think about it overnight.

Main reason is that you can have more than 1 camera (eg. in VR), so we would have to be careful about handling that case. Maybe it could be added a an option?

moosichu:
How are the tiles being disabled? I'm more than happy to debug and look into it. Will merge this for now and we can consider manually disabling culled tiles at a later point.

daleeidd:
> How are the tiles being disabled?

Manually through the editor. The first image is without disabling any tiles but I guess it could be one of the many other line issues. I'll check it again once everything is finished.

moosichu:
image

I've been able to reproduce the issue, it's happening because "OnWillRenderObject" is no longer being called, will whip-up a fix.

PR #500, opened 2020-05-10
Scale caustics strength by primary light and depth fog

daleeidd:
Scales the caustics strength by primary light colour, depth fog density and scene depth.

Solves the problem of caustics showing up under underwater lights when they shouldn't. This solution works independently from the _UnderwaterEnvironmentalLighting_ component, but it is affected by it since it scales the primary light. But that is a separate problem.

I am not certain how they should co-exist with depth of field and focal depth, but right now they don't appear to conflict.

I plan to port this to HDRP to replace limiting caustics by scene depth, but it might be a bit more involved due to different lighting units.


huwb:
Looks good to me, nice one. I think this co-exists fine with DoF etc.

One tiny nitpick - could you make the constants like `1.0`, not `0` or `1.` . I recently found this in the shader coding standard in SRP which i think gives a nice consistency/readability.

daleeidd:
Thanks. Fixed and merged.

PR #501, opened 2020-05-11
Experiment - emit particles when emitter submerged

huwb:
As requested on discord - emit particles when emitter submerged.

I think this is orthogonal/separate to our other experiments emitting particles from wave crests.

Examples added to main.unity. Prefab changes not applied yet.

Not sure if this is mergeable, what do you guys think. One issue is the effect looks programmer-y. Anyone keen to try to art them up?

image


knxrb:
I've had a go at updating the particle effects to look more like water splashing up and falling back down. It has a new material and texture, as well as changes to the particle system prefab itself. I created everything myself so if it's liked then you can use everything.

WaterParticlesEffect

I'm not able to commit to this repo directly so I've made a fork and updated my branch there:
https://github.com/knxrb/crest/commit/116c89ac1e6406b379f3ba3927285bc9491a930c

huwb:
Thanks @knxrb :). It looks a lot better. i'll add a task for us to look at adding this into crest.

huwb:
I integrated it and experimented with different settings for how it emits. I think its working fairly well now.

It can also be attached to moving objects:

ezgif com-video-to-gif (1)

Maybe like this can help with spray from boats and etc, although right now it only detects vertical velocity.

moosichu:
@huwb maybe throw this sphere into one of the scenes with dynamic waves for the fully effect? :)

huwb:
Good point - with dynamic waves

ezgif com-video-to-gif (3)

I think i need to look at faking some lighting on the particles

huwb:
@daleeidd not urgent but at some point it would be good to get your thoughts on where we should take this.

i like what you did with the events on the other merged pr.

should i aim to convert this branch to vfx graph? it seems like that way we should be able to feed velocity through as a custom param into the graph, and that way i should be able to reuse your event. is this what you had in mind as well?

daleeidd:
I made a branch that adds the necessary events to get this working: [feature/emit-particles-when-submerged-with-events](https://github.com/crest-ocean/crest/tree/feature/emit-particles-when-submerged-with-events)

Unfortunately, the particle system doesn't expose all properties so I had to fetch the particle system from the event listener to modify the start speed multiplier. Also, I had to disable the listener to implement the _allow multiple emissions_ property. It works but not sure about performance implications yet. It could be mitigated by setting this up in the OnEnable call if needed.

VFX Graph doesn't expose properties either. But we can just interface with our own component (Spray Effect it is called in the other FX branch).

There are also performance concerns. [Unity events don't perform as wells as C# events](https://jacksondunstan.com/articles/3335) from what I read.

I think UnityEvents is a great for experimenting or for users that don't know scripting (adding more events to that component from the other PR still sounds like a good idea for this reason). But I am not really sure if I would see it as an effective substitute for a tailored solution like this.

huwb:
Thanks, thats helpful

VFX graph allows you to add properties to the blackboard. Are you saying these arent exposed to events? I can try it as well, just confirming in case you know off the top of your head.

daleeidd:
They aren't. I already checked :(

We could create a helper component which bridges UnityEvent to the VFX blackboard. In its simplest form it would require users to enter the name of the property. Or we could create an enum so there is a drop down.

huwb:
digging up an old one.. working out how to pull this together and move it forward.

1. i think we should have a component that allows to easily send events when submerged. so that users could e.g. hook up a SFX to the camera going underwater by just slapping a component on the camera. i assume this means using a UnityEvent. 
2. ideally we could send through info like velocity etc. can this be tacked onto the UnityEvent in a generic way? i dont see much api to set generic data. or is it a case of checking the target type and poking it through as you have done in https://github.com/wave-harmonic/crest/commit/bdfe2c606cae395b0442052a11ce1798d83d2fc5 ? 
3. i saw the UnityEvent perf issues - slower invoke and garbage on first invocation? - but i think nevertheless they are a good solution for easy setup for non-advanced users. advanced users can always write their own stuff, and we could potentially support that further on our side once we have the basic approach merged.

PR #502, opened 2020-05-12
Add dynamic horizon line safety margin prevent artifacting

### LINKED ISSUE #456: [HDRP] Horizon line issue

moosichu:
Shimmies the horizon-line up and down based on the camera position. It's a hack but it's pretty-robust. This commit also takes into account the world space height of the camera frustum for determining wether or not to force the post-process shader to run full-screen, fixing an issue where this would kick-in prematurely if the ocean didn't have any waves which I was using to test the horizon-line fudge.

Will fix #456 when it's merged into HDRP.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterPostProcess.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterPostProcess.cs
@@ -26,8 +26,9 @@ namespace Crest
         [SerializeField, Tooltip("Assign this to a material that uses shader `Crest/Underwater/Post Process`, with the same features enabled as the ocean surface material(s).")]
         Material _underwaterPostProcessMaterial;
 
-        [Header("Debug Options"), SerializeField]
-        bool _viewOceanMask = false;
+        [Header("Debug Options")]
+        [SerializeField] bool _viewOceanMask = false;
+        [SerializeField] bool _disableOceanMask = false;
         // end public debug options
 
         private Camera _mainCamera;
@@ -162,8 +163,10 @@ namespace Crest
             PopulateOceanMask(
                 _maskCommandBuffer, _mainCamera, OceanBuilder.OceanChunkRenderers, _cameraFrustumPlanes,
                 _textureMask, _depthBuffer,
-                _oceanMaskMaterial
+                _oceanMaskMaterial,
+                _disableOceanMask
             );
+
         }
 
         void OnRenderImage(RenderTexture source, RenderTexture target)
--- a/crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterPostProcess.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterPostProcess.cs
@@ -27,8 +27,10 @@ namespace Crest
         Material _underwaterPostProcessMaterial;
 
         [Header("Debug Options")]
-        [SerializeField] bool _viewOceanMask = false;
+        [SerializeField] bool _viewPostProcessMask = false;
         [SerializeField] bool _disableOceanMask = false;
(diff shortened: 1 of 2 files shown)


huwb:
Thanks Tom.

Just to understand, is it possible that the lined is moved too far in this change? Or is that a non issue?

It would be good to understand the core issue here. What is currently happening for, say, the camera underwater case. Is it estimating the horizon as being too low, leaving a gap?

Is there a reliable repro case for this? If so could you temporarily add data to this branch to set up the repro case so I could take a look? I don't think I ever had a good repro so my horizon line change was purely speculative and might simply have a bug or a bad assumption somewhere.

moosichu:
For me I just have to fly-around the main scene for a couple seconds (previously) and I would always get it very quickly. I can record an OBS video if that helps. 

I tested-this solution by actually making the fudge-factor an editable-variable at runtime and modifying it until the horizon-line issue disappeared.

I have actually done some very crude-mathematics as to why this should be fine. Basically, as long as the linear-rate at which the proportion of the ocean that covers the screen vertically grows at a rate greater than the distance the horizon-line moves, then ocean mask is guarenteed to be covering it. 

The thing that would break this then would be if there was an ocean input that pushed the *entire* ocean down except for the edges, but I don't think that's something anyone would do. 

If it's worth mentioning, I've never been able to repro the horizon-line issue in HDRP, only in BIRP. But if you use the `feature/post-proc-underwater-2` branch on a Linux machine you will get the problem very quickly. (it's less common on Windows and never happens on my Mac).

huwb:
I decided to do the root cause analysis. i did the test where i compared the horizon line with the 'real' horizon line (by creating a big plane) and found something pretty funny - the ocean is simply not covering the full extents of the frustum :D . If you set the ocean min scale to 8 instead of 4, it will cover the frustum and the line goes away!

Alternative fix is to reduce the far clip distance. Looking at the camera settings, the far plane value is **huge**: 100000

I will look at reducing the far plane value to 10000, and hopefully we don't have to fudge anything (yet)?

@moosichu i think you mentioned that there is another changed mixed into this branch. How to proceed - maybe a separate PR?

huwb:
Created #506 for the far plane fix

There's probably a brainstorm to have about if this could be caught with validation I think it would be possible to compute the far extents of the ocean mesh analytically and assert if it doesn't cover all frustum orientations. Probably going to leave this for future work..

moosichu:
I still think this PR should be merged, but I couldn't make it an option that we can ask people to enable if the horizon-line problem still crops of for them. Commented on it in #506.

huwb:
Yeah that's fair. Is that done or not yet? Could you add a status line to the pr? Thanks

Oh man I just realized - the whole horizon line computation in c# - was this unnecessary?? I'll test with the far plane fix..

huwb:
Just read your note on my other pr that this doesn't fix all cases. Will go back to analysing.

huwb:
Alright i give up - i got a repro after about 20min of trying, but then i accidentally change my viewport size and lost it again :). it must be when the horizon is very close to the pixel centers and precision error is pushing it over. doubles might help but its still theoretically closer, so feels like adding a small safety buffer is required.

I think if this line of thinking is correct, an offset would always be required and i'd vote for rebranding it from "hack" to "safety margin" and turning it on by default and getting it merged. (the rebrand might sound like an odd thing to do but might be good from a product/code quality point of view).

moosichu:
image
 Added a slider as a debug option and included a tooltip, hopefully the fact that it's a debug option means most people won't touch it.

PR #506, opened 2020-05-16
Reduce far plane distance to 10000

huwb:
**Status:** Ready to merge

It was set to 100k which is huge, much bigger than previous projects i've worked on.

This large distance meant the ocean (with default settings) was not covering the frustum, messing up the ocean mask.

I think this is a good change and safe. Tested demo scenes and they look fine to me in editor.

### DIFF
--- a/crest/Assets/Crest/Crest-Examples/Shared/Prefabs/Internal/DemoCameraCore.prefab
+++ b/crest/Assets/Crest/Crest-Examples/Shared/Prefabs/Internal/DemoCameraCore.prefab
@@ -57,7 +57,7 @@ Camera:
     width: 1
     height: 1
   near clip plane: 1
-  far clip plane: 100000
+  far clip plane: 10000
   field of view: 60
   orthographic: 0
   orthographic size: 5


daleeidd:
Thinking about this, there should be a way to include the far clip plane as part of validation. But that can be for another PR. I'll add a task for it.

moosichu:
image

Whilst this *does* improve things *a lot* (and worth merging-in), the reason I'm still partial towards introducing the horizon-line fudge (even though it seems hacky) is because it shouldn't really break-down under any circumstance. (Unless the entire ocean is sloped all-the way until the horizon).

I definetely think this is worth merging-in, but it still presents this issue. I hadn't realised that forgot to mention that I tested this and it still had the problem.

moosichu:
It's worth pointing out that to even be able to get a screenshot of the issue, I ended up having to record myself flying around with the camera using OBS as otherwise it would be too difficult to find the repro case. The horizon line fudge is the only solution I've found that I've not been able to find any problems-with whatsoever.

moosichu:
I'ts worth pointing out that this is already the far clipping plane distance in HDRP (where the issue did happen a lot less frequently) which is where Crest users have been reporting the problem.

huwb:
Ah okkkk that's disappointing. I'll try harder to repro it and do another analysis pass

huwb:
I'm going to merge this in any case, let's move conversation back to 502

PR #507, opened 2020-05-16
Cleanup ocean chunk renderers list, thanks @daleeidd

moosichu:
Make the list of ocean chunk renders readonly.


moosichu:
Will merge this as it is a pretty minor change anyway, if you do have any feedback I can apply-it post-facto.

PR #509, opened 2020-05-16
Initiliase chunks for ocean mask even if it's disabled in the scene hiearchy

moosichu:
This ensures that for when we disable chunk rendering in the scene
hiearchy for things like local water bodies or other optimisations, if
they will still be correclty renderered as part of the ocean mask which
assumes that chunks will always form a continous surface for the post
process effect to work accurately.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/OceanChunkRenderer.cs
+++ b/crest/Assets/Crest/Crest/Scripts/OceanChunkRenderer.cs
@@ -71,8 +71,16 @@ namespace Crest
             _currentCamera = camera;
         }
 
+        // Used by the ocean mask system if we need to render the ocean mask in situations
+        // where the ocean itself doesn't need to be rendered or has otherwise been disabled
+        internal void BindOceanData(Camera camera)
+        {
+            _currentCamera = camera;
+            OnWillRenderObject();
+        }
+
         // Called when visible to a camera
-        internal void OnWillRenderObject()
+        void OnWillRenderObject()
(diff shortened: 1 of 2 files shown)


moosichu:
image

Previously with this setup, you would have gaps in the mask resulting in areas with no post-process effect being applied. This circumvents that issue:

image

huwb:
Thanks dude. Was naively hoping this wasnt required after i seemed to get good results just stupidly disabling ocean tiles. i'm guessing this was thanks to the `FULL_SCREEN_EFFECT` keyword and i didnt test enough at shallow depths. d'oh.

The only note i have is that OnWillRenderObject silently depends on outside state - Camera. current being set, or OceanChunkRenderer.BeginCameraRendering() having been called first to set _currentCamera (would not be the case if hte renderer/tile was disabled). In HDRP this camera variable is barely used, but i checked and it is used in a meaningful way in URP. could we perhaps leave OnWillRenderObject private, and move its contents into a new general function which receives the camera as input, for example:

`internal void BindOceanData(Camera camera)`

does that make sense?

moosichu:
Works for me, will give that a go.

moosichu:
Done :)

huwb:
Sure thing! Happy to make it, I feel bad for misleading you :). But I'm also heading out, so let's just sync later. Will make hdrp release tomorrow I think so no critical rush. Cheers!

huwb:
Merged as suggested!

PR #510, opened 2020-05-16
Cull ocean tiles if you aren't going to be able to see them through underwater fog anyway

moosichu:
From @huwb:

> I profiled an underwater view on my laptop at 1080p and it came to around 5ms underwater surface + 0.7ms underwater effect, and this is without the crest simulation (~1ms), so we're sitting a bit high on hdrp.
> 
> Since light does not travel far in the volume, i realised there was no point drawing the surface too far away from the viewer.
> 
> This change simply knocks out tiles that are too far away. This change saved around 2ms for this kind of view where surface is 'almost' invisible:
> 
> When the camera is deeper the benefit is even greater - the 5ms goes away completely, and no tiles are drawn into the mask. The 0.7ms to render the effect still remains. Two potential options would be to have a special variant of the shader when viewer is definitely underwater which might save some work (?), or do the whole effect at half res (would need careful treatment of edges).

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
+++ b/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
@@ -68,6 +68,9 @@ namespace Crest
         [SerializeField, Tooltip("Number of ocean tile scales/LODs to generate."), Range(2, LodDataMgr.MAX_LOD_COUNT)]
         int _lodCount = 7;
 
+        [Tooltip("Proportion of visibility below which ocean will be culled underwater. The larger the number, the closer to the camera the ocean tiles will be culled."), SerializeField, Range(0.000001f, 0.01f)]
+        public float _underwaterCullLimit = 0.001f;
+
 
         [Header("Simulation Params")]
 
@@ -100,9 +103,6 @@ namespace Crest
         [Tooltip("Clip surface information for clipping the ocean surface."), SerializeField]
         bool _createClipSurfaceData = false;
         public bool CreateClipSurfaceData { get { return _createClipSurfaceData; } }
-
-        [Tooltip("Proportion of visibility below which ocean will be culled underwater."), SerializeField, Range(0.000001f, 0.01f)]
-        public float _underwaterCullLimit = 0.001f;
 
         [Header("Debug Params")]
 
@@ -373,10 +373,11 @@ namespace Crest
             var definitelyUnderwater = ViewerHeightAboveWater < -5f;
 
             // TODO - dont call this every frame!! save off list from oceanbuilder?
-            var tiles = GetComponentsInChildren<Renderer>();
+            var tiles = OceanBuilder.OceanChunkRenderers;
             foreach (var tile in tiles)
             {
-                tile.GetComponent<Renderer>().enabled = (!definitelyUnderwater) || (_viewpoint.position - tile.bounds.ClosestPoint(_viewpoint.position)).magnitude < VolumeExtinctionLength;
+                Renderer tileRenderer = tile.Renderer;
+                tileRenderer.enabled = (!definitelyUnderwater) || (_viewpoint.position - tileRenderer.bounds.ClosestPoint(_viewpoint.position)).magnitude < VolumeExtinctionLength;
             }
         }
 
--- a/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
+++ b/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
@@ -165,7 +165,6 @@ namespace Crest
         readonly int sp_texelsPerWave = Shader.PropertyToID("_TexelsPerWave");
         readonly int sp_oceanCenterPosWorld = Shader.PropertyToID("_OceanCenterPosWorld");
         readonly int sp_meshScaleLerp = Shader.PropertyToID("_MeshScaleLerp");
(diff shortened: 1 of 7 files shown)


moosichu:
Still in progress, but either-way shouldn't be merged until after #509 as it is based on it.

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
```diff
@@ -355,6 +366,18 @@ void LateUpdateLods()
...
 
+        void LateUpdateTiles()
+        {
+            var definitelyUnderwater = ViewerHeightAboveWater < -5f;
+
+            // TODO - dont call this every frame!! save off list from oceanbuilder?
+            var tiles = GetComponentsInChildren<Renderer>();
+            foreach (var tile in tiles)
+            {
+                tile.GetComponent<Renderer>().enabled = (!definitelyUnderwater) || (_viewpoint.position - tile.bounds.ClosestPoint(_viewpoint.position)).magnitude < VolumeExtinctionLength;
```
moosichu, line 377: @huwb if you are already getting a list of renderers, is there a reason why GetComponent is being called here as well?

moosichu:
image

I think some tweaking needs to be done, as you can still see where the ocean is being clipped.

moosichu:
So one thing worth bearing in mind with this change is that it potentially conflicts with #462 especially if you have a submarine or somesuch with windows - the fog won't be rendered when you are inside the volume so you will not want to cull the ocean shader in that case. I think it makes sense for this to still be merged, but then we should address that problem in that branch.

moosichu:
One thing that needs to be done here as well potentially is disabling the ocean mask explicitly under circumstances where this would be happy anyway. However, I want to do that as a seperate PR *after* #462 has been merged-in as it would interact with that in complex ways, and rendering the mask is cheaper than rendering the ocean shader, so this should save on a large chunk of the cost.

huwb:
Good point about submarines!

Yeah good point, can mask rendering be avoided if FULL_SCREEN_EFFECT keyword is active?

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
huwb: Shitty prototyping code reasons :). good catch!

moosichu: Sweet, using the list we already generate for the manual culling :) so sorted.

moosichu:
Yeah, that is what I want to do, but the object mask is used to cull the post-process effect in the submarine case, so it still needs to be sampled there. I'm going to cleanup that particular branch with all the culling sorted on the mask in one fell swoop.

moosichu:
Yeah, that is what I want to do, but the object mask is used to cull the post-process effect in the submarine case, so it still needs to be sampled there. I'm going to cleanup that particular branch with all the culling sorted on the mask in one fell swoop.

moosichu:
Ready to merge after upstream issues have been sorted.

moosichu:
Something that I think *will* work under all circumstances is:

if FULL_SCREEN_EFFECT is enabled *and* ocean tiles are beyond the kill distance, then we don't render into the mask. (The mask is also used to do things like avoid-caustics from being drawn on the surface, which is why you might still want it even if FULL_SCREEN_EFFECT is enabled).

huwb:
I think there were a few prototypey things i dont think we want, forgive me if i cleaned out too much.

I'll merge as it seems like we can fix later.

crest/Assets/Crest/Crest/Shaders/Ocean.shader
```diff
@@ -419,13 +419,8 @@ Shader "Crest/Ocean"
 
 			half4 Frag(const Varyings input, const float facing : VFACE) : SV_Target
 			{
-				float pixelZ = LinearEyeDepth(input.positionCS.z);
 				#if _UNDERWATER_ON
 				const bool underwater = IsUnderwater(facing, _ForceUnderwater);
-				if(underwater)
-				{
-					clip(_CrestVolumeExtinctionLength - pixelZ);
```
moosichu: Although it's an extra branch, won't the fact that this could be happening for a fair few fragments in parallel mean that it could be a potential saving?

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
```diff
@@ -372,12 +364,15 @@ void LateUpdateTiles()
...
+            var density = _material.GetVector("_DepthFogDensity");
+            var minimumFogDensity = Mathf.Min(Mathf.Min(density.x, density.y), density.z);
+            var volumeExtinctionLength = -Mathf.Log(_underwaterCullLimit) / minimumFogDensity;
+
             var tiles = OceanBuilder.OceanChunkRenderers;
             foreach (var tile in tiles)
             {
                 Renderer tileRenderer = tile.Renderer;
-                tileRenderer.enabled = (!definitelyUnderwater) || (_viewpoint.position - tileRenderer.bounds.ClosestPoint(_viewpoint.position)).magnitude < VolumeExtinctionLength;
+                tileRenderer.enabled = (!definitelyUnderwater) || (_viewpoint.position - tileRenderer.bounds.ClosestPoint(_viewpoint.position)).magnitude < volumeExtinctionLength;
```
moosichu, line 375: This always seems to be true for some reason in my most recent tests... debugging now.

PR #514, opened 2020-05-23
Fix terrain with more than four layers being black in planar reflections

daleeidd:
Reported by jRocket on Discord. If more than four layers is used on terrain, these layers will appear black in planar reflections.

If I disable HDR on the planar reflection script, it works as expected. So it appears that Unity is packing the alpha channel to achieve layering.

Saturating the alpha fixed this. I don't think this is an issue for anything else. Not sure if there are any negative implications of taking this route.


huwb:
I reckon this is a bug with the terrain system.

But I'm happy with your change. Don't think there's any other impact. Thanks!

PR #515, opened 2020-05-23
[In Progress] Signed Distance Shoreline

moosichu:
Creating a PR to track progress. Not yet ready for review/merge or anything.

Current Status:

- I've added the option to generate a signed-distance field inside the depth-cache: R is depth, G is distance to nearest shoreline, B & A are normalised vector to nearest shoreline.

- This works fine with *either* the island example *or* a slanted land-mass example I added to the main scene (see images)

- In terms of rendering the waves the thing which ended-up giving the *best* results (see `AnimWavesGerstnerBatch.shader` and `GerstnerShared.hlsl`) was to explicitly define a wave for shorelines that is lerped into existance as the distance decreases (it could potentially use depth instead? - it would require a larger cache size though to define the SDF in more places in space). This works fairly well - I tried a lot of different approaches so happy to go in-more depth into them at some point if that would be helpful. This included using the directional-wave code we already had and also trying to generate shoreline-waves from the existing ocean spectrum in-other ways - all seemed to produce poorer results than this. 

Island example:
image

Slanted land mass example:
image

## Issues

- [ ] If *both* the land mass and the island are enabled - the positions of the final ocean-output is messed-up. I have no idea what the root cause is - if it's in the SDF generation, depth cache generation, Jump Flood algorithm, shader bugs, CS bugs or even the gerstner wave sampling. I haven't looked too-much into-it.

- [x] The ocean can clip the land-mass in an unpleasant way (see image below) - should we set a floor on the surface height or do some kind of lerping here?

- [ ] Especially notice-able on shallow-angled shores, the ocean just comes to a stop instead of exhibiting the horizontal motion you would expect (see [here](https://il5.picdn.net/shutterstock/videos/1199614/thumb/1.jpg)). I don't know how to best implement this.

- [ ] We need to add noise to break-up the waves and their phases. 

- [ ] We need to add particles to the breaking waves.

- [ ] What kind of configurable parameters can we add?

Ocean Clipping example:

image


huwb:
Looking forward to playing around with this!

moosichu:
Current Status:

- I've added the option to generate a signed-distance field inside the depth-cache: R is depth, G is distance to nearest shoreline, B & A are normalised vector to nearest shoreline.

- This works fine with *either* the island example *or* a slanted land-mass example I added to the main scene (see images)

- In terms of rendering the waves the thing which ended-up giving the *best* results (see `AnimWavesGerstnerBatch.shader` and `GerstnerShared.hlsl`) was to explicitly define a wave for shorelines that is lerped into existance as the distance decreases (it could potentially use depth instead? - it would require a larger cache size though to define the SDF in more places in space). This works fairly well - I tried a lot of different approaches so happy to go in-more depth into them at some point if that would be helpful. This included using the directional-wave code we already had and also trying to generate shoreline-waves from the existing ocean spectrum in-other ways - all seemed to produce poorer results than this. 


Island example:
image

Slanted land mass example:
image


## Issues

- [ ] If *both* the land mass and the island are enabled - the positions of the final ocean-output is messed-up. I have no idea what the root cause is - if it's in the SDF generation, depth cache generation, Jump Flood algorithm, shader bugs, CS bugs or even the gerstner wave sampling. I haven't looked too-much into-it.

- [ ] The ocean can clip the land-mass in an unpleasant way (see image below) - should we set a floor on the surface height or do some kind of lerping here?

- [ ] Especially notice-able on shallow-angled shores, the ocean just comes to a stop instead of exhibiting the horizontal motion you would expect (see [here](https://il5.picdn.net/shutterstock/videos/1199614/thumb/1.jpg)). I don't know how to best implement this.

- [ ] We need to add noise to break-up the waves and their phases. 

- [ ] We need to add particles to the breaking waves.

- [ ] What kind of configurable parameters can we add?


Ocean Clipping example:

image

crest/Assets/Crest/Crest/Shaders/OceanInputs/GerstnerShared.hlsl
```diff
@@ -91,3 +88,24 @@ half4 ComputeGerstner(float2 worldPosXZ, float3 uv_slice)
...
+	half3 result = (half3)0.0;
+
+	if(depth > 0.0)
+	{
+		const float pi = 3.141;
+		const float twoPiOverWavelength = (2.0 * pi) / 2.5;
+		const float twoPiOverPeriod = (2.0 * pi) / 2.0;
+		const float amplitude = 0.12;
+		// Chop increases as depth increases
+		const float chopAmplitude = 2.0/(1.0+sqrt(distanceToShore));
```
moosichu, line 104: You will notice sqrt(distanceToShore) is used as a paramater to increase the chop as the distance to the shore increases - another option here could be to use depth - but this worked well in the experimentation.

crest/Assets/Crest/Crest/Shaders/OceanInputs/GerstnerShared.hlsl
```diff
@@ -91,3 +88,24 @@ half4 ComputeGerstner(float2 worldPosXZ, float3 uv_slice)
...
+
+		const float angle = (twoPiOverWavelength * sqrt(distanceToShore)) + (_CrestTime * twoPiOverPeriod);
+		result.y = amplitude * cos(angle);
+		result.xz = chopAmplitude * directionToShore * sin(angle) * result.y;
```
moosichu, line 108: Multipliying the chop by result.y means that the waves tip-forward near the top - and it also helps prevent self-intersection (I don't know why! but when I added the term they went away).

crest/Assets/Crest/Crest/Shaders/OceanInputs/GerstnerShared.hlsl
```diff
@@ -91,3 +88,24 @@ half4 ComputeGerstner(float2 worldPosXZ, float3 uv_slice)
...
+	{
+		const float pi = 3.141;
+		const float twoPiOverWavelength = (2.0 * pi) / 2.5;
+		const float twoPiOverPeriod = (2.0 * pi) / 2.0;
+		const float amplitude = 0.12;
+		// Chop increases as depth increases
+		const float chopAmplitude = 2.0/(1.0+sqrt(distanceToShore));
+
+		const float angle = (twoPiOverWavelength * sqrt(distanceToShore)) + (_CrestTime * twoPiOverPeriod);
+		result.y = amplitude * cos(angle);
```
moosichu, line 107: Increasing amplitude slightly as depth increases is a common thing to do - I didn't do it as it didn't seem neccessary - particles could end up increase the "height" of waves anyway. But it's worth bearing in mind. I have that change locally and am happy to check it in once the current state of the PR has been checked-out.

crest/Assets/Crest/Crest/Shaders/OceanInputs/GerstnerShared.hlsl
```diff
@@ -91,3 +88,24 @@ half4 ComputeGerstner(float2 worldPosXZ, float3 uv_slice)
...
+	if(depth > 0.0)
+	{
+		const float pi = 3.141;
+		const float twoPiOverWavelength = (2.0 * pi) / 2.5;
+		const float twoPiOverPeriod = (2.0 * pi) / 2.0;
+		const float amplitude = 0.12;
+		// Chop increases as depth increases
+		const float chopAmplitude = 2.0/(1.0+sqrt(distanceToShore));
+
+		const float angle = (twoPiOverWavelength * sqrt(distanceToShore)) + (_CrestTime * twoPiOverPeriod);
```
moosichu, line 106: The phase is done along the square root of the distance - this makes waves closer to the shoreline bunch together whilst leaving waves further-out in the distance more spread-out.

moosichu:
I've managed to fix the terrain-intersection issue, I have commented the code a bit-better and done some other things to improve the overall visual-result.

The next-task I think is going to be adding noise of some form to break-up the visual make-up of the shoreline waves - so they aren't just one-wavefront for the entire terrain.

huwb:
Thanks for the update and sorry i havent been able to contribute yet.

Just a thought on the one-wavefront - yeah an offset to the the phase based on world space noise or etc should help a lot (we have gpu noise functions in gpunoise.hlsl that could help). probably already your plan but just mentioning.

huwb:
Had a bit of a play, looking promising!

Regarding your horizontal motion todo above, yeah its tricky. There is the option to just add more and more math :) (c.f. https://twitter.com/greje656/status/1276611073077108736 ) but i think its hard to get something realistic this way. An option would be to look at something bespoke just for shallow gradient shoreline/beaches, for example baking out from a sim https://vimeo.com/261901539

Having the distance field for the shore, and aligning the wave spectrum to the shore, is awesome in itself. Maybe we could divide out PRs to tackle this separately from 'breaking waves'?

On this, the main problem seems to be the waves 'emerging from a point' like in the shallow bay area in main.unity. It would be worth trying to mix in a wind direction with the coast direction. This can mean that away from the coast, the waves tend to follow the wind direction, and then as it gets closer to the coast the shoreline direction takes over, and theres never a single point where waves appear to emanate from a single point.

The other mitigation might be to generate the waves using an OceanInput which means the user can control what appears in which areas of the world - i.e. they can place some geometry on the outside of the island where they might want shoreline-directed waves, but leave interior of the bay to have single direction waves - things like that. Does this make sense?

moosichu:
> Regarding your horizontal motion todo above, yeah its tricky. There is the option to just add more and more math :) (c.f. https://twitter.com/greje656/status/1276611073077108736 ) but i think its hard to get something realistic this way, as well as being quite difficult. An option would be to look at something bespoke just for shallow gradient shoreline/beaches, for example baking out from a sim https://vimeo.com/261901539

Yeah - I think I'm going to do a deep research dive on all the difference techniques now - and in the meant-time refine this branch, because as you said - if I cleanup the code there is nothing to prevent it from being merged as-is.

> Having the distance field for the shore, and aligning the wave spectrum to the shore, is awesome in itself. Maybe we could divide out PRs to tackle this separately from 'breaking/tumbling waves'? Not sure if this makes sense.

Yes - for sure! Will clean this branch up as my the next step!

> On the waves-directed-towards-shore, the main problem i noticed is the waves 'emerging from a point' like in the shallow bay area in main.unity. It would be worth trying to mix in a wind direction with the coast direction. This can mean that away from the coast, the waves tend to follow the wind direction, and then as it gets closer to the coast the shoreline direction takes over, and theres never a single point where waves appear to emanate from a single point.

I'm not sure I quite understand this. I think I will go with the approach described below if that makes sense? Could try this afterwards as well?

> The other mitigation might be to generate the waves using an OceanInput which means the user can control what appears in which areas of the world - i.e. they can place some geometry on the outside of the island where they might want shoreline-directed waves, but leave interior of the bay to have single direction waves - things like that. Does this make sense?

Yeah - I think since I've added the auto-dampening for when the distance-field is near-zero - I think the solution is to insert 0s into the parts of the distance field where you don't want waves to be generated. I'm currently leaning towards masking-*out* directional waves - but the other option could be to mask them in using an OceanInput as you've said. Which do you think would make more sense?

moosichu:
Yeah - whilst I think this might be in an "experimental" state for a while - If I cleanup the Ocean Depth Cache Script, add some basic-validation, disable all these features by default, and revert the main prefabs and scene - I think this can be merged-in to master and we can then tackle individual issues as they crop-up.

moosichu:
```CSharp
// The API for this will likely change - which is why sdfInjectors are currently hidden in
                    // the inspector
                    if (_noDepthSdfInjectors != null)
                    {
                        jumpFloodCommandBuffer.SetRenderTarget(voronoiPingPongTexture0);
                        jumpFloodCommandBuffer.SetViewProjectionMatrices(_depthCacheCamera.worldToCameraMatrix, _depthCacheCamera.projectionMatrix);
                        foreach (GameObject sdfInjector in _noDepthSdfInjectors)
                        {
                            if(sdfInjector == null)
                            {
                                continue;
                            }
                            jumpFloodCommandBuffer.DrawRenderer(sdfInjector.GetComponent<Renderer>(), new Material(Shader.Find("Crest/Inputs/Depth/Inject Signed Distance Field From Geometry")));
                        }
                    }
```

Saving this as a reference for later

crest/Assets/Crest/Crest/Shaders/Resources/SdfJumpFlood.compute
```diff
@@ -28,14 +28,14 @@ void SdfJumpFlood(uint3 id : SV_DispatchThreadID)
...
-	candidatePositions[4] = downIsInBounds                    ? _FromTexture[id.xy + uint2(0         , -_jumpSize)] : currentNearestPoint;
-	candidatePositions[5] = downIsInBounds && leftIsInBounds  ? _FromTexture[id.xy + uint2(-_jumpSize, -_jumpSize)] : currentNearestPoint;
-	candidatePositions[6] = leftIsInBounds                    ? _FromTexture[id.xy + uint2(-_jumpSize,          0)] : currentNearestPoint;
-	candidatePositions[7] = leftIsInBounds && upIsInBounds    ? _FromTexture[id.xy + uint2(-_jumpSize,  _jumpSize)] : currentNearestPoint;
+	candidatePositions[0] = upIsInBounds                      ? _FromTexture[id.xy + uint2(0,                       _jumpSize)] : currentNearestPoint;
+	candidatePositions[1] = upIsInBounds && rightIsInBounds   ? _FromTexture[id.xy + uint2(_jumpSize,               _jumpSize)] : currentNearestPoint;
+	candidatePositions[2] = rightIsInBounds                   ? _FromTexture[id.xy + uint2(_jumpSize,                       0)] : currentNearestPoint;
+	candidatePositions[3] = rightIsInBounds && downIsInBounds ? _FromTexture[id.xy + uint2(_jumpSize,        -(int) _jumpSize)] : currentNearestPoint;
```
huwb, line 34: just realised - i'm not sure how this is working - casting a negative number to a uint - does that work due to some kind of wrapping behaviour? or is directx being very lenient (as usual..).

other platforms may be less forgiving. is it worth subtracting as int, then casting the result to uint. or making everything operate on ints? the cast may be a non-trivial operation on gpu (i'm not sure but it seems feasible).

moosichu: I will do more research in this. The hope is that two's compliment means that if you have unsigned wrapping everything should work fine. But will see if I can represent this more explicitly.

PR #518, opened 2020-05-23
Bind Ocean Mask Data for ocean chunks not rendered due to occlusion culling

### LINKED ISSUE #479: [HDRP] Underwater draws through buildings

moosichu:
Manually set a bool on each ocean chunk that checks if the data is bound, a simpler and more general solution anyway. 

Fixes #479, the repro was harder as we were rendering the mask chunks even if culled in the latest release, but this was causing gaps to appear which now seem to be fixed.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/OceanChunkRenderer.cs
+++ b/crest/Assets/Crest/Crest/Scripts/OceanChunkRenderer.cs
@@ -24,6 +24,11 @@ namespace Crest
         PropertyWrapperMPB _mpb;
 
         public Renderer Renderer => _rend;
+        // We need to ensure that all ocean data has been bound for the mask to
+        // render properly - this is something that needs to happen irrespective
+        // of occlusion culling because we need the mask to render as a
+        // contiguous surface.
+        internal bool OceanDataHasBeenBound = true;
 
         // Cache these off to support regenerating ocean surface
         int _lodIndex = -1;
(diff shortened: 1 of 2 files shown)


huwb:
Took me a while to get this but I think I get it now - the guard didn't hold up with occlusion culling, because the gameobject/renderer was active?

Thumbs up from me, but could we use the underscore naming convention as it's a member variable, or maybe make it a property I guess?

huwb:
There's a small logical thing that binding is per camera, but we only support one camera properly right now, and it could be expanded to a bool per camera in the future, so I'm happy to go with this.

PR #526, opened 2020-05-30
Zero horizon safety margin multiplier if in doubt

### LINKED ISSUE #520: [HDRP] Gap between waves and underwater area.
There is a gap between the waves and the underwater area. 

It looks like this, see red arrow:

gap

It doesn't happen all of the time, so there's no way to have the camera coordinates at a given position. However, I made a video, see below.

Reproduction steps:

* Unity 2019.3.14f1, HDRP 7.3.1, Crest HDRP 4.2 (Released: May 20, 2020) from the Asset Store
* install Crest HDRP
* open the threeboats demo scene
* dive under water, slightly below the surface so that you have both the waves and the underwater area visible
* move around with the camera

Here's a video, the bug starts to appear at around 00:16:

https://www.youtube.com/watch?v=lA8YE4BIYc8

moosichu:
Only apply the horizon safety margin multiplier to horizon if and only
if concrete height of the camera relative to the water and the height of
the camera relative to the sea-level are the same. This ensures that in
incredibly turbulent water - if in doubt - use the neutral horizon.

Fixes #520 

I tested this in the three boats scene in HDRP, the incredibly localised water spike caused by the ocean liner was causing trouble, so this change makes the horizon-safety-margin multipler much more conservative and has worked really well in testing. 

Justification: the check that I have added essentially checks to see if the camera is inside a swell or a dip - in that case there is no need to apply the horizon safety margin multipler is counter-productive (and with time we might find that applying it inverted might be the most stable option - I just want to slowly iterate to a solution that means we no longer get any horizon-line issues).

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterPostProcess.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterPostProcess.cs
@@ -38,12 +38,14 @@ namespace Crest
         private RenderTexture _depthBuffer;
         private CommandBuffer _maskCommandBuffer;
         private CommandBuffer _postProcessCommandBuffer;
+        private readonly SampleHeightHelper _sampleHeightHelper = new SampleHeightHelper();
 
         private Plane[] _cameraFrustumPlanes;
 
         private Material _oceanMaskMaterial = null;
 
         private PropertyWrapperMaterial _underwaterPostProcessMaterialWrapper;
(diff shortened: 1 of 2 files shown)


huwb:
Thanks @moosichu!

Ok so it doesn't apply the offset in swells or dips? Think I get the gist.

Thumbs up from me

moosichu:
Exactly, cheers Huw!

PR #537, opened 2020-06-10
Support cinemachine virtual cams

huwb:
Status: WIP

Question: will we get errors if Cinemachine is not present, similar to: `Assembly has reference to non-existent assembly Cinemachine`


daleeidd:
> will we get errors if Cinemachine is not present, similar to: Assembly has reference to non-existent assembly Cinemachine

We can use assembly version defines in 2019 LTS. Been using them for XR in HDRP and it works well.

huwb:
Thanks @daleeidd .

I've actually created this PR off the wrong branch. I will create a new PR off the correct one now.

PR #538, opened 2020-06-11
Support cinemachine

huwb:
**Status**: Ready to merge (but will wait til after 4.3 i think)

Support cinemachine. It only exposes position/rotation, not a transform (the transforms i could get to were not updating properly, cinemachine is weird).

So to get in line with this, the viewer transform now gives a pos and forward vector, which cinemachine can provide. I'm not aware of any issues or negative impact from this.

The cinemachine-specific code only compiles in if Cinemachine is present (thanks @daleeidd)

This wont work in 2018 so we should probably leave out the asmdef changes (the rest is probably ok) when merging into the legacy branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 # Intro
 
-*Crest* is a technically advanced ocean renderer implemented in Unity3D 2018.4 and later.
+*Crest* is a technically advanced ocean renderer implemented in Unity3D 2019.4 and later.
 The version hosted here targets the **built-in render pipeline**, a link to the scriptable render pipeline versions (URP/HDRP) on the Asset Store are below.
 
 ![Teaser](https://raw.githubusercontent.com/huwb/crest-oceanrender/master/img/teaser5.png)
@@ -44,7 +44,7 @@ There is also a getting started video here: https://www.youtube.com/watch?v=qsge
 
 * Unity version:
   * The SRP assets on the Asset Store specify the minimum version required.
-  * Releases on this GitHub target the built-in render pipeline, and each release specifies which version of Unity it was developed on. Currently Unity 2018.3 or later is the minimum version.
+  * Releases on this GitHub target the built-in render pipeline, and each release specifies which version of Unity it was developed on. Currently Unity 2019.1 or later is the minimum version. A legacy version which supports [2018.4](https://github.com/crest-ocean/crest/tree/legacy/unity-2018) is available here.
 * *Crest* example content:
   * The content requires a layer named *Terrain* which should be added to your project.
   * The post processing package is used (for aesthetic reasons), if this is not present in your project you will see an unassigned script warning which you can fix by removing the offending script.
--- a/crest/Assets/Crest/Crest/Scripts/Collision/QueryBase.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Collision/QueryBase.cs
@@ -87,10 +87,20 @@ namespace Crest
             {
                 var lastIndex = _segmentAcquire;
 
-                _segmentAcquire = (_segmentAcquire + 1) % _segments.Length;
+                var newSegmentAcquire = (_segmentAcquire + 1) % _segments.Length;
 
-                // The last index should never increment and land on the first index - it should only happen the other way around.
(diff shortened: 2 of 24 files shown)


daleeidd:
Sure. I've added [test/asm-version-defines](https://github.com/crest-ocean/crest/tree/test/asm-version-defines). It is 2019 so I recommend cloning rather than checking out branch. It has Cinemachine installed. Uninstall it and check the Crest asmdef.

We will need to upgrade to 2019.
Then we will have assembly version defines available to us.

We will still need to add a GUID for Cinemachine in the asmdef. If the package isn't installed for a user or uninstalled then it still holds a spot without error.

The version defines allows us to create a defines for a package/version combination. So we would define CREST_CINEMACHINE and use it like UNITY_2019. (see example branch)

huwb:
Hi, i merged it in and it looks like i have a version that works well in 2019. Thanks for that!

So this is not compatible with 2018 at all? (not the end of the world). It seems like it almost works, but maybe the "holds a spot" behaviour that you mention above doesnt function - it gives an error for cinemachine:

> Assembly has reference to non-existent assembly 'Cinemachine' (Assets/Crest/Crest/Scripts/Crest.asmdef)

Is that the bit thats expected to fail? Just checking.

I guess the upgrade is the next thing - should i do it? could probably create a `legacy/unity-2018` branch or something just before upgrading.

daleeidd:
> Is that the bit thats expected to fail? Just checking.

Assembly version defines were introduced in 2019.1. I am surprised you are seeing them work at all. I just checked and I couldn't get the defines to work in 2018.

> I guess the upgrade is the next thing - should i do it? could probably create a legacy/unity-2018 branch or something just before upgrading.

I'm happy to do it. Is there anything we want to merge beforehand?

huwb:
Ok cool

Re things to merge beforehand - no i dont think so, at least from my end.

daleeidd:
I'll do it tomorrow if you haven't already.

huwb:
Go for it - thanks!

skaughtx0r:
Sorry to interject, but I'm curious what this PR is for?

I've been using Crest with Cinemachine for a long time with no issues.

huwb:
Thanks @skaughtx0r . Thats very interesting to hear.

We are working with an artist on an example scene which has a Main Camera with a CinemachineBrain, and a few CinemachineVirtualCamera's that are animated. The Main Camera moves around in the scene view, but the Ocean does **not** follow it.

Maybe its something with our cinemachine settings?

image

I tried a few variations of settings to no avail..

skaughtx0r:
@huwb That's weird, if the transform of the MainCamera is changing, it sounds like the Viewpoint is just not set to the MainCamera GameObject. I have the Viewpoint manually assigned to the camera GameObject. Technically I have another game object that is the real Viewpoint target that takes the camera FOV into account. I just tried it with it set to the MainCamera though and it works fine.

I do have a slightly unusual setup, I have a CameraController GameObject that has the Cinemachine brain. Then the MainCamera GameObject is a child, this is mainly for VR mode. At run time I check if I'm not in VR then I removed that brain and add the brain to the MainCamera (I'm not completely sure why I did that, because I also just tested it without doing that and it seems to work fine with the Cinemachine brain on a parent GameObject).

I am using Unity 2018.4 and Cinemachine 2.2.9.

CinemachineHierarchy

image

image

huwb:
Yeah i really dont get it. I can see my main camera moving in the scene view, and i can see the position changing on the Transform in the Inspector, but if I print out `transform.position` it is not changing.

I also created a child transform and attached the ocean to that instead, and verified its using the child, but again the transform position in code does not change.

Ok it looks like i do get the position if move the ocean to `Update()` instead of `LateUpdate` :/.

Do you have any changes to the Script Execution Order in your project?

skaughtx0r:
Oh that's interesting.

```
Crest.OcenRenderer            -100
Cinemachine.CinemachineBrain  100
```

Maybe that's it?

huwb:
Moving ocean update order to 200 - after cinemachine (100) - seems to fix it for me, so i made a new PR #552 . If we are happy with that then i'll close this one.

huwb:
Closing

PR #543, opened 2020-06-13
Predicate enabled/disabled state of GUI based on a named property

huwb:
**Status**: Ready to merge

Thought i'd try a fun experiment - give a easy way to enable/disable GUI fields based on other fields. Example:

image

Here the 'search..' option is disabled because a Light is provided in the reference field above it (so no search will be done).

I find this kind of behaviour useful to know what parameters are active, and to help understand how a system works.

Issue: When multiple property drawers are provided, only the first one takes effect. Unforunately this means this feature can not be used in conjunction with another drawn attribute like `Range()`.

Let me know what you guys think.


huwb:
Yes that would work!

PR #544, opened 2020-06-14
Add dynamic resolution and camera change test components

daleeidd:
For testing:
- Camera switching
- Dynamic resolution

We have had problems with both for underwater. I have used version defines so it works in both built-in and SRP.

I have also enabled _Allow Dynamic Resolution_ on the camera and _Enable Frame Timing Stats_ in the project settings to stop Unity complaining on the camera inspector.


huwb:
Looks good to me. Are you planning to add multiple cameras to a scene or will you add them when needed?

daleeidd:
I was going to add them as needed (Command+D and disable duplicate so pretty easy) for the time being. The component won't be shipped so we would have to add it to the scene as needed too.

daleeidd:
Actually... I will add a simple test scene that will live under Development.

daleeidd:
Added a test scene. I split the CameraTesting component into their respective features too.

daleeidd:
@huwb ok to merge this one after those changes?

PR #549, opened 2020-06-16
WaterBody script for tile culling

huwb:
**Status**: Ready for review

A piece of the local-water-bodies branch that can be merged in separately.

This change adds a `WaterBody` component which registers its AABB with the ocean. During update, if there is one or more water bodies present, bounding box overlap checks will be performed between each ocean tile and each water body. If there are no overlaps, the tile is turned off. Turning off a tile means OnWillRenderObject will not get called, no culling will be performed, no draw call, etc, so this is a nice early off-switch where water should not be present.

These changes unify the code a bit towards what is in HDRP. The HDRP version also loops over the tiles for underwater purposes and i expect we will be able to combine concerns nicely.

**Testing**:

Add WaterBody component to a gameobject. Set x&z scale to set the size. Water should cull outside of the aabb drawn by the gizmo.

**Issues**:

- [x] Recompiling code in edit mode seems to put it in an odd state (until entering play mode or doing something else to reset state). I wasn't able to figure out why yet. My debugger doesnt seem to want to attach these days and all my debug prints seemed to indicate things were fine.. Will try again later. Any testing/review would anyway be helpful. **FIXED**
- [ ] The 'Interior' tiles at lod0 seems to get too large bounds and i could not figure out why. This means they sometimes dont cull. This is annoying but hopefully not a blocker .. for now.
Thanks

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Helpers/UnderwaterEffect.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Helpers/UnderwaterEffect.cs
@@ -26,6 +26,9 @@ namespace Crest
         [Tooltip("Copy ocean material settings on each frame, to ensure consistent appearance between underwater effect and ocean surface. This should be turned off if you are not changing the ocean material values every frame."), SerializeField]
         bool _copyParamsEachFrame = true;
 
+        [SerializeField]
+        bool _turnOffOutsideWaterBodies = true;
+
         [Header("Advanced")]
 
         [Tooltip("This GameObject will be disabled when view height is more than this much above the water surface."), SerializeField]
@@ -95,7 +98,7 @@ namespace Crest
             }
 #endif
 
-            if (OceanRenderer.Instance == null)
+            if (OceanRenderer.Instance == null || !ShowEffect())
             {
                 _rend.enabled = false;
                 return;
@@ -156,6 +159,32 @@ namespace Crest
 
                 _rend.SetPropertyBlock(_mpb.materialPropertyBlock);
             }
+        }
+
+        bool ShowEffect()
+        {
+            if (_turnOffOutsideWaterBodies && OceanRenderer.Instance.WaterBodies.Count > 0)
+            {
+                var inOne = false;
+                float x = transform.position.x, z = transform.position.z;
+                foreach (var body in OceanRenderer.Instance.WaterBodies)
+                {
+                    var bounds = body.AABB;
+                    if (x >= bounds.min.x && x <= bounds.max.x &&
+                        z >= bounds.min.z && z <= bounds.max.z)
+                    {
+                        inOne = true;
+                        break;
+                    }
+                }
+
+                if (!inOne)
+                {
+                    return false;
+                }
+            }
+
+            return true;
         }
 
         static Mesh Mesh2DGrid(int dim0, int dim1, float start0, float start1, float width0, float width1, int divs0, int divs1)
--- a/crest/Assets/Crest/Crest/Scripts/Helpers/UnderwaterEffect.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Helpers/UnderwaterEffect.cs
@@ -26,9 +26,6 @@ namespace Crest
         [Tooltip("Copy ocean material settings on each frame, to ensure consistent appearance between underwater effect and ocean surface. This should be turned off if you are not changing the ocean material values every frame."), SerializeField]
         bool _copyParamsEachFrame = true;
 
-        [SerializeField]
-        bool _turnOffOutsideWaterBodies = true;
-
         [Header("Advanced")]
 
         [Tooltip("This GameObject will be disabled when view height is more than this much above the water surface."), SerializeField]
@@ -37,6 +34,8 @@ namespace Crest
         bool _overrideSortingOrder = false;
         [Tooltip("If the draw order override is enabled use this new order value."), SerializeField]
         int _overridenSortingOrder = 0;
+        [Tooltip("Disable underwater effect outside areas defined by WaterBody scripts, if such areas are present."), SerializeField]
+        bool _turnOffOutsideWaterBodies = true;
 
         // how many vertical edges to add to curtain geometry
         const int GEOM_HORIZ_DIVISIONS = 64;
--- a/crest/Assets/Crest/Crest/Scripts/Helpers/UnderwaterEffect.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Helpers/UnderwaterEffect.cs
@@ -162,11 +162,11 @@ namespace Crest
 
         bool ShowEffect()
         {
-            if (_turnOffOutsideWaterBodies && OceanRenderer.Instance.WaterBodies.Count > 0)
+            if (_turnOffOutsideWaterBodies && WaterBody.WaterBodies.Count > 0)
             {
                 var inOne = false;
                 float x = transform.position.x, z = transform.position.z;
-                foreach (var body in OceanRenderer.Instance.WaterBodies)
+                foreach (var body in WaterBody.WaterBodies)
                 {
                     var bounds = body.AABB;
                     if (x >= bounds.min.x && x <= bounds.max.x &&
--- a/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
+++ b/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
@@ -254,6 +254,8 @@ namespace Crest
         float _lodAlphaBlackPointFade;
         float _lodAlphaBlackPointWhitePointFade;
 
+        bool _canSkipCulling = false;
+
         readonly int sp_crestTime = Shader.PropertyToID("_CrestTime");
         readonly int sp_texelsPerWave = Shader.PropertyToID("_TexelsPerWave");
         readonly int sp_oceanCenterPosWorld = Shader.PropertyToID("_OceanCenterPosWorld");
@@ -335,6 +337,8 @@ namespace Crest
             {
                 lodData.OnEnable();
             }
+
+            _canSkipCulling = false;
         }
 
         private void OnDisable()
@@ -722,7 +726,10 @@ namespace Crest
         {
             // If there are local bodies of water, this will do overlap tests between the ocean tiles
             // and the water bodies and turn off any that don't overlap.
-            if (WaterBody.WaterBodies.Count == 0) return;
+            if (WaterBody.WaterBodies.Count == 0 && _canSkipCulling)
+            {
+                return;
+            }
 
             foreach (OceanChunkRenderer tile in _oceanChunkRenderers)
             {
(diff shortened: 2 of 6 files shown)


crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
```diff
@@ -711,6 +721,42 @@ void LateUpdateLods()
             _lodDataShadow?.UpdateLodData();
         }
 
+        void LateUpdateTiles()
+        {
+            // If there are local bodies of water, this will do overlap tests between the ocean tiles
+            // and the water bodies and turn off any that don't overlap.
+            if (_waterBodies.Count == 0) return;
```
moosichu, line 728: We want to re-enable all the ocean chunk renderers in this case - default back to a full ocean. Otherwise you just get all the previously enabled tiles just drifting around.

crest/Assets/Crest/Crest/Scripts/Helpers/UnderwaterEffect.cs
```diff
@@ -158,6 +160,32 @@ private void LateUpdate()
             }
         }
 
+        bool ShowEffect()
+        {
+            if (_turnOffOutsideWaterBodies && OceanRenderer.Instance.WaterBodies.Count > 0)
+            {
+                var inOne = false;
+                float x = transform.position.x, z = transform.position.z;
+                foreach (var body in OceanRenderer.Instance.WaterBodies)
```
moosichu, line 169: Wouldn't having a function like "IsInsideWaterBody" be useful on the ocean renderer instance? As this pattern appears to crop-up a couple of times.

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
```diff
@@ -711,6 +721,42 @@ void LateUpdateLods()
...
+                foreach (var body in _waterBodies)
+                {
+                    var bounds = body.AABB;
+
+                    bool overlapping =
+                        bounds.max.x > chunkBounds.min.x && bounds.min.x < chunkBounds.max.x &&
+                        bounds.max.z > chunkBounds.min.z && bounds.min.z < chunkBounds.max.z;
+                    if (overlapping)
+                    {
+                        overlappingY = bounds.center.y;
```
moosichu, line 750: is this local variable "overlappingY" needed for anything? or is it there for debugging of some form?

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
huwb: Yeah its a good point.. i guess it probably wants state so that it doesnt need to loop over all the renderers each frame? i can look at that

crest/Assets/Crest/Crest/Scripts/Helpers/UnderwaterEffect.cs
huwb: yeah it could be, ill have a look

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
huwb: nope, removed!

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
```diff
@@ -244,8 +242,15 @@ public ITimeProvider TimeProvider
...
         public float ViewerHeightAboveWater { get; private set; }
 
+        List<LodDataMgr> _lodDatas = new List<LodDataMgr>();
+
+        List<OceanChunkRenderer> _oceanChunkRenderers = new List<OceanChunkRenderer>();
+
         SampleHeightHelper _sampleHeightHelper = new SampleHeightHelper();
 
+        public List<WaterBody> WaterBodies => _waterBodies;
+        List<WaterBody> _waterBodies = new List<WaterBody>();
```
daleeidd, line 252: It might be easier to have this as a static list on the WaterBody component instead. This is how Unity does it with the NavMeshComponents.

huwb:
thanks @daleeidd 

i noticed that issue with tiles near the camera too and have been looking - sometimes a tile close to the camera inexplicably fails to cull:

image

it seems that sometimes, some of the tile meshes have 2x larger bounds. it seemed to be correllated with `PatchType.Interior` but after stripping away a bunch of code i now have *inverted* behaviour where the interior ones are fine but the rest are 2x bigger. i'm not sure what to try next. will keep thinking about it

huwb:
> there is a nasty feedback loop as we are reading/writing _mesh.bounds in OceanChunkRenderer update

Even though this does not seem to cause the above bug (of interior tiles having 2x too large bounds), it does actually cause the other issue where after a recompile the tile culling appears to break. Simply storing the original bounds on the chunk component instead of grabbing it from the mesh appears to fix this.

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
huwb: yeah that made things better in a few ways - good shout!

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
```diff
@@ -792,6 +838,9 @@ private void CleanUp()
 
             FlowProvider.CleanUp();
             FlowProvider = null;
+
+            _oceanChunkRenderers.Clear();
+            _waterBodies.Clear();
```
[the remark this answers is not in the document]
huwb: fixed by moving the list into WaterBody

crest/Assets/Crest/Crest/Scripts/WaterBody.cs
```diff
@@ -0,0 +1,91 @@
...
+        {
+            if (_runValidationOnStart)
+            {
+                Validate(OceanRenderer.Instance, ValidatedHelper.DebugLog);
+            }
+        }
+
+        private void OnDrawGizmos()
+        {
+            Gizmos.color = Color.blue;
```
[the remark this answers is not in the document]
huwb: i made it white for now as a minimum fix for visibility. i wasnt sure how the box or wall should look but happy to see proposals if you feel like proposing sth. will resolve this for now.

crest/Assets/Crest/Crest/Scripts/WaterBody.cs
```diff
@@ -0,0 +1,91 @@
...
+                    $"Water body {gameObject.name} has a very small size (the size is set by the scale of its transform). This will be a very small body of water. Is this intentional?",
+                    ValidatedHelper.MessageType.Error, this
+                );
+
+                return false;
+            }
+
+            return true;
+        }
+    }
```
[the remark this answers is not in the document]
huwb: oh thanks! done

huwb:
> The only issue is that it doesn't work in edit mode without entering/exiting play mode (and after clicking rebuild ocean button). Not sure if this is easily solved. Seeing this on Mac and pretty certain I saw this on Windows too.

I'm not sure that i'm seeing this. Maybe its fixed by my latest commits. if you feel like retesting then let me know if its better now

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
```diff
@@ -711,6 +721,42 @@ void LateUpdateLods()
...
+                        bounds.max.z > chunkBounds.min.z && bounds.min.z < chunkBounds.max.z;
+                    if (overlapping)
+                    {
+                        overlappingY = bounds.center.y;
+                        overlappingOne = true;
+                        break;
+                    }
+                }
+
+                tile.Rend.enabled = overlappingOne;
```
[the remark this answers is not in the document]
huwb: erm I don't understand, could you expand?

crest/Assets/Crest/Crest/Scripts/Helpers/UnderwaterEffect.cs
huwb: i think i feel inclined to leave as is as the code is simple enough, and its executed ~100 times each frame so i'm not sure incurring the overhead of a function call is worth it. let me know if that sounds ok.

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
huwb: i put in a flag that defers this early-out until after a final culling pass has been performed. so if the number of water bodies is 0, it does a final culling pass (which enables everything), then sets the flag true.

this seemed to be the safest way to do it. let me know if this works

crest/Assets/Crest/Crest/Scripts/WaterBody.cs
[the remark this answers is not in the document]
daleeidd: Partially transparent box looks nice:
```
var oldColor = Gizmos.color;
Gizmos.color = new Color(1f, 1f, 1f, 0.5f);
var center = AABB.center;
if (OceanRenderer.Instance != null && OceanRenderer.Instance.Root != null)
{
    center.y = OceanRenderer.Instance.Root.position.y;
}
Gizmos.DrawCube(center, new Vector3(AABB.extents.x, 1f, AABB.extents.z));
Gizmos.color = oldColor;
```

Let me know what you think.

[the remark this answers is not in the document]
daleeidd: Change to OnDrawGizmosSelected if you feel it is too intrusive.

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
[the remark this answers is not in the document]
daleeidd: Seems like it works fine now. Before draw render bounds didn't work for culled tiles for me.

daleeidd:
> I'm not sure that i'm seeing this. Maybe its fixed by my latest commits. if you feel like retesting then let me know if its better now

Great. That is fixed for me too.

> Pushed the tests so far to fix/interior-tile-bounds

I had a look at the 2018 branch and they appear large there too. The padding value is 7-8. Not sure if that is correct. But yeah, no point holding up this.

crest/Assets/Crest/Crest/Scripts/WaterBody.cs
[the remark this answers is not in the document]
huwb: Thanks, looks good to me, i'll go with that! i will move to onselected as well, not sure why i didnt have it on this already

huwb:
thanks, i think i've addressed something but let me know about the open threads @moosichu

crest/Assets/Crest/Crest/Scripts/Helpers/UnderwaterEffect.cs
moosichu: Sure thing. Although it's worth pointing out that the C# compiler can inline functions if they are suitable for it and you can tell it to do so:

https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.methodimploptions?view=netcore-3.1

I believe this works with IL2CPP as Unity use it in a lot of their internal code.

huwb: Thanks ill be lazy and go with whats there then, and keep this for future record. i think we'll need to review the cpu perf at some point which might be a good time to try this out and verify it doesnt hurt perf. (i could see this going to burst jobs as well)

huwb:
thanks all, merged

PR #552, opened 2020-06-17
Cinemachine attempt 2 - update ocean after cine

huwb:
@daleeidd @skaughtx0r - attempt 2 at cinemachine fix - much simpler than #538 

Push ocean update to 200, which is after cinemachine (100), which seems to make it work for me.


huwb:
Thanks. I don't really get why I'm getting it then, but I believe in this change regardless, so will merge it

PR #553, opened 2020-06-17
Scale gerstner waves using vert colours

huwb:
Option to use vertex colour as weight to scale waves, when using the gerstner-from-geo shader. This was the ultimate intention but did not have test data before.

@EliGould tested this and indicated it works well.


huwb:
Thanks, merging!

PR #558, opened 2020-06-21
Invertable default clip state

huwb:
**Status**: Ready to merge

Another big piece of #361 

Allows 'default to clip everything' so that limited areas can easily be unclipped. This can be much more convenient for defaulting to 'no water' and then including areas for each body of water. This adds little to no run-time cost, so it seemed like it was worth adding.

**Testing**

Set *Default Clipping State* on the ocean component to *Remove Everything*, then add clip inputs that use the *ClipSurfaceIncludeArea* material. I could push temporary test data that has this set up if this is useful.

**Issues**

None that I'm aware of.


crest/Assets/Crest/Crest/Shaders/OceanGlobals.hlsl
```diff
@@ -17,6 +17,7 @@ float _TexelsPerWave;
 float3 _OceanCenterPosWorld;
 float _SliceCount;
 float _MeshScaleLerp;
+float _ClipByDefault;
```
daleeidd, line 20: The TODO at the top of this file says we should prefix these with Crest.

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/ClipSurfaceIncludeArea.shader
```diff
@@ -0,0 +1,49 @@
...
+			};
+
+			Varyings Vert(Attributes input)
+			{
+				Varyings o;
+				o.positionCS = UnityObjectToClipPos(input.positionOS);
+				return o;
+			}
+
+			half Frag(Varyings input) : SV_Target
```
daleeidd, line 42: Changing half to half4 will clear an error on MacOS raised by the material preview:
```
Metal: Error creating pipeline state (Crest/Inputs/Clip Surface/Include Area): fragment shader color output does not have enough components for the pixel format (MTLPixelFormatRGBA16Float)
```

crest/Assets/Crest/Crest/Shaders/OceanGlobals.hlsl
huwb: Good point, fixed

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/ClipSurfaceIncludeArea.shader
huwb: Good catch, thanks, fixed

huwb:
> The placement of our materials is a bit strange though

Well spotted. I think i have made it more sane/clean now.

I'll merge as you guys sounded happy but if you have any follow up feedback let me know and i'll get it actioned.

Thanks!

PR #561, opened 2020-06-27
Wizard for placing water bodies

huwb:
Part of #361 

Window for creating local bodies of water.

This is similar to what is in #361, except i gave the window a state (idle vs placing) so that it doesnt always show the translation gizmo, and i removed the proxy object option as it only shows when in placement mode.

I also moved some gerstner validation which was firing a warning while gerstner waves were being created.


moosichu:
Sweet! Will have a look eitehr later this afternoon or tomorrow. :)

huwb:
Thanks.

What do you mean by other primitives?

daleeidd:
Meshes. Like spheres and capsules.

huwb:
Ah gotcha.

I think i'd be looking for 2d primitives - quads, circles, etc. The primitive will probably feed the sea level as well so wants to be flat. Not sure if using 3d primitives still makes sense.

PR #562, opened 2020-06-27
Fix wave directions which were 180deg off. Gulp.

huwb:
As pointed out on discord, the waves are rotated 180 degrees. This is due to a sign error in the gerstner code.

My first thought was to leave it as is - however im hoping we have more flexible wave distributions in the future where the wave directions can vary freely in the world, and Tom's shoreline stuff may also utilise this, so i think i'd rather address the problem instead of adding more and more hacks that compensate for it.

Users will have to rotate the waves 180deg to compensate for this fix. If we're happy going for this then i'll fixup the example content.

What do you guys think?

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstnerBatched.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstnerBatched.cs
@@ -583,7 +583,7 @@ namespace Crest
                 float k = 2f * Mathf.PI / _wavelengths[j];
 
                 float x = Vector2.Dot(D, pos);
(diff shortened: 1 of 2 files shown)


moosichu:
It's a breaking change so maybe this could be a good thing to hold back for Crest "5.0" - as the major version bump implies major (potentially breaking) changes. Is there we a tool could provide to allow waves to easily be flipped authoring side?

huwb:
Yeah good point, can wait for 5.0

I'll think about the fix-up, need to think about if there's a safe way to auto change authored data. I'll follow up but any ideas welcome in the meantime.

PR #573, opened 2020-07-03
Add VR/XR Single Pass Instanced support

daleeidd:
I didn't see any issues when testing. I am not able to test SPI with the new XR framework though - only legacy VR SPI. Surprisingly simple to add.


daleeidd:
Thank you

Issue #523, opened 2020-05-26
Height Queries in FixedUpdate no longer work with VSync disabled


skaughtx0r:
**Describe the bug**
A recent change (501680bb6977d45c4ee45ce1ea354cee8dabead1) in QueryBase to clear guids makes it so height queries are now bound to the game's frame rate. If the frame rate is really high, all of the height queries in the FixedUpdate loop get returned as 0.

**Versions**
Crest: master branch
built-in render pipeline
Unity 2018.4.10f1

**To Reproduce**
Open the crest project, load threeboats scene. **Disable vsync** in the ProjectSettings -> Quality. This mostly seems to happen if the frame rate is well above 60fps.

**Platform**
 - Happens in both editor and standalone, the higher the framerate, the more likely it is to happen.
 - Platform: Windows

**Hardware**
GPU: NVIDIA RTX 2080

huwb:
Sorry no reply before now.

Yeah this is potentially a thorny one. I realised im relying pretty heavily on the async gpu readback getting the queries back each frame. i can imagine that if gpu and cpu are free to update at different rates, results may not come back in a frame. With the latest version, guids are cleared each frame so it depends on results coming back every frame.

Ideas:
* Not support turning off vsync (?)
* Put the guid clearing on a optional switch, so it can at least be worked around. will leak guids when this switch is turned on.
* Instead of clearing the guids and flipping query registers etc on update, flip on QueryBase.DataArrived() instead. this might be a significant change. I dont have time now to look into it.

Let me know your thoughts for now..

huwb:
I tried the third point in a test branch - commit referenced above. would you be up for trying it? its a hack but would be interested to know if it helps.

skaughtx0r:
Yeah, I'll try it out, thanks!

skaughtx0r:
@huwb This seems to work correctly in the Crest threeboats scene now. Unfortunately, I'm having some issues testing with my game because I'm using the `feature/post-proc-underwater-2` branch. Merging this branch into that has some file conflicts that I'm not completely sure how to resolve.

skaughtx0r:
@huwb curious what's the status on this? It seemed like the changes fixed it for me.

huwb:
We've had this issue reported by someone else. Need to get this fixed.

I don't feel good about what is prototyped in this branch. The steps to clean it up are not clear. And I don't feel flipping on DataArrived is robust. What if a query fails for whatever reason, seems like that could make the system grind to a halt.

[...]

PR #575, opened 2020-07-04
Store query age and discard old ones. Repack query segments each frame

huwb:
**Status**: Ready to merge

This attempt 2 of PR i created yesterday #574

This solves problem #523 where queries not working when vsync disabled, which was due to FixedUpdate being called infrequently and query guids were being lost.

This change makes queries persistent, so that it doesnt matter if FixedUpdate is missed for a few frames, but if a query gets too old it is automatically discarded. This is achieved by storing a frame number with the query segment information.

Importantly this PR also recomputes query data segment indices each frame. This effectively compactifies the data every frame, instead of accumulating air bubbles in the data that would waste compute processing.

This should fix the problem, and also not require changes to the API and increased complexity for calling code that #574 added.

@skaughtx0r FYI

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Collision/QueryBase.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Collision/QueryBase.cs
@@ -55,7 +55,8 @@ namespace Crest
         /// </summary>
         class SegmentRegistrar
         {
-            public Dictionary<int, Vector2Int> _segments = new Dictionary<int, Vector2Int>();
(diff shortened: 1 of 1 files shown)


huwb:
> I have to disable the higher end of the spectrum. I think you are aware of this with surface velocities issue

Could you speak more to this?

huwb:
I'll merge for now, let me know about other issue

daleeidd:
> Could you speak more to this?

Sure. I will see if I can isolate and open an issue.

PR #582, opened 2020-07-11
Feature to clip surface underneath terrain

huwb:
**Status**: Ready to merge

Does not require clip surface feature to be enabled, or a new type of ocean input to be set up - its purely based on depth. Pretty nifty.

This is not for performance, its simply to remove water under the terrain so it does not get in the way of other things.

Before

image

After

image


crest/Assets/Crest/Crest/Shaders/Ocean.shader
```diff
@@ -171,6 +171,8 @@ Shader "Crest/Ocean"
 		[Header(Clip Surface)]
 		// Discards ocean surface pixels. Requires 'Create Clip Surface Data' enabled on OceanRenderer script.
 		[Toggle] _ClipSurface("Enable", Float) = 0
+		// Clips purely based on water depth
+		[Toggle] _ClipUnderTerrain("Clip under terrain", Float) = 0
```
daleeidd, line 175: Using "Terrain" might cause confusion since it will happen to anything that adds depth. Some might take it as Unity Terrain too. "ClipUnderDepthCacheObjects" or something doesn't sound as accessible though.

huwb: Yeah fair point! How about "clip below land" perhaps?

huwb:
Thanks. Hopefully the layer can be specified on the depth cache to tick that box.

crest/Assets/Crest/Crest/Shaders/Ocean.shader
daleeidd: We can probably keep terrain with something like "Clip Under Terrain (requires depth cache)"

huwb: Ok can go for that

huwb:
@StarChick971 if you have feedback let us know. I'll aim to merge this soon.

huwb:
Merging

PR #583, opened 2020-07-11
Depth cache populate on demand instead of every frame

huwb:
**Status**: Ready to merge

Part of the ocean-in-edit-mode changes included having the depth cache automatically populate itself every frame when in edit mode. This was to have the ocean automatically respond to changes in the terrain/sea floor depth.

In 4.4 we fixed a bug where the cache was always populating every frame, and it was found because populating the cache is a significant cost for realistic scenes which users noticed.

I regret making the depth caches update every frame - I feel automatically responding to changes in terrain is not worth the significant performance/power cost of populating the cache every frame. This PR disables that behaviour and instead ensures the that *Populate cache* button is visible in the inspector so that users can populate it themselves if they change the terrain.

Additionally, this PR also enables the "save to texture" command in edit mode; this was previously only available in play mode.

Let me know what you think.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
@@ -343,7 +343,7 @@ namespace Crest
                 dc.PopulateCache();
             }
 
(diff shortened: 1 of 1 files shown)


huwb:
Thanks @daleeidd !

PR #584, opened 2020-07-12
Use local shader keywords

daleeidd:
Required no scripting changes. I haven't tested everything yet, but I am confident it works fine. Providing we enable keywords on the material instead using Shader.EnableKeyword (which we do. Other than for the compute shaders where I think we need them for the moment at least).

This might break someone's project if they need global keywords with Crest. But I don't really see why they would.


daleeidd:
Thanks!

PR #590, opened 2020-07-16
Fix underwater effect appearing at top of screen when looking up

### LINKED ISSUE #572: Underwater effect appears as a bar at top of screen
**Describe the bug**
Depending on camera orientation and settings, the underwater effect can appear as a bar at top of the viewport.

**Screenshots / video**

Main scene. Camera settings untouched

**Versions**
Crest 4.3

**To Reproduce**
Open the main scene. Descend to height where underwater effect is active. And look up.

It is dependent on the viewport height. Decreasing the FoV on the camera makes the issue appear sooner (closer to the horizon). Decrease it to something like 10 and it will be hard to miss.

**Platform**
 - Editor
 - Mac
 - Windows

**Additional context**
It is caused by the _Horizon Safety Margin Multiplier_. Setting it to zero is a workaround.

daleeidd:
Fixes #572

There is probably a proper way to fix this in the horizon calculation. But this seemed like the easiest way for me.

All it does is check to see if the horizon is within the far clip plane. And if it is not, set the _Horizon Safety Margin Multiplier_ to zero since this is the cause of the bar of underwater fog.

Is there a reason we are using the _camera.farClipPlane_ value rather than a hard-coded one?

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterPostProcessUtils.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterPostProcessUtils.cs
@@ -294,6 +294,14 @@ namespace Crest
                     v_world[i] = camera.ViewportToWorldPoint(v_screenXY_viewZ[i], eye);
                 }
 
+                // This is a temporary fix to the horizonSafetyMarginMultiplier thinking the horizon is the top of the
+                // screen causing the underwater effect to render from the top of the screen downwards when looking up.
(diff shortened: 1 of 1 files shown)


huwb:
Thanks looks good to me.

Yeah the underwater stuff generally needs a deep dive to resolve the issues. This is good for now though.

A hard coded far clip plane sounds not good, if that's a bug could you propose the fix?

daleeidd:
I made a typo. I meant that I think we should be using a hard-coded far clip plane. But on further thinking, it won't be necessary. Thanks. I'll merge.

PR #591, opened 2020-07-17
Fix dynamic waves crashing without flow or depth sims

### LINKED ISSUE #588: Dynamic Waves crashes when Sea Floor Depth and Flow are not enabled
**Describe the bug**
Dynamic Waves no longer work correctly when Sea Floor Depth and Flow simulations are not running. The dynamic waves dissipate as they speed away.

Loading Unity with both disabled will prevent Crest from working. It crashes builds.

**Screenshots / video**

Loading Unity with a scene with both Flow and Sea Floor Depth disabled and Dynamic Waves enabled.

**Log**

From crash in build:
```
Begin MonoManager ReloadAssembly
- Completed reload, in  0.095 seconds
Metal RecreateSurface[0x111b856d0]: surface size 2880x1800
UnloadTime: 0.827029 ms

Native stacktrace:

	0   libmonobdwgc-2.0.dylib              0x0000000116413656 mono_handle_native_crash + 242
	1   libmonobdwgc-2.0.dylib              0x000000011637432d mono_sigsegv_signal_handler + 220
	2   libsystem_platform.dylib            0x00007fff6ee205fd _sigtramp + 29
	3   ???                                 0x0000000000000000 0x0 + 0
	4   UnityPlayer.dylib                   0x000000010f1500a9 _ZdaPvRKSt9nothrow_t + 413625
	5   UnityPlayer.dylib                   0x000000010f14fce3 _ZdaPvRKSt9nothrow_t + 412659
	6   UnityPlayer.dylib                   0x000000010f30833e _ZdaPvRKSt9nothrow_t + 2216526
	7   UnityPlayer.dylib                   0x000000010f308630 _ZdaPvRKSt9nothrow_t + 2217280
	8   UnityPlayer.dylib                   0x000000010f3086ae _ZdaPvRKSt9nothrow_t + 2217406
	9   UnityPlayer.dylib                   0x000000010f3088b8 _ZdaPvRKSt9nothrow_t + 2217928
	10  UnityPlayer.dylib                   0x000000010f307cf6 _ZdaPvRKSt9nothrow_t + 2214918
	11  UnityPlayer.dylib                   0x000000010f404bcc _ZdaPvRKSt9nothrow_t + 3250908
	12  libsystem_pthread.dylib             0x00007fff6ee2c109 _pthread_start + 148
	13  libsystem_pthread.dylib             0x00007fff6ee27b8b thread_start + 15
```

**Versions**
Built-in on master. I was debugging some other issues, and I was on a commit just after the 2019 upgrade and it was there too. (I'll have to look further back to find it)

**To Reproduce**
Use the boat scene. Disable the Sea Floor Depth simulation. Enter play mode and no dynamic waves will be working. Restart Unity and the ocean won't be working. Or do a build and it will crash.

**Platform**
 - Editor
 - Standalone
 - Mac
 - Windows

**Hardware**
MacBook Pro 2014. No dedicated graphics

**Additional context**
Toggling the Sea Floor Depth simulation option in play mode is the most revealing.

I narrowed it down to being in the LodDataMgrDynWaves.SetAdditionalSimParams function. If flow's or depth's data is not bound, then the dynamic waves simulation fails.

I will have to look further into this.

daleeidd:
Fixes #588 

The issue was the __BindData_paramIdPosScales_ not being set as suggested by @huwb. I went with just calling _BindData_, but we could set only that property.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgr.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgr.cs
@@ -114,7 +114,7 @@ namespace Crest
         }
 
         // Avoid heap allocations instead BindData
-        private Vector4[] _BindData_paramIdPosScales = new Vector4[MAX_LOD_COUNT + 1];
+        protected Vector4[] _BindData_paramIdPosScales = new Vector4[MAX_LOD_COUNT + 1];
         // Used in child
         protected Vector4[] _BindData_paramIdOceans = new Vector4[MAX_LOD_COUNT + 1];
         protected virtual void BindData(IPropertyWrapper properties, Texture applyData, bool blendOut, ref LodTransform.RenderData[] renderData, bool sourceLod = false)
--- a/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgrPersistent.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgrPersistent.cs
@@ -126,6 +126,9 @@ namespace Crest
                     DataTexture
                 );
 
+                // Bind current data
+                BindData(_renderSimProperties, null, false, ref OceanRenderer.Instance._lodTransform._renderData, false);
+
                 _renderSimProperties.DispatchShaderMultiLOD();
 
                 for (var lodIdx = lodCount - 1; lodIdx >= 0; lodIdx--)


huwb:
Ah awesome. Does it makes sense to bind the texture in the same call - pass true as the third argument to binddata and remove the preceding line?

Either way happy for this to be merged!

daleeidd:
Thanks.

I had a look and it looks like that texture is for the compute shader which BindData doesn't handle.

Issue #587, opened 2020-07-15
unused code in ObjectWaterInteraction.cs


klauskobald:
I think that line can be removed, right?

            var rnd = 1f + _noiseAmp * (2f * Mathf.PerlinNoise(_noiseFreq * ocean.CurrentTime, 0.5f) - 1f);

rnd is never used

daleeidd:
Thanks. It can be. I will have to do a sweep for unused code. I am unsure if this was planned to be used because it became unused as part of a "revert" natured commit. It was previously applied to velocity before being made unused.

huwb:
I think it can be nuked, take. We should look at adding randomness to the dynamic waves so that they look more random and natural. This simple approach didn't really work

daleeidd:
Thanks. Unused code has been removed. I'll close this since I don't think anyone else will report it.

PR #592, opened 2020-07-18
Remove unused code

daleeidd:
Related to #587

I just removed what was reported to me as unused by code analyser. I wasn't sure if any of it should go the other way and be used rather than removed.

Let me know what you think.


huwb:
Love it! How did you analyze it exactly?

I think we have dummy variables in more places. The `out _` notation seems like it would help in a few places. I'll have a quick scan

huwb:
Ok the existing dummy's are pass by ref instead of using out. Thats a shame. I tried moving it to outs and i think its better, but its a more significant change. So i'll make a separate PR

daleeidd:
Thanks. I'm using vscode which uses omnisharp/roslyn to report this sort of stuff (and more).

PR #593, opened 2020-07-18
Use 'out' which allows some reduction around the sample helpers

huwb:
My 'out game' has not been on point in the past, some annoying cruft can be removed by moving the sampling helpers to use out params.

Let me know what you think


huwb:
Cheers, merged

PR #594, opened 2020-07-19
Fix ocean tiles frustum culling

### LINKED ISSUE #585: Ocean tiles culled at edges of frustum
**Describe the bug**
Ocean tiles are culled at edge of frustum when they shouldn't be.

**Screenshots / video**

**Versions**
4.4 and higher. Not sure if earlier. Looks like all pipelines. 

**To Reproduce**
In the boat.unity, weigh in the global spectrum to increase turbulence which will help with exposes culling issues.

**Platform**
 - Editor and Standalone
 - Windows and Mac

**Hardware**
i7 and GTX 1080Ti

**Additional context**
I disabled the culling done in _LateUpdateTiles_ and disabled occlusion culling in camera and on tile renderers just to rule them out, but the issue still appears. It might have something to do with the displacement of the tiles pushing tiles that would be culling into the frustum?

daleeidd:
Fixes #585 by fixing both the displacement reporting and reverting shared mesh usage.

Displacement reporting had shared state that was being overwritten which is now fixed. There are still some very minor issues with displacement reporting. I have to freeze time to get correct culling when using the frame debugger. It works fine otherwise. I think using delegates and making Crest more event oriented should help with this further.

And lastly there is moving away from shared meshes. We will lose instancing as a result, but culling is more important. Using sharedMesh in the editor is still correct from testing. The mesh bounds gizmos don't expand in edit mode, but using a mesh instance didn't appear to fix it.

I've also added a custom inspector to show the renderer bounds in the inspector.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/OceanChunkRenderer.cs
+++ b/crest/Assets/Crest/Crest/Scripts/OceanChunkRenderer.cs
@@ -5,6 +5,10 @@
 using UnityEngine;
 using UnityEngine.Rendering;
 
+#if UNITY_EDITOR
+using UnityEditor;
+#endif
+
 namespace Crest
 {
     /// <summary>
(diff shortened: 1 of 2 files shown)


daleeidd:
Thanks

PR #596, opened 2020-07-22
Support 2018.4 LTS

skaughtx0r:
This makes the latest master branch compatible with Unity 2018.4 LTS.

I'm not sure if this should go back into master. Basically the way I use Crest is by just copying the `crest/Assets/Crest` directory out of this repo. Using that method, these changes work fine.

The main issues I've found are the SRP version in the asmdef of the Development directory and the package manager manifest.json. I'm not sure how to make these compatible with both 2019 and 2018 (might not be possible).

If we merge these changes into the Crest directory, it may be easier to maintain the legacy/2018 branch with fewer conflicts?

I'm proposing this because I'm currently stuck on 2018.4 due to the Legacy InputSystem, but I'd like to keep tracking the latest version of Crest mainly because I'm still using the unmerged underwater postprocessing branch/PR. I also needed to pull in the fixes for the FixedUpdate water queries.


huwb:
After some consideration I think we should probably use the legacy branch for this job and not put 2018 stuff back into master. There are other opportunities we can take advantage of by dropping 2018 support in master that we hope to pursue in the future, and we don't want to manage ifdef'd code and logic in future changes/refactors/fixes for a version we're not testing. Let me know what you think..

daleeidd:
> but I'd like to keep tracking the latest version of Crest mainly because I'm still using the unmerged underwater postprocessing branch/PR

Would having that branch merged into master help? There is a bit if a push to have this done. I am keeping this branch more up to date lately.

> The main issues I've found are the SRP version in the asmdef of the Development directory and the package manager manifest.json. I'm not sure how to make these compatible with both 2019 and 2018 (might not be possible).

You can exclude that directory when you add it to your project since it is only useful to us.

skaughtx0r:
I agree this would probably best be targeted to the legacy/2018 branch.

@daleeidd For me personally, yes, getting the underwater postprocessing branch merged into master would help. I know there's some outstanding issues there, I commented with a possible path forward to integrate with the PostProcessing Stack that I'm not sure you guys have looked into.

Currently I only pull the Crest directory into my project, so I'm able to get everything working with the handful of #ifdefs I have in this PR.

huwb:
Thanks @skaughtx0r !

This is fairly tricky to review, due to the nature of it, but i think its a thumbs up from me. Lets give @daleeidd / @moosichu a day to give feedback.

daleeidd:
There is still a few shader_feature_local pragmas in other shaders like the underwater curtain. I didn't notice anything else wrong.

skaughtx0r:
Oh, whoops, I missed those because I'm actually using another branch based on this that has the underwater postprocessing branch merged into it.

skaughtx0r:
I just merged latest master into this branch and addressed the last comment about fixing the `shader_feature_local` keywords in the underwater curtain. Looks like they were changed to `multi_compile_local` in master. I just dropped the "local" part and it seems to work in 2018. This should be ready for review again if you want to merge it.

skaughtx0r:
@huwb This PR can be declined, I should have my project updated to Unity 2019 soon, so I won't be maintaining this branch anymore.

huwb:
Thanks for the note, will close

PR #599, opened 2020-07-24
Fix underwater safety margin edge cases

### LINKED ISSUE #520: [HDRP] Gap between waves and underwater area.

daleeidd:
Hopefully fixes #520

Solves a handful of edge cases for the underwater horizon.

### Edge case where camera is +/-90° roll

The horizon normal would sometimes be inverted when camera has +/-90° roll which may have been due to precision errors. I have added a fuzzy check to catch this.

### Safety margin didn't handle camera roll

The margin was added the horizon position up. But up will be along the normal when the camera is rolled. This result in the multiplier either having no effect or the opposite effect. I just used then normal to take roll into account.

### Safety margin could be too large

The camera position to sea level was being used for the safety margin which lead to it sometimes covering a quarter of the screen. I have changed it to only use the sign so the margin's size is determined by the multiplier property. This fixed a lot of edge cases where checks to disable the margin failed.

### Fix Single Pass XR

The ocean surface on the right eye is slightly higher than the left, leaving a gap and the safety margin falling short. Looks like we had to set the projection matrix on the camera.

## Testing

I know how to reproduce on my machine. An ocean with no waves and:

```
Camera Position
X = 0
Y = -0.2
Z = 0.16

Camera Rotation
X = -4.68
Y = 0
Z = 0
```

In the game view, there is a dropdown where you can add your own screen dimension presets. Create one with a height of 684. Width doesn't appear to make a difference. Then set that preset.

I've added a debug option to disable camera kill roll.

### DIFF
--- a/crest/Assets/Crest/Crest-Examples/Shared/Scripts/CamController.cs
+++ b/crest/Assets/Crest/Crest-Examples/Shared/Scripts/CamController.cs
@@ -23,6 +23,9 @@ public class CamController : MonoBehaviour
     [System.Serializable]
     class DebugFields
     {
+        [Tooltip("Disables controller preventing the camera from rolling (rotating on the z axis).")]
+        public bool disableCameraRollPrevention = false;
+
         [Tooltip("Disables the XR occlusion mesh for debugging purposes. Only works with legacy XR.")]
         public bool disableOcclusionMesh = false;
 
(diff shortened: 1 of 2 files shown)


daleeidd:
Thanks

PR #602, opened 2020-08-02
Fix caustics shadows projected position misalignment

daleeidd:
The shadows are projected after being sampled so they align with the shadows on the scene. But it is currently using the projected position which includes a fudge factor. This calculates the projected position without the fudge factor.

The fudge factor does the following:
> project along light dir, but multiply by a fudge factor reduce the angle bit - compensates for fact that in real life caustics come from many directions and don't exhibit such a strong directonality

Before

After

### DIFF
--- a/crest/Assets/Crest/Crest/Shaders/OceanEmission.hlsl
+++ b/crest/Assets/Crest/Crest/Shaders/OceanEmission.hlsl
@@ -155,6 +155,8 @@ void ApplyCaustics(in const half3 i_view, in const half3 i_lightDir, in const fl
 
 #if _SHADOWS_ON
 	{
+		// Calculate projected position again as we do not want the fudge factor.
+		const float2 surfacePosXZ = scenePos.xz + i_lightDir.xz * sceneDepth / i_lightDir.y;
 		half2 causticShadow = 0.0;
 		// As per the comment for the underwater code in ScatterColour,
 		// LOD_1 data can be missing when underwater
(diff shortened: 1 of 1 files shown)


huwb:
I guess that code was mine, probably should have written more comments..

It does look wrong so probably needs to be removed.

Did you try just removing the fudge factor, rather than making a new position var?

daleeidd:
> Did you try just removing the fudge factor, rather than making a new position var?

I just tried it. Without the fudge factor, the caustics move around with the waves more. Here is a video with and without the fudge factor applied respectively (you'll notice the change).

https://drive.google.com/file/d/13MJH7VgplYIEWZPJNi1wquaQqjblvX97/view?usp=sharing

huwb:
That seems better I guess?

daleeidd:
With the fudge factor removed completely, I notice that the caustics stretch and distort too much in certain areas. I think it is probably best to leave it untouched for now.


Stretching in the middle. Can get worse than this

huwb:
Ok, still a good fix. do you mind adding comments for what you found?

daleeidd:
> do you mind adding comments for what you found?

Done

huwb:
Thanks. Sorry 1 more thing - is it worth giving the two position variables different names to distinguish them? I'm thinking that relying on shadowing could be error prone ("I'm changing the variable but it's having no effect"), and could conceivably generate warnings on some platforms (maybe).

daleeidd:
Good idea. Thanks. All done and merged.

PR #610, opened 2020-08-09
Refactor - remove dispatch logic from PropertyWrapperCompute

huwb:
**Status:** Ready to merge

Removes dispatch logic from PropertyWrapperCompute.

Few reasons for this:

- Brings in line with other wrappers which serve only to wrap
- The dispatch is a single line of code and can be done outside
- Resetting state after dispatch means wrapper needs to be re-initd
- Thread group size should in general vary based on shader/workload,
so should not be baked into wrapper

I bumped into these a couple of times on dev branches so thought i should
submit this pr.

Oh i also thought we should move THREAD_GROUP_SIZE_X/Y into LodDataMgr
as i guess logically it can be considered a default value for lod-data related things
(and is paired with constants in OceanLodData.hlsl..).


huwb:
Thanks

PR #611, opened 2020-08-09
Update displacement correction policy for ocean inputs

huwb:
I had a nasty case where i was adding an anim waves input which had
displacements, and since the correction was on by default, there was
a feedback loop where the displacement in the input moved the input
itself around. This disables correction on anim waves inputs by default.
I looked for a fancier way to do this (rather than duping some code)
but settled on this.

Also the tooltip was wrong on the gui, fixed.


huwb:
Thanks

PR #615, opened 2020-08-12
Example height component using UnityEvents

daleeidd:
Adds a component that will invoke an event for when the game object changes state from above to below the ocean surface and vice versa.

It can be used to trigger audio and many other things. Pretty versatile and powerful for those who don't know scripting.


huwb:
Great shout, thumbs up from me!

We may want to extend it with an optional velocity threshold, that would be good for triggering particle FX. But even if that does make sense, that could be a second/later step after this is merged. I can try to move that splash FX branch onto this system later.

daleeidd:
Thanks.

After merge makes sense. It will be good to evaluate the branch against this feature as that branch might not be simple enough for UnityEvents. Using velocity will need to be a separate event I think and the particle FX branch might make it redundant on here.

I added another event for the distance from ocean surface. It can be plugged into the audio volume. I think these two events should satisfy a lot of audio needs.

PR #618, opened 2020-08-15
OceanLODData.hlsl removed

huwb:
Goodbye OceanLODData.hlsl. You won't be missed!

Removes helper functions which relied implicitly on hidden state, and added unnecessary concepts such as the 'bigger'/'smaller' naming. A downside of this is some fairly verbose calling code. A potential fix for this is grouping args in structs.

I think there is more cleanup to follow (OceanHelpersNew -> OceanHelpers?), but thought i'd stop at this point as thats more than enough change for one PR..


huwb:
Ok perfect thanks! Merging

PR #619, opened 2020-08-16
Add support for adding signed distance field to shoreline in depth cache

moosichu:
Cherry picked from #515.

This PR adds the tooling used to add a signed distance field inside a depth cache by checking a box at the bottom of the component:

image

The texture that is generated is a two-channel texture, with depth and signed distance field in the red and green channels respectively.

As a standalone change - this is pretty useless, but you can test it's effects in the shoreline branch.

Whilst the shorelines waves are still in active-development, this stuff is pretty stable now and good to be subjected to scrutinity, so I thought it would be good to seperate-it-out into a seperate PR, so that we can merge this with little risk to the end-user, so the runtime can be evaluated seperately at a later date.

The actual signed distance field is generated using the JumpFlood algorithm and applying it by ping-pong-ing between two textures. It actually generates a voronoi diagram containing the positions of the nearest worldspace pixels that are above-water - which is then converted to a signed distance field in the SDFApplyShader.

https://www.comp.nus.edu.sg/~tants/jfa.html


crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/TerrainSignedDistanceField.shader
```diff
@@ -0,0 +1,58 @@
...
+
+// Renders ocean depth - signed distance from sea level to sea floor
+Shader "Crest/Inputs/Depth/Initialise Signed Distance Field From Geometry"
+{
+	SubShader
+	{
+		Pass
+		{
+			// TODO(TRC):Now figure-out how to handle this blend operation properly!
+			// BlendOp Min
```
moosichu, line 13: Just realised I need to fix this! Will do it soon.

moosichu: Current plan to do this is to have one shader for rendering the depths that blends using this - and then having an init shader handle this.

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/InjectorSignedDistanceField.shader
```diff
@@ -0,0 +1,54 @@
...
+			Varyings Vert(Attributes input)
+			{
+				Varyings output;
+				output.positionCS = UnityObjectToClipPos(input.positionOS);
+
+				float3 worldPos = mul(unity_ObjectToWorld, float4(input.positionOS, 1.0));
+
+				output.depth = _OceanCenterPosWorld.y - worldPos.y;
+				output.worldPosXZ = worldPos.xz;
+
```
huwb, line 41: feel bad even pointing this out but theres a double space here, is this intentional? generally the code looks great though!

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/InjectorSignedDistanceField.shader
```diff
@@ -0,0 +1,54 @@
+// Crest Ocean System
+
+// This file is subject to the MIT License as seen in the root of this folder structure (LICENSE)
+
+// Renders signed distance field without any depth
```
huwb, line 5: Could you expand on this if possible? I'm looking at the code and not quite understanding how it works/what it does. I think the 'depth' value being unused is throwing me

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/InjectorSignedDistanceField.shader
```diff
@@ -0,0 +1,54 @@
...
+
+			struct Attributes
+			{
+				float3 positionOS : POSITION;
+			};
+
+			struct Varyings
+			{
+				float4 positionCS : SV_POSITION;
+				float depth : TEXCOORD0;
```
huwb, line 28: i think this is unused, is that intentional?

crest/Assets/Crest/Crest/Shaders/SdfShared.hlsl
```diff
@@ -0,0 +1,20 @@
+// Crest Ocean System
+
+// This file is subject to the MIT License as seen in the root of this folder structure (LICENSE)
+
+// Setting this to 0.0 means that gometry at exactly the origin won't be handled
```
huwb, line 5: just to understand, what will the artifact be? is this only if the texel center is exactly at the origin?

it seems like a moderately possible scenario, but if the impact is low then its fine. i was trying to think of a better magic number, but didnt come up with anything (safe) so far!

huwb:
A couple more dumb questions occurred to me when thinking about the wider context of how this fits into the existing system and what kinds of things users are likely to run into. (the answers would probably become clear from the rest of the code, but recording so i dont forget)

* Is the jump flood intended to run offline, or is it supported at runtime? If it's offline only, do runtime/dynamic ocean depth inputs still function as usual (probably yes but just checking!).
* Can multiple depth caches be overlapped / will multiple sdfs blend? Think we already discussed this but i forgot, apologies.
* Dynamic depth inputs (objects tagged with RegisterOceanDepthInput component and `Crest/Inputs/Depth/Ocean Depth From Geometry` shader assigned) writes this into the sea floor depth loddata `float4(input.depth, 0.0, 0.0, 0.0);`. I'm guessing this would stomp over the distances - we could set a WriteMask R to only write the first channel if that is the correct solution.

It would be great to test these wider integration bits if not tested already to make sure it doesn't conflict with the current systems. Hope this makes sense, and thanks again for this, cant wait to get my hands on the sdf!

crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
```diff
@@ -77,11 +78,25 @@ public enum OceanDepthCacheRefreshMode
         bool _runValidationOnStart = true;
 #pragma warning restore 414
 
-        RenderTexture _cacheTexture;
-        public RenderTexture CacheTexture => _cacheTexture;
+        [Tooltip("Generate a signed distance field for shorelines"), SerializeField]
```
daleeidd, line 81: The SDF options could go under a heading to keep it seperate from the other options, and adding something that makes it clear to users that this feature is not currently being used or that it is an upcoming feature?

moosichu:
>  I think there is code missing such as the changes to the ocean depth cache component, and the texture format change in the sea floor depth lod data?

I didn't realise you wanted that part as well - will bring it in!

Thanks for the good feedback. Will jump on acting on it and will post when ready for the next round. :)

moosichu:
> Question - what will happen if a texel lands on the 0,0 in worldspace? 

It will be treated as an initialised value when the depth cache is generated, so one pixel in the depth cache with landmass on it will be treated as open ocean. I highly doubt that would be noticeable at all.

moosichu:
Warnign should be fixed now, although I found it very difficult to trigger even with a reimport, so let me know if it's still there. If not, please feel free to merge on my behalf as I'm quite busy today.

huwb:
> I highly doubt that would be noticeable at all

I guess this assumed that this is under landmass, but the origin could be anywhere from open water to under landmass and any point in between? 

I did some more googling. How about we use this value: 65504.0

If my source (wikipedia) was correct, this is the largest number that can be put in a half. This moves it far away from the origin.

crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
```diff
@@ -77,11 +78,26 @@ public enum OceanDepthCacheRefreshMode
...
-        RenderTexture _cacheTexture;
-        public RenderTexture CacheTexture => _cacheTexture;
+        [Header("Signed Distance Field (Experimental)")]
+        [Tooltip("Generate a signed distance field (experimental)"), SerializeField]
+        internal bool _generateSDF = false;
 
-        GameObject _drawCacheQuad;
-        Camera _camDepthCache;
+        [PredicatedField("_generateSDF")]
+        [Tooltip("How many additional Jump Flood Algorithm rounds to use - (over the standard log2(Resolution)"), SerializeField]
```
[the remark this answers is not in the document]
moosichu: Yes it does - but jumpflood isn't technically an accurate algorithm - it's result is an approximiation that trends towards the actual SDF. Additional rounds of JFA can help correct for some of these innacuracies. It's described in the paper. One sec...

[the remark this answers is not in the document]
moosichu: > We  observe  in  our  experiments  that  most  grid  points  with  errors are grid points clustered around Voronoi vertices, or around Voronoi  edges  intersecting  the  boundary  of  the  grid.  Moreover,  many  grid  points  with  errors  are  each  a  single  error  and  itself  a  Voronoi  vertex.  As  such,  a  simple  yet  effective  way  to  remove  many  errors  is  to  execute  JFA  for  one  additional  round  with  step  length  of  1,  i.e.  a  total  of  (logn)+1  rounds  where  both  of  the  last  two rounds are with step length of 1. We call this variant JFA+1.   It is easy to realize that one can then use JFA+2 which is the usual JFA plus two additional rounds, with step length of 2 and 1 respectively.  Likewise,  one  can  do  as  many  additional  rounds  as  needed  to  achieve  good  accuracy  (while  converging  to  no  better  than  the  standard  flood).  We  have  also  used  JFA  plus  lognadditional rounds, i.e. JFA+logn or simply JFA2. Our experiments as reported in Section 6 confirm the effectiveness of these variants. In  particular,  JFA2  produces  correct  results  in  most  cases,  though  it  fails  for  the  difficult  case  as  shown  in  Figure  5.  (This  is  also  a  case  where  [Borgefors1984]  cannot  correctly  handle  with  a  3×3  mask.) In this case, if the center grid point p is incorrect at the end of JFA, it will still remain incorrect at the end of JFA2. This is so as the Voronoi cell (shown as a wedge) does not “cover” any grid points along the dotted lines that p refers to during the additional logn  rounds,  i.e.  p  thus  never  receives  the  correct  seed  from  any  grid points of its Voronoi cell. 

https://www.comp.nus.edu.sg/~tants/jfa/i3d06.pdf

crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgrSeaFloorDepth.cs
```diff
@@ -7,23 +7,41 @@
...
+        public override RenderTextureFormat TextureFormat { get { return Settings._enableSignedDistanceFields ? RenderTextureFormat.RGHalf : RenderTextureFormat.RHalf; } }
         protected override bool NeedToReadWriteTextureData { get { return false; } }
 
         bool _targetsClear = false;
 
         public const string ShaderName = "Crest/Inputs/Depth/Cached Depths";
 
         // We want the null colour to be the depth where wave attenuation begins (1000 metres)
-        readonly static Color s_nullColor = Color.red * 1000f;
+        readonly static Color s_nullColor = new Color(1000f, 100f, 0f, 0f);
```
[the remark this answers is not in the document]
moosichu: No - that's a typo... it should also be 1000f.. was just matching what we do for depth. Happy for you do to this but will at least change it to 1000f for now.

moosichu:
> Another option could be Nan?

I was thinking that as well - unfortunately creating NaN's appears to be non-trivial. However inf seems to have worked!

huwb:
Looks great to me nice one!

Could you capture that info about extra jump flood steps maybe in the code doc - maybe just a brief mention that it helps correct errors and is described in the paper world be enough as a bread crumb trail for future reference, if that makes sense?

crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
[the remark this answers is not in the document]
huwb: That's really interesting thanks

PR #626, opened 2020-08-22
New foam texture

huwb:
This enables the new foam texture in all scenes. Texture was already pushed, this turns it on.

I think this is simply better and is good to go, but wanted to get your thoughts to make sure i didnt miss anything.

### DIFF
--- a/crest/Assets/Crest/Crest-Examples/Whirlpool/Data/WavesWhirlpool.asset
+++ b/crest/Assets/Crest/Crest-Examples/Whirlpool/Data/WavesWhirlpool.asset
@@ -64,5 +64,6 @@ MonoBehaviour:
   - 1
   - 1
   - 1
-  _chop: 1.5
+  _chop: 1.3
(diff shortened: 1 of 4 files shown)


huwb:
I don't think it has worked as well as i hoped. Let me know how you guys are finding it.

It fills up to nearly 100% foam in some cases. I think finessing the foam input strength helps here and is probably a good thing to do in general. I did a quick test and it helped for shoreline foam. I'll see if i can get better results everywhere with that, or will revert back to the old texture if need be.

daleeidd:
I prefer it. It does exacerbate some issues depending on the foam settings like when foam pools under the camera but looks bare everywhere else (not sure if this is just a Mac issue or not). But overall it does look a lot better. It is especially nice in rougher waters.

huwb:
Thanks. Fine tuning the shoreline foam strength made it look a fair bit better so it probably just needs more finessing

Issue #392, opened 2020-01-09
Reflections bleeding through walls in darkness


Kandob:
I've made a room with complete darkness and thick walls but it seems like the reflection from outside is bleeding through. Also, the shadowing doesn't make the ocean completely dark. Is this a bug or am I doing something wrong?

![Annotation 2020-01-09 222444](https://user-images.githubusercontent.com/14958256/72106206-38581e80-332f-11ea-8f18-0e4c80333769.png)

huwb:
Hi there, i'm not sure about this one. I believe this would also be an issue for any reflective material (besides water). I don't have enough experience to know how this kind of thing is solved. Maybe something with reflection probe settings?

Kandob:
My guess is that the specular is not being masked in shadowed areas. I can see some fresnel at certain angles which shouldn't be visible in complete darkness.

Is the fact that the ocean can't be completely dark a limitation in unity or is it something else?

huwb:
I think its something with the reflection probe settings - but I'm not an expert at lighting and have never tackled a tricky case like this, so I may be limited help.

Kandob:
The thing is, i'm not using any reflection probes.

huwb:
Ok then. In that case you might try reducing the jitter / blur radius of the shadow data, on the shadow settings:

![image](https://user-images.githubusercontent.com/939901/72217090-a2d0b080-3521-11ea-9916-ed11dae2bb42.png)

Kandob:
I've tried it but it has no effect on the reflections

b1gry4n:
Another thing to keep in mind is that the reflection camera is actually another camera, any image effects youre applying to your main camera arent taken in to consideration when rendering the reflection. So for example if you are darkening your colors in the main camera its possible the sky/objects are still bright or brighter to the reflection camera.

Kandob:
No that's not he issue here, the planar reflection script that comes with crest uses the camera it's attached to.
https://github.com/crest-ocean/crest/blob/e74e5e5ad630820658b5ffe2aa73f4477887e9c9/crest/Assets/Crest/Crest/Scripts/Reflection/OceanPlanarReflection.cs#L107

b1gry4n:
youre right, I must have confused crest with other water assets. my bad

daleeidd:
@Kandob Can you please provide a minimal scene/project that reproduces this? We won't be able to determine the issue or improve Crest for this use case without having your setup.

daleeidd:
[...]

PR #631, opened 2020-08-25
Add shadow input component

daleeidd:
Adds a shadow input component similar to other LOD data. It could be useful for those who may have edge cases with shadows. The one advantage is complete control as it bypasses jitter.

An example use case is issue #392 where a small enclosed room can appear not shadowed inside because jitter samples outside the walls. 

Fixing the problem by reducing shadow leaks probably won't work because the fixes so far use depth.

### DIFF
--- a/crest/Assets/Crest/Crest/Materials/OceanInputs/ShadowOverride.mat
+++ b/crest/Assets/Crest/Crest/Materials/OceanInputs/ShadowOverride.mat
@@ -1,0 +1,80 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!21 &2100000
+Material:
+  serializedVersion: 6
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  m_Name: ShadowOverride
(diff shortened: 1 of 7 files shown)


huwb:
Yeah this makes sense, looks good to me!

(I actually thought the input object would cast a shadow onto the plane (rather than replace shadow data contents). That also sounds quite useful. But i guess it would have to be rendered from the lights POV which is more complicated. And what happens for point light shadows and stuff.)

daleeidd:
Thank you.

daleeidd:
@huwb So I forgot to test this for VR/XR and it doesn't work with it. It appears that we cannot override the camera matrices for the _DrawRenderer_ command when rendering in stereo.

I've tried adding a second command buffer to render during the _CameraEvent.AfterEverything_, but the data added by the input flickers when applied to the ocean surface (doesn't appear to flicker in the render texture). Otherwise, it seemed to work fine.

Another workaround is to use *Graphics.ExecuteCommandBuffer* to take the pass outside of the camera render queue. But this would add a delay of one frame to the shadow data.

Otherwise, another option is a separate command buffer which renders the shadow input objects to a temporary texture and then blends into the shadow data afterwards.

daleeidd:
All fixed. I have it working for all pipelines. Only single pass double-wide was an issue and the solution was simple enough.

PR #633, opened 2020-08-25
Minor refactor of update shadow

daleeidd:
UpdateShadow now uses the shadow term constants from downstream. This works towards making it easier to remove the hard shadow term in HDRP.


huwb:
Looks good!

daleeidd:
Thanks!

Issue #624, opened 2020-08-21
Bad performance on PS4


woistjadefox:
**Describe the bug**
Dear Crest community, I've ran a few performance tests today (main scene) on the major consoles and found out that Crest performs quite bad on the PS4. Even worse than on the Switch. Tested a few different versions, URP and HDRP and the verdict stays the same (check the screenshot). I remember having issues to implement async gpu readback on PS4 in our last game, it was super slow and I had to split my requests over frames into multiple readbacks/requests. This happened only on the PS4 and according to the internal PS4 forum, it has something to do with the memory access management of the device. Unfortunately I'm not allowed to give here more insights due to NDA.  I wonder if something similar could be the case here and if there is any way to get a better performance on PS4.

**Screenshots / video**
![image](https://user-images.githubusercontent.com/9056181/90875222-20567c00-e3a1-11ea-8fbd-b9920f8bba5d.png)

**Additional notes**
You can get all the profiler logs here: 
[crest-profiler-data.zip](http://goga.ch/r/crest-profiler-data.zip)

huwb:
Thanks for the detailed numbers.

Could you please provide some more details about the ocean setup. The following would be ideal:

* Which features you have running on the ocean component (dyn waves, foam, ...)
* How many gerstner inputs running
* How many ocean inputs in the scene (depth caches, any thing else..)
* What the *Lod Count* is set to on the ocean component

These numbers are much larger than i would expect. It would be interesting to know how it performs on Release (on ps4 only would be fine). We take Development numbers with a big grain of salt, but even with the salt this is very high cost.

Just to check - it's not spewing errors to the log is it? That would increase the cost a bunch.

Could you try disabling *Evaluate Spectrum At Runtime* on the ShapeGerstnerBatched component? We evaluate the spectrum each frame by default in case it changes, and this work is nontrivial (especially if there is more than one gerstner script).

A good next step would be to add more profiling tags to pin down where the cost is coming from. Adding them to each type of LodData/feature could be a good step. Having said that `OceanChunkRenderer.OnWillRenderObject()` is self contained, and really does not do very much, so its disturbing to see such a high cost attributed to it.

OkoGoran:
> Could you please provide some more details about the ocean setup.

The ocean setup is the default setup from Crest's main scene (also the default ocean material which comes with the URP / HDRP package from the asset store). 
[...]

PR #634, opened 2020-08-26
Remove virtually all work from OceanChunkRenderer

huwb:
Massively WIP - not working atm

Factors out work that was repeated a lot. For example, cascade params (transforms etc) were being set by each loddata, for each tile, when it can be set just once globally for all tiles.

Driven by #624 

Opening early to collect any feedback


huwb:
Closing as this is superseded by #639

PR #637, opened 2020-08-28
Add XR SPI support to underwater post processing

daleeidd:
Add changes to support XR SPI. It brings the XR helper class from downstream too to unify codebase.

The helper will probably require some changes if we switch to the post processing stack. Same goes with using the helper in URP. I am uncertain if the post processing stack will make things easier to worse for support XR.


crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterPostProcess.cs
```diff
@@ -145,14 +145,16 @@ private void ViewerLessThan2mAboveWater(OceanRenderer ocean)
...
+
             // Allocate planes only once
             if (_cameraFrustumPlanes == null)
             {
                 _cameraFrustumPlanes = GeometryUtility.CalculateFrustumPlanes(_mainCamera);
                 _maskCommandBuffer = new CommandBuffer();
                 _maskCommandBuffer.name = "Ocean Mask Command Buffer";
                 _mainCamera.AddCommandBuffer(
-                    CameraEvent.BeforeForwardAlpha,
+                    CameraEvent.AfterForwardAlpha, // BeforeForwardAlpha breaks XR SPI right eye completely
```
daleeidd, line 157: This is probably the only change that might be an issue. If we use BeforeForwardAlpha, the right eye for VR/XR SPI no longer renders the ocean. I am guessing this is a matrix issue.

Even though it fixes it for the ocean, would it cause issues for anything after this in the pipeline? Not sure. But it is contained to XR SPI since everything else works fine.

So the question is should I make this conditional on whether VR/XR SPI is enabled?

huwb: So this is the mask rendering? My guess is that after forward alpha would work, if the ocean surface render itself does not make use of the mask. But @moosichu may be better placed to confirm.

daleeidd: It is. It appears to work. I opened an issue with Unity so hopefully it will be fixed and we can always move it back:
https://issuetracker.unity3d.com/product/unity/issues/guid/1273307

moosichu: The Ocean Renderer *does* make use of the mask, and this branch depends on behaviour from it even more heavily:

https://github.com/crest-ocean/crest/pull/462

moosichu: I think we should make this conditional on whether the VR/XR SPI is enabled, alongside a comment explaining that it's needed due, but does lead to a reduction in potentially supported features.

daleeidd: @moosichu Thanks. I tried again to solve the issue, but uncertain as to the exact cause. I have made it conditional and added comment. I might need to open an issue with Unity.

daleeidd: I've submitted a bug report. Is there a reason why we aren't rendering the mask outside of the camera's render queue? In HDRP the mask is rendered between the LOD data and the camera. It might be an alternative if Unity decides not to fix it.

daleeidd: Looks like it will be fixed in 2020.2. Hopefully they will bring the fix to 2019.4 too.

https://issuetracker.unity3d.com/product/unity/issues/guid/1261545

moosichu:
As I'm working on porting post-process effect to the post-process stack - it might be worht holding-off on any post-process BIRP work until then as it changes quite a lot unfortunately :(

daleeidd:
No worries. I'll leave this for now.

PR #638, opened 2020-08-28
Spectrum Driven Small Details

magenta404:
I have attempted to drive small details strength with values from the ocean wave spectrum. I am not sure I have done it correctly, but it appears to be working. Small details is covered by normals and roughness.

I used the amplitude values. Not sure if that was the best way to go.

For the "enabled" and "disabled" screenshots, the roughness was set to one and zero respectively. The base normal scale needs to be improved and can look strange when the scale is low enough to be visible but too small to look good. I am thinking  that driving the scale and strength might work so we ensure normals always look nice when visible.

Let me know what you think.

### Main Scene without changes

1_1_Enabled
Enabled

1_2_Disabled
Disabled

### Main Scene with low spectrum weight

2_1_Disabled
Enabled

2_2_Disabled
Disabled

### Main Scene with zero spectrum weight

3_1_Enabled
Enabled

3_2_Disabled
Disabled

### Noticeable scale change

When weighing out a spectrum, the normal scale will decrease. When the camera has a high altitude, this is very noticeable. It looks almost like the normals scroll at fast speed.

### Local Gerstners

Not sure if we should try and handle this case.

### Multiple Gerstners

I think just a static list of global gerstners could work. And then we take the highest value.

## Testing

On the _OceanRenderer_, there are two toggles to enable/disable driven normals and roughness. I have also exposed properties on the _OceanChunkRenderer_ for debugging.

### DIFF
--- a/crest/Assets/Crest/Crest-Examples/Main/Scenes/main.unity
+++ b/crest/Assets/Crest/Crest-Examples/Main/Scenes/main.unity
@@ -38,7 +38,7 @@ RenderSettings:
   m_ReflectionIntensity: 1
   m_CustomReflection: {fileID: 0}
   m_Sun: {fileID: 0}
-  m_IndirectSpecularColor: {r: 0.40711796, g: 0.5924692, b: 0.9424595, a: 1}
+  m_IndirectSpecularColor: {r: 0.40657905, g: 0.5909717, b: 0.9444369, a: 1}
   m_UseRadianceAmbientProbe: 0
 --- !u!157 &4
 LightmapSettings:
@@ -54,7 +54,7 @@ LightmapSettings:
     m_EnableBakedLightmaps: 1
     m_EnableRealtimeLightmaps: 0
   m_LightmapEditorSettings:
-    serializedVersion: 10
+    serializedVersion: 12
     m_Resolution: 1
     m_BakeResolution: 50
     m_AtlasSize: 1024
@@ -62,6 +62,7 @@ LightmapSettings:
     m_AOMaxDistance: 1
     m_CompAOExponent: 0
     m_CompAOExponentDirect: 0
+    m_ExtractAmbientOcclusion: 0
     m_Padding: 2
     m_LightmapParameters: {fileID: 0}
     m_LightmapsBakeMode: 1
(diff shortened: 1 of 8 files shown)


daleeidd:
Added some debugging stuff and some possible improvements.

daleeidd:
Solved that one issue with LODs. Was a pretty obvious bug. Added a new issue about scale changing.

daleeidd:
Not ready to merge but ready for some feedback. It demonstrates the idea enough I think.

huwb:
Thanks for the bump it's next on my list to look at

huwb:
Thanks for this! Really interesting to see this coming through.

I've been playing around and i think i get the jist of it. I've realised it's quite tricky thing to make work correctly.

One basic problem is that the normal maps do not capture the appearance of the waves at all. In an ideal world, if you zoom in on a patch of water like the below, the only visual thing that changes is detail subtlely gets blended in.

image

Instead what happens is when the actual waves come in they can be very different from the normals (the waves are ramped up a bit too much here and trigger the foam like crazy, which is another potential problem).

image

I need to think about if the normals could be a good approximation to the underlying waves. It makes me think that should be possible to render a patch of waves that tiles horizontally, which could then be used as a normal map? Like what would happen with FFT. This is an entirely new rabbit hole / tangent but just thinking aloud..

If they cant match then maybe they are not driven by the spectrum and function the way they are now, leaving the roughness. I played with the spectrum and observed the roughness. I can see it changing but am struggling to decide if this is a good thing or if it should be left entirely to art.

Its possible i was failing at tweaking or focusing on the worst cases, apologies if thats the case. Let me know what you think.

moosichu:
I haven't looked at the code, so haven't seen if this is taken into account, but it's worth pointing out that the physics of small (capillary) waves is quite different than those of large (gravity) waves.

https://en.wikipedia.org/wiki/Capillary_wave

Mainly because the restoring force in the former is surface tensions, whilst in the latter it is gravity.

huwb:
Yeah that's a cool effect

> It makes me think that should be possible to render a patch of waves that tiles horizontally, which could then be used as a normal map? 

I decided to scratch the itch and try this out. I believe it does tile - at least the waves do. However it's pretty easy to break the tiling property. Going to write notes to capture for the future

* Only the waves in the Lod tile, not bigger wavelengths in larger lods, so it would need to be taken before the combine pass from the wave buffer. This is possible.
* Shallow water dampens waves and breaks tiling (water depth does not tile periodically). I think the shoreline depth could be taken into account later instead of during generation.. maybe..
* Alignment to shorelines breaks periodicity
* Flow probably breaks periodicity too and would need a rejig

I think it's possible to render a repeating patch of waves in a canonical frame and then transform them to achieve the above points. I think we'd probably work in this way if we were using fft as well. Needs more investigation though, will leave for now.

daleeidd:
> I think it's possible to render a repeating patch of waves in a canonical frame and then transform them to achieve the above points.

I have been thinking about this too but to solve another problem where small waves contribute to other data like foam. Because the smaller wavelengths aren't present in the distant LODs, foam generation can be noticeable more potent under the camera with certain settings. I wonder if we could also similarly tile a patch of smaller waves into the foam data?

huwb:
Yes i think that could help as well. I think it would be worth exploring moving to purely-repeating wave buffers to allow testing this stuff, at some point..

Tangentially related, i tried a hack to boost the foam - branch `test/increase-foam-in-dist`

PR #639, opened 2020-08-31
Upload ocean driven data via compute buffers

huwb:
**What / why**:

As ID'd in #624 theres a lot of repeated work in OceanChunkRenderer.OnWillRenderObject(). In fact a lot of this work doesnt need to be done per frame at all.

This change puts shader params related to cascades data and cascade meshes and puts them into a compute buffer which is uploaded once per frame. These are bound globally now, so any graphics shader will receive them automatically (compute still needs to be bound manually).

With this no per-frame binding needs to be done at all, except for the foam/dyn waves because the target texture depends on the simulation steps (ping/pong). This simplifies things that need the ocean data, like clip data input and render alpha on surface, which now receive the data automatically. This also means that any shadergraph will get this data automatically and brings us closer to allowing anyone to author a water shader and assign to any mesh.

**Tests**:

Example scenes in BIRP: Win64 editor, standalone. Unity 2019.4.1

@daleeidd if you get a moment could you please run this through the paces. It touches pretty much every system in core crest.

**Issues**:

I don't love that the compute buffer is read many times across the HLSL, rather than being read once and passed around. I think its probably worth fixing this as part of this PR.


daleeidd:
I did a pass and the only thing I could notice wrong was that shadow data pops when scaling vertically. [Video](https://drive.google.com/file/d/1blgV7LJ-7Q1NMxQt5xxbyI2eyBhog05w/view?usp=sharing)

huwb:
thanks! i looked into it and everything looks ok, at least in the texture array, and the code looks fine. need ot think about this..

huwb:
ok fixed the shadow pop.. i missed that the shadow does ping pong rendering, the debug gui was drawing the correct texture, but it was not binding the correct texture for rendering, so it was out of date 50% of the time.

huwb:
Ok I believe its ready to merge.

If it works on your end @daleeidd and you're happy I'll get it merged

.gitignore
```diff
@@ -26,6 +26,7 @@ crest/*.pidb
 crest/*.booproj
 crest/*.svd
 crest/*.pdb
+crest/.idea
```
moosichu, line 29: Are you using Rider now for Crest?

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
```diff
@@ -319,6 +358,15 @@ void OnEnable()
             Instance = this;
             Scale = Mathf.Clamp(Scale, _minScale, _maxScale);
 
+            _bufPerCascadeInstanceData = new ComputeBuffer(_perCascadeInstanceData.Length, Marshal.SizeOf<PerCascadeInstanceData>());
+            Shader.SetGlobalBuffer("_PerCascadeInstanceData", _bufPerCascadeInstanceData);
+
+            _bufCascadeDataTgt = new ComputeBuffer(_cascadeParamsTgt.Length, Marshal.SizeOf<CascadeParams>());
```
moosichu, line 364: UnsafeUtilty.SizeOf is the preferred way of getting the size of structs in Unity. You should avoid using Marshal.

https://docs.unity3d.com/ScriptReference/Unity.Collections.LowLevel.Unsafe.UnsafeUtility.SizeOf.html

The Marshal functions are slower, generate garbage, and the UnsafeUtility functions optimise better when compiled to IL2CPP.

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
```diff
@@ -286,6 +290,41 @@ public enum DefaultClippingState
...
+            public float _farNormalsWeight;
+            public float _geoGridWidth;
+            public Vector2 _normalScrollSpeeds;
+
+            // Align to 32 bytes
+            public Vector3 __padding;
+        }
+        public ComputeBuffer _bufPerCascadeInstanceData;
+
+        CascadeParams[] _cascadeParamsSrc = new CascadeParams[LodDataMgr.MAX_LOD_COUNT + 1];
```
moosichu, line 323: Minor point, but something that might be worth breaking-out into a wider discussion. Now that we are on 2019.4 it might be better to default to using `NativeArray<T>` instead of C# managed arrays. Unity recommend using the former where possible because even if you don't use burst they can have slightly better perf characteristics - and if we do ever decide to burstify our code - this would make doing so easier.

crest/Assets/Crest/Crest/Shaders/Ocean.shader
```diff
@@ -308,15 +309,15 @@ Shader "Crest/Ocean"
 				// Sample shape textures - always lerp between 2 LOD scales, so sample two textures
 
 				// Calculate sample weights. params.z allows shape to be faded out (used on last lod to support pop-less scale transitions)
-				const float wt_smallerLod = (1. - lodAlpha) * _LD_Params[_LD_SliceIndex].z;
-				const float wt_biggerLod = (1. - wt_smallerLod) * _LD_Params[_LD_SliceIndex + 1].z;
+				const float wt_smallerLod = (1. - lodAlpha) * _CascadeData[_LD_SliceIndex]._weight;
```
moosichu, line 312: Will shaders optimise the fact we are sampling the same compute buffer over-and-over again into a single access?

crest/Assets/Crest/Crest/Shaders/OceanInputsDriven.hlsl
```diff
@@ -8,28 +8,35 @@
...
-
-// _LD_Params: float4(world texel size, texture resolution, shape weight multiplier, 1 / texture resolution)
-float4 _LD_Params[MAX_LOD_COUNT + 1];
-float3 _LD_Pos_Scale[MAX_LOD_COUNT + 1];
 uint _LD_SliceIndex;
-float4 _LD_Params_Source[MAX_LOD_COUNT + 1];
-float3 _LD_Pos_Scale_Source[MAX_LOD_COUNT + 1];
 CBUFFER_END
 
+struct CascadeParams
```
moosichu, line 14: We should add a comment to these structs with a tag linking them back to their C# counterparts to make it very clear that they must be identical. (and vice-versa in C# land)

moosichu (changes requested):
I really like how this change reduced LOC count which is always nice to see 👍 

I do think the uses of Marshal should be replaced though.

.gitignore
huwb: Not generally, this was an outlier.

crest/Assets/Crest/Crest/Shaders/Ocean.shader
huwb: Thanks yeah i forgot about that and am hoping to have a stab at this this weekend. I'm not totally sure but i could imagine it is not optimal to do this. I could check the compiled code to say for sure but i think factoring it out is a good thing in general so that the functions dont rely implicitly on these driven values..

crest/Assets/Crest/Crest/Shaders/OceanInputsDriven.hlsl
huwb: Will do

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
huwb: Ok thx will change it later

crest/Assets/Crest/Crest/Shaders/OceanInputsDriven.hlsl
huwb: Done, hopefully this is sufficient

crest/Assets/Crest/Crest/Shaders/Ocean.shader
huwb: I mostly did it. theres a variety of different cases (some sample with different array indices and etc). theres a few notable cases i left in ocean helpers. i think there might be more consolidation we need to do. also things may be a little different in the RPs in terms of how best to pass data around, so i'd propose settling it downstream before going further.

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
huwb: I could not seem to use it, the namespace seemed to be there but the utility UnsafeUtility did not seem to resolve for me. Would it be possible to get an assist with this?

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
huwb: I wont manage to do it for this PR if thats ok but i'm fine with it in principle if we're ok diverging further from 2018 which we probably are?

huwb:
Thanks i think i addressed most of the feedback, with the notable exception of UnsafeUtility, if i could get your assist with that @moosichu that would be appreciated. thanks!

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
moosichu: Done :)

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
moosichu: Sounds like a good idea.

moosichu (approved):

huwb:
Merged in latest and tested and it seems fine. Merging.

Issue #621, opened 2020-08-18
Check if the created texture in LodDataMgr is actually working on the GPU


klauskobald:
As we know, there is issues on Android which can be maybe solved easily by testing, if a texture is throwing errors. I think these errors could occur on all platforms as well depending on the GPU. 

SystemInfo.SupportsRenderTextureFormat is not enough to check!
So,
I am currently trying to implement a texture-format probe system that iterates thru a replacement list for a given format.

I create a texture like so:
var desc = new RenderTextureDescriptor(64, 64, rtf, 0);
[…]
result.enableRandomWrite = true;
result.Create();

BUT….

it seems that result.create() does not fail, even when readwrite is not supported.
Maybe a call has to be made to the GPU? How would I do that? My suspicion is to actually have commands sent to the GPU and wait for the result? The LodDataMgr would have then to wait before going into operation.

A second approach I could try is this: 
All LodDataMgr register with a textureProvider. If an error is happening later, then analyse the exception in the callback Application.logMessageReceived and change to the next replacement texture format and replace texture in LodDataMgr.

daleeidd:
https://docs.unity3d.com/ScriptReference/Experimental.Rendering.FormatUsage.LoadStore.html

Huw mentioned this on the other issue. It's experimental. Do you think it is worth trying?

klauskobald:
I found this:
SystemInfo.GetCompatibleFormat(gf, FormatUsage.LoadStore)
where gf is a GraphicsFormat. And return value is again a GraphicsFormat.
But how would I convert a RenderTexture into a graphicformat?

daleeidd:
The following utility should help:
GraphicsFormatUtility.GetGraphicsFormat()
GraphicsFormatUtility.GetRenderTextureFormat()

klauskobald:
I think we are getting somewhere!
I added this function to LodDataMgr and used it everywhere instead of TextureFormat.
<pre>
protected RenderTextureFormat GetValidatedTextureFormat()
{
    var gf = GraphicsFormatUtility.GetGraphicsFormat(TextureFormat, true);
    var cgf = NeedToReadWriteTextureData ? SystemInfo.GetCompatibleFormat(gf, FormatUsage.LoadStore) : gf;

    if (cgf == GraphicsFormat.None)
        throw new System.Exception("RenderTextureProbe: cannot find a replacement for " + TextureFormat);

    var r = GraphicsFormatUtility.GetRenderTextureFormat(cgf);
[...]

PR #640, opened 2020-08-31
Use supported GraphicsFormat for LodDataMgr

daleeidd:
Uses the new GraphicsFormat instead of RenderTextureFormat. GraphicsFormat can handle both textures and render textures. The API is experimental, but is being used in Unity's SRP packages for a while now.

This PR could go further in changing over to the new API for other classes like the OceanDepthCache, but it is probably better to start small and use it for something we really need first.

It also tries to use a format which supports the texture's usage like random access. This should help with #621.

I still need to test this on Windows.

Formats:
[RenderTextureFormat](https://docs.unity3d.com/ScriptReference/RenderTextureFormat.html)
[GraphicsFormat](https://docs.unity3d.com/ScriptReference/Experimental.Rendering.GraphicsFormat.html)

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgr.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgr.cs
@@ -3,6 +3,7 @@
 // This file is subject to the MIT License as seen in the root of this folder structure (LICENSE)
 
 using UnityEngine;
+using UnityEngine.Experimental.Rendering;
 using UnityEngine.Rendering;
 
 namespace Crest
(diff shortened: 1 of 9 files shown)


huwb:
Looks good and will give a nice memory saving for shadow if that precision works

daleeidd:
> will give a nice memory saving for shadow if that precision works

Hmmm... That's the one the GraphicsFormatUtility gave me for both MacOS and Windows based on RenderTextureFormat.RG16.

huwb:
Is that a good thing? Not sure if I understand

daleeidd:
I thought that it would be the same format that we have been using. So it is a lower precision format than RG16? The description for both is two channel, 8-bit.

There is R8G8_SRGB and R8G8_UNorm. They description that Unity gives reads the same to me. But GraphicsFormatUtility gave me the latter.

huwb:
In all my travels I have never seen something like RG16 mean 8 bit channels, but you're right that's what the doc says!! That's going to confuse me forever.

huwb:
Ok I've been getting confused.

Have been thinking of rg16 as r16g16 . My other comments about memory usage on the thread with Klaus about this were likely nonsense.

daleeidd:
No worries. I tested on Windows a little while ago and it worked. Merging this one in.

PR #641, opened 2020-08-31
Disable simulation on initialisation failure

daleeidd:
This should help with #621. It makes more sense once integrated with #640.

The idea is if a simulation fails to initialise, then it should be disabled. It should make it more robust where less important simulations are not necessary.

I wasn't sure how to handle animated waves since it appears to be almost necessary and more tightly coupled.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgr.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgr.cs
@@ -56,8 +56,7 @@ namespace Crest
 
         public virtual void Start()
         {
-            InitData();
-            enabled = true;
+            enabled = InitData();
         }
 
         public static RenderTexture CreateLodDataTextures(RenderTextureDescriptor desc, string name, bool needToReadWriteTextureData)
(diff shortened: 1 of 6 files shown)


huwb:
Maybe we should add a static create method for each lod data which returns null if the init failed which should save some logic.

Could the "failure" flags be removed in this case?

daleeidd:
> Maybe we should add a static create method for each lod data which returns null if the init failed which should save some logic.

So this would be defined on the LodDataMgr* class? That makes sense. Would be nicer than checking _enabled_. We could also use exceptions but not sure if that is good practice for game development.

> Could the "failure" flags be removed in this case?

The __wasLodData*InitFailure_ flags? We run CreateDestroySubSystems every ocean tick so we need something to prevent the init from running after failure. The only alternative that comes to mind is that we don't destroy the managers on failure and have them store state internally.

huwb:
Ahh ok

Something I've been thinking about is more effectively tracking settings changes rather than doing all the checks every frame. For example, pack the bools into a bitfield and only run OnCreateDestroy when this value changes. This would avoid calling the function most frames, and would probably be ok to remove those *InitFailure flags.

Failing that, the last alternative you outline would probably get my vote, if the logic added to lods is simple.

Issue #540, opened 2020-06-12
Support multi-camera scenarios


huwb:
We've had a few requests recently for supporting multiple simultaneous cameras.

This probably involves at least:

* Removing the global OceanRenderer.Instance singleton. Probably have a lookup function with viewpoint as input?
* Ocean geometry would need to know which ocean their rendering
* More..?

daleeidd:
The underwater environmental lighting component probably won't work well with this. Casting our own shadows instead of decreasing the light intensity could be a solution.

Timelapser:
This would be great to make split-screen multiplayer possible.

huwb:
This is still on the list, but I still don't have a timeline for when we might look at it as our bandwidth is absorbed by the day to day support and fixes and etc, and a small amount of feature dev.

We're hoping to increase our bandwidth through sponsorship and i wanted to mention that once here in case there is interest in supporting us: https://github.com/sponsors/wave-harmonic

daleeidd:
> Edit: Having tried it, it seems multiple cameras fight over the underwater effect, but otherwise seem to cooperate pretty well for my use case.

Underwater will support multiple cameras in the next version.

redwyre:
Hi, what's the current status for this? We have a game where the players can be quite far apart on the map so using a point between them as suggested in the docs doesn't seem like a good solution for us.

magenta404:
Closing this issue, as it was implemented in Crest 5.

PR #650, opened 2020-09-12
Explore what multiple ocean instances would mean

huwb:
**Status:** Definitely not ready to merge. I'm creating this as someone enquired about it and it took me a while to find it.

Experiment with what would be required for multiple oceans: #540 

This branch should work like before in the case of a single ocean. There are a few issues before it will support multiple instances.

Questions / problems that arose:

- [ ] RegisterLodDataClipSurface - this sets per-ocean data on MPB. might be tricky to fix..
- [x] How to connect underwater effect to ocean instance?
- [ ] How to connect underwater environment lighting with ocean instance?
- [ ] OceanChunkRenderer.OnWillRenderObject - tiles will need Layer so it only goes in one camera??

Misc notes:

- [ ] Validation gets passed "AnyInstance" in a few places, is this ok?
- [ ] RenderData.Validate - uses any instance
- [ ] Debuggui always takes first instance of ocean, shall we make this selectable?
- [ ] Lod data resolution should be global setting, right now taken from one instance

### DIFF
--- a/crest/Assets/Crest/Crest-Examples/BoatDev/Scripts/BoatAlignNormal.cs
+++ b/crest/Assets/Crest/Crest-Examples/BoatDev/Scripts/BoatAlignNormal.cs
@@ -66,7 +66,7 @@ public class BoatAlignNormal : FloatingObjectBase
     {
         _rb = GetComponent<Rigidbody>();
 
-        if (OceanRenderer.Instance == null)
+        if (OceanRenderer.AnyInstance == null)
         {
             enabled = false;
             return;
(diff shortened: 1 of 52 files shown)


daleeidd:
I am running the following assumption:

_Each camera can have many ocean instances.
Each ocean instance must have only one camera._

> How to connect underwater effect to ocean instance?

We could do a camera lookup. It would require #677 so that we have a reference to the camera.

> How to connect underwater environment lighting with ocean instance?

Same as above. Although, developers would have to use light layers and have two separate lights for it to work.

> OceanChunkRenderer.OnWillRenderObject - tiles will need Layer so it only goes in one camera??

We already have this for built-in. I think SRP uses something different. We could validate this if we have a camera property.

daleeidd:
> Validation gets passed "AnyInstance" in a few places, is this ok?

We should be passing all instances. Since these components are shared by both instances, they should be valid for both instances. The validate function will need to be updated to take list of OceanRenderers. I will look into doing this at some point.

PR #652, opened 2020-09-15
Fix broken underwater material

daleeidd:
Remove _ExecuteAlways_ for _UnderwaterEffect_. It didn't provide much benefit and is the primary cause of the material breaking.

But by removing it, it is possible for the underwater curtain to not be synced with the ocean due to shader stripping. Changing to multi_compile will prevent Unity from stripping the shader variants. I have brought back stripping based on the ocean material so we only ever get the variants we need. The ocean materials that are available to us to probe are ones that are used in the scene. The query seems pretty thorough as it will even find a material from an ocean being instantiated from prefab at runtime.

I have removed the copy on start param since it should always do this. This is inline with the underwater post process. I now just handle the meniscus by targeting it by shader name.

The build processor should also work for the underwater post processing too.

I have done a fair bit of testing and it appears to work correctly. Tested on MacOS but it should work on Windows too (will test later).


huwb:
Thanks this seems great to me

Minor feedback:

- license on new file
- could you please also add some documentation preamble to the new file that take about what it does and why
- debug logs Which probably should be removed (?)

That's all I spotted, am very happy for this to be merged once you've tested on Windows

daleeidd:
Thanks. All done. Went with wrapping the debug logs with a defines. I tested on Windows and there were no issues.

moosichu:
Hey @daleeidd, are you please able to merge these changes in a way that makes sense into the post process branch?

daleeidd:
@moosichu Done. If you need to do the same for the post-processing stack branch: a5898a41aba241dd711c039beb601190b63a5ab4.

PR #663, opened 2020-10-04
Implement post-process shader as custom V2 stack

moosichu:
This is mainly done because the post-process V2 stack has a feature to apply post processing directly to camera target:

"Use the final blit to the camera render target for postprocessing. This has less overhead but breaks compatibility with legacy image effect that use OnRenderImage."

In #299 this was reported to cause the entire screen to go black in VR.

Another benefit of this is that because the post-process stack is used (although in a different form) in both SRP renderers, the code for BIRP should be able to more-closely match those implementations.

However - this will probably affect #637, as it will change the underlying implementation of underwater post-processing. 

How this is merged-into the SRP renderers needs to be considered as well.

Finally - we need to make sure that it's OK to require people to install the post-processing stack if they want to use underwater, and have some validation for that.


moosichu:
This is the documentation I followed to implement this:

https://docs.unity3d.com/Packages/com.unity.postprocessing@2.3/manual/Writing-Custom-Effects.html

moosichu:
I've managed to get this functional - the PR itself will still need cleanup, but it's in a state where everything works. @daleeidd  if you can let me know how this fits-in with the VR work you are doing that would be great.

The pattern used here now more-closely aligns with what we are doing in HDRP, if we do decide to merge this into HDRP - I'm more than happy to handle re-aligning HDRP downstream so we can take advantage of this and reduce the size of the diffs we have.

moosichu:
image

This branch does have this issue, which will need resolving.

huwb:
thanks dude. i read the above but i think I'm going to need to lean on dale to give feedback when he has time, this stuff is really involved.

re the last point - requiring post proc stack - will this impact URP setup? we need to be quite gentle with urp setup steps so if it does impact URP then it would be good to get an illustration of how the setup would change. i'm less concerned if this is birp, although it would be good to understand what it means for birp as well (will it mean compile errors or can we make use of the package define magic to make it more graceful?).

sorry not very useful!

moosichu:
> Re the last point - requiring post proc stack - will this impact URP setup? 

No, both HDRP and URP come packaged with their own (completely different!) implementation of the post processing stack. Unfortunately it means that whilst the BIRP implementation is now *more similar* to what those two (will be), things will still be different.

This does not affect URP by adding extra reqs.

> i'm less concerned if this is birp-specific, although it would be good to understand what it means for birp as well (will it mean compile errors if the package is missing, or can we make use of the package define magic to make it more graceful?).

Yeah - we might need to add some package define magic. I have put all the code that depends on the post-processing stack in its own folder with its own asmdef which should simplify that. That's work that still needs to be done, but I've held-off doing as it shouldn't change-much here and is just busywork. I realise the last sentence in the PR of having validation for the existence of the post-process stack alluded to that as work that needed to be done - but wasn't super clear.

daleeidd:
I had a look at fixing VR/XR for this but my first attempt didn't go too well. The post processing stack is quite a bit different from a custom image effect.

daleeidd:
> image
> 
> This branch does have this issue, which will need resolving.

This is caused by the post processing package when scene reloading is disabled.

https://forum.unity.com/threads/how-do-i-fix-assertion-failed-on-expression-m_instanceid-instanceid_none.878329/#post-5890187

daleeidd:
> @daleeidd if you can let me know how this fits-in with the VR work you are doing that would be great.

The VR SPI situation is quite different between this and HDRP. SPI here behaves similar to multi-pass where the render function (in UnderwaterRenderer) is called per eye. Whereas for HDRP its render function is called once for both eyes. I have only checked the old VR system properly but not the new XR system.

It is possible to change HDRP to behave more like here by changing the DrawFullScreen single call to two calls. This would also allow us to drop the right eye specific code. Not sure if it would prevent the SPI performance benefits. Also not sure if the PPSv2 will change to be more like HDRP. The PPS SPI stuff is behind that of HDRP. Most of the effects were broken until recently.

I'll need a little more time with it.

daleeidd:
Closing as a separate underwater pass is used instead.

Issue #670, opened 2020-10-19
Remove foam input


Kandob:
There is a way to deposit foam onto the ocean but sometimes I would like to remove foam in shallow areas and I currently haven't found a way to do that, using a black texture instead of a white one doesn't work. Is it possible to add this feature?

daleeidd:
Thanks. Added. Will be out in 4.7.

Kandob:
I've tested it and it works great, thanks!

daleeidd:
Thanks. 4.7 is out so everyone else should have it too now.

PR #675, opened 2020-10-24
Add foam override to remove foam

daleeidd:
Solves #670 

The same shader as the shadow override but for foam. When used, the input _sets_ the value of the foam data. Can be used to add or remove foam.


huwb:
Thanks, approved.

PR #677, opened 2020-10-26
Add camera property to ocean renderer

daleeidd:
Adds a camera property so that we have both viewpoint and camera.

Before we only had a viewpoint transform which could differ from the camera transform. There are times where we must use the camera like when populating the shadow data. This was a bug which @skaughtx0r reported on Discord.

We could use the _main camera_ tag but using own property allows shadows to work with the scene camera correctly.

The viewpoint property should be kept as it still has use cases like camera zooming.

I've made _ViewerHeightAboveWater_ sample the camera transform as I couldn't think of a reason to have that separate.


huwb:
Looks good to me

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
```diff
@@ -68,12 +61,39 @@ public Transform Viewpoint
...
+                    }
+                }
 #endif
 
+                if (_camera != null)
+                {
+                    return _camera;
+                }
+
+                return Camera.main;
```
skaughtx0r, line 86: Should this set `_camera = Camera.main` to cache it before returning?

skaughtx0r:
Line 827 of `OceanRenderer.cs` should be:
```
Vector3 pos = Viewpoint.position;
```

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
daleeidd: I have removed setting the properties. We will now work directly from _Camera.main_. If anyone wants the performance benefit, they will have to set the properties themselves. I've added a warning for below 2019.4.9 since performance for _Camera.main_ has been greatly improved in that version. I have added some local caching here and there for when the camera property not set where it makes sense.

The reason being is that now camera switching will work out-of-the-box providing the _MainCamera_ tag is used.

daleeidd:
> Line 827 of `OceanRenderer.cs` should be:
> 
> ```
> Vector3 pos = Viewpoint.position;
> ```
> 
> Otherwise the ocean just follows the camera instead of the Viewpoint.

Sorry. I couldn't see the line you are referring to. Could you check again?

skaughtx0r:
@daleeidd Hmmm, I just checked the code in this branch and it looks fine.

I think I messed it up trying to merge this with my 2018 and PostProcessing branch.
Somehow in `LateUpdatePosition()` it was doing `ViewCamera.transform.position`.

PR looks good 👍

huwb (approved):
Looks great to me

daleeidd:
Thank you both.

PR #678, opened 2020-10-28
VR/XR Underwater Post Processing Stack

daleeidd:
Adds support for VR/XR to the underwater effect implemented for the post-processing stack.

Working:
- [x] Multi-Pass
- [ ] Single Pass
- [x] Single Pass Instanced

Single Pass is not working because the depth texture is double-wide and I am not sure how to sample that one correctly yet. We could drop single pass support for the underwater effect since it is deprecated.

One difficulty with the underwater post-process shader is that it is a CG shader and not an HLSL shader. It appears that Unity does not set the variables correctly for CG so sampling textures becomes a re-engineering effort. HLSL is easier for VR/XR from what I can see.

The other is that single-pass instanced behaves a bit different than it does in HDRP. The _Render_ function is called per eye. This is normally how multi-pass works. Some extra code had to be added to handle that.

Also, I am using the latest post-processing package from Unity's github. They have some fixes for SPI but I need to recheck if it is actually necessary.

This PR has #637 merged into #299 which this is branched from.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Crest.asmdef
+++ b/crest/Assets/Crest/Crest/Scripts/Crest.asmdef
@@ -1,3 +1,24 @@
-﻿{
-	"name": "Crest"
+{
+    "name": "Crest",
+    "references": [],
+    "includePlatforms": [],
+    "excludePlatforms": [],
+    "allowUnsafeCode": false,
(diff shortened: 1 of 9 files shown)


daleeidd:
Closing as is no longer needed. Will keep the branch around though.

PR #680, opened 2020-11-02
Use supported GraphicsFormat for LodDataMgr

daleeidd:
This is a fixed version of #640

Apologies. That was unpleasant. I missed a pretty significant part in the last PR. That is some data uses the texture format twice, and I was only applying it to one of them. So I reverted. This has everything included with the fix.

The only issue is that the foam sim settings has a property for the texture format which won't match up to the new one. I fixed ours but it will be an issue for others.

### DIFF
--- a/crest/Assets/Crest/Crest-Examples/Main/Data/SettingsFoam.asset
+++ b/crest/Assets/Crest/Crest-Examples/Main/Data/SettingsFoam.asset
@@ -17,4 +17,4 @@ MonoBehaviour:
   _waveFoamCoverage: 0.8
   _shorelineFoamMaxDepth: 0.8
   _shorelineFoamStrength: 0.5
-  _renderTextureFormat: 15
+  _renderTextureFormat: 45
--- a/crest/Assets/Crest/Crest-Examples/Main/Data/SettingsFoam.asset
+++ b/crest/Assets/Crest/Crest-Examples/Main/Data/SettingsFoam.asset
@@ -17,4 +17,4 @@ MonoBehaviour:
   _waveFoamCoverage: 0.8
   _shorelineFoamMaxDepth: 0.8
   _shorelineFoamStrength: 0.5
-  _renderTextureFormat: 45
+  _renderTextureGraphicsFormat: 45
--- a/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgr.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgr.cs
@@ -15,7 +15,11 @@ namespace Crest
     {
         public abstract string SimName { get; }
 
+        // This is the texture format we want to use.
         public abstract GraphicsFormat TextureFormat { get; }
+
+        // This is the platform compatible texture format we will use.
+        public GraphicsFormat CompatibleTextureFormat { get; private set; }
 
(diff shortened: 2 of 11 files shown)


huwb:
Thanks dude. Could you request review when you want eyes on this.. I took a quick look under the assumption that its ready now the changes look good to me.

crest/Assets/Crest/Crest-Examples/Main/Data/SettingsFoam.asset
```diff
@@ -17,4 +17,4 @@ MonoBehaviour:
   _waveFoamCoverage: 0.8
   _shorelineFoamMaxDepth: 0.8
   _shorelineFoamStrength: 0.5
-  _renderTextureFormat: 15
+  _renderTextureFormat: 45
```
huwb, line 20: That's correct right? I assume so but not easy to tell :). Are there any other bits of data like this that need fixup?

crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgr.cs
```diff
@@ -14,7 +15,11 @@ public abstract class LodDataMgr
     {
         public abstract string SimName { get; }
 
-        public abstract RenderTextureFormat TextureFormat { get; }
+        // This is the texture format we want to use.
+        public abstract GraphicsFormat TextureFormat { get; }
```
huwb, line 19: It may be an easy trap to fall into to use this desired/requested format in future code which could generate tricky bugs. Could it be marked as potentially incorrect - maybe call it 'RequestedGraphicsFormat' or something? Could it be made protected instead of public?

crest/Assets/Crest/Crest-Examples/Main/Data/SettingsFoam.asset
daleeidd: Yes and no.

I could use _RenderTextureFormat_ for the _SimSettingsFoam_ and use _GraphicsFormatUtility.GetGraphicsFormat_ to convert it in the _LodDataMgrFoam_? By doing so users get the much simpler _RenderTextureFormat_ enum dropdown (28 options versus 100+ options).

It has the added benefit of not having to change the serialised enum value. Otherwise, I think we should change the name of this property so it reverts to a sane default instead of having the wrong format serialised.

huwb: Thanks, i think i get it. Just realised if people have existing foam data, then it could give them the wrong value now if we use GraphicsFormat? Shall we just make a clean break and rename this param to _renderTextureGraphicsFormat? That will still lose the values users have currently (if anyone has actually changed it - prob unlikely), but it wont suddenly give them a completely different value? Lmk if that makes sense, i may not be understanding.

daleeidd: Sounds good. Done.

crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgr.cs
daleeidd: Sure. Done both.

huwb (approved):
Nice one thanks

daleeidd:
Thanks.

PR #692, opened 2020-11-19
Use ViewCamera in underwater post-processing

daleeidd:
Use the _ViewCamera_ and _ViewerHeightAboveWater_ properties on the _OceanRenderer_ for underwater effects instead of using the post-processing camera. _ViewCamera_ and the post-processing camera will be the same. Reduces code complexity and saves a height query. This will be merged into HDRP if okay to merge here.


huwb (approved):
Yeah if it doesnt change the functionality in any negative way thumbs up from me - looks good

daleeidd:
Thanks

Issue #662, opened 2020-10-02
Height Queries not working on Linux Standalone build (Query ring buffer exhausted.)


skaughtx0r:
**Describe the bug**
When running a standalone build of Crest on Linux the height queries are not working. The log is filled with `Query ring buffer exhausted. Please report this to developers.` messages.

**Screenshots / video**
Everything renders normally.

**Log**
```
Underwater must be enabled on the ocean material for UnderwaterPostProcess to work
UnityEngine.DebugLogHandler:Internal_Log(LogType, LogOption, String, Object)
UnityEngine.DebugLogHandler:LogFormat(LogType, Object, String, Object[])
UnityEngine.Logger:Log(LogType, Object, Object)
UnityEngine.Debug:LogError(Object, Object)
Crest.UnderwaterPostProcess:InitialisedCorrectly() (at /home/chops/repos/crest-oceanrender/crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterPostProcess.cs:85)
Crest.UnderwaterPostProcess:Start() (at /home/chops/repos/crest-oceanrender/crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterPostProcess.cs:114)

(Filename: /home/chops/repos/crest-oceanrender/crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterPostProcess.cs Line: 85)

Query ring buffer exhausted. Please report this to developers.
UnityEngine.DebugLogHandler:Internal_Log(LogType, LogOption, String, Object)
UnityEngine.DebugLogHandler:LogFormat(LogType, Object, String, Object[])
UnityEngine.Logger:Log(LogType, Object)
UnityEngine.Debug:LogError(Object)
Crest.SegmentRegistrarRingBuffer:AcquireNew() (at /home/chops/repos/crest-oceanrender/crest/Assets/Crest/Crest/Scripts/Collision/QueryBase.cs:100)
Crest.QueryBase:UpdateQueries() (at /home/chops/repos/crest-oceanrender/crest/Assets/Crest/Crest/Scripts/Collision/QueryBase.cs:486)
Crest.OceanRenderer:RunUpdate() (at /home/chops/repos/crest-oceanrender/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs:701)
Crest.OceanRenderer:LateUpdate() (at /home/chops/repos/crest-oceanrender/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs:689)

[...]

Issue #605, opened 2020-08-03
[URP] Query ring buffer exhausted


kroer:
Many error messages in a log after play.

E/Unity(20684): Query ring buffer exhausted. Please report this to developers.
E/Unity(20684): Crest.QueryBase:Update()

I don't know what info can help you, so ask any questions.

**Versions**
Unity 2019.4.6f1, URP 7.3.1, OceanCrest 4.3

**Platform**
 - Android 10

**Hardware**
 - Galaxy S10 (Mali-G76 MP12)
 - Galaxy Note 10 Lite (Mali-G72 MP18)

daleeidd:
Thank you for reporting.

Is this a developer or release build?
Does this issue only occur for some devices?
Is this one of our example scenes or one of your own?
We did make some improvements in this area for Crest 4.4. Would you be willing to try upgrading and see if it solves it?

kroer:
- Development build.
 - Now it reproduced only on Galaxy S10.
 - It occurs on our scene.
 - We update to 4.4 an hour ago and will be testing it again today or tomorrow.

If after the testing message will still appear, I will try to make a minimal scene to reproduce it and send it to you.

daleeidd:
Thanks. Look forward to what you find.

kroer:
We receive the error on 4.4 from time to time. Can't build a minimal scene to reproduce it yet. But the error message slightly changed.

08-04 15:13:13.820: E/Unity(9078): (Filename: ./Runtime/Export/Debug/Debug.bindings.h Line: 35)
08-04 15:13:13.884: E/Unity(9078): Query ring buffer exhausted. Please report this to developers.
08-04 15:13:13.884: E/Unity(9078): Crest.SegmentRegistrarRingBuffer:AcquireNew()
08-04 15:13:13.884: E/Unity(9078): Crest.QueryBase:UpdateQueries()
08-04 15:13:13.884: E/Unity(9078): Crest.OceanRenderer:RunUpdate()

huwb:
It might just be a case of the queries taking longer to do the round trip occasionally on some devices and we may just need to increase the number of buffers. I'll take a look for the param now.

huwb:
Could you please try increasing max requests to 5 or 6:

https://github.com/crest-ocean/crest/blob/master/crest/Assets/Crest/Crest/Scripts/Collision/QueryBase.cs#L28

huwb:
Just bumping this to check for an update

daleeidd:
[...]

PR #696, opened 2020-11-22
Increase query pool size

huwb:
Fix for #662 

@skaughtx0r will this fix the issue do you think? I've bumped pool size instead of max requests as a first step.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Collision/QueryBase.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Collision/QueryBase.cs
@@ -66,9 +66,8 @@ namespace Crest
         /// </summary>
         class SegmentRegistrarRingBuffer
         {
-            // Requests in flight plus 2 held values, plus one current, plus three more which a particular
-            // Linux setup needed :(
-            readonly static int s_poolSize = s_maxRequests + 2 + 1 + 3;
+            // Requests in flight plus 2 held values, plus one current
+            readonly static int s_poolSize = s_maxRequests + 2 + 1;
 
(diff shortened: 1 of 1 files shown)


skaughtx0r:
@huwb This fix does not work :frowning_face:

skaughtx0r:
@huwb I just tested this on another Linux desktop computer that's running an NVIDIA GTX 970 and I have the same problem. (My previous tests were on a laptop with a Nvidia Quadro M3000M)

huwb:
Thanks @skaughtx0r , updated

crest/Assets/Crest/Crest/Scripts/Collision/QueryBase.cs
```diff
@@ -25,7 +25,8 @@ public abstract class QueryBase
         protected abstract string QueryShaderName { get; }
         protected abstract string QueryKernelName { get; }
 
-        const int s_maxRequests = 4;
+		// 4 was enough for a long time, but Linux setups seems to demand 7
```
daleeidd, line 28: Comment indentation whitespace is tabs when file is spaces.

daleeidd (changes requested):

daleeidd:
#605 might also be fixed with this PR. Just linking the issue here.

crest/Assets/Crest/Crest/Scripts/Collision/QueryBase.cs
huwb: Thanks

huwb:
I assume that was the only feedback so i'm going to merge @daleeidd so i can incroporate it in the next release today. Hope thats ok!

PR #701, opened 2020-11-26
Fixed "_Strength" parameter being ignored.

emaber:
Looks like a simple oversight.


huwb:
Nice one. Could we get just the two relevant lines and the rest reverted? Thanks!

emaber:
Sure. I added a commit here. Not sure this is how you wanted it.

huwb (approved):
Thanks

PR #702, opened 2020-11-27
Set displacement to input properly

### LINKED ISSUE #666: 'Follow Horizontal Motion' not working (on dynamic waves from player boat in boat.unity)
**Describe the bug**
If i run boat.unity, part of the wake from the players boat meanders around and is detached. I looked into it for a while and it appears that something is going wrong with the displacement queries. If i disable the other boats it fixes it.

**Versions**
2019.4.1, editor

**To Reproduce**
Play boat.unity in editor.

**Platform**
Editor, win64

huwb:
Fixes #666 

Thanks for the hint @emaber !


daleeidd (approved):
Nice. Works well.

PR #704, opened 2020-11-28
Add underwater sun occlusion

### LINKED ISSUE #730: [HDRP] i would like subsurface scattering to be occluded by opaque geometry
i would like subsurface scattering to be occluded by opaque geometry - like shown on the screen shot attached.
here the submarine occludes the scattering (somehow).
you may also notice that subsurfacescattering here causes almost  no bending resulting in a relatively clear and smooth image :) 

sssOccluded

magenta404:
Occludes the subsurface scattering. Someone asked about this on Discord. I did it a little while ago so I thought I would put it here rather than let it sit.

Screenshot (2)
Before

Screenshot (1)
After

I don't think it looks correct for objects above water. Not sure whether to exclude them or if refraction needs to be applied.

It uses the depth buffer. The other option could be using the Unity shadow map to occlude. But I think that will be more work across pipelines.

Happy to hear your thoughts.

### Testing

Other than the keyword, it has two properties:

#### Sun Occlusion Distance Strength
Applied to the linear depth to make it more usable.

#### Sun Occlusion Fall-Off
The same as the SSS fall-off but for occlusion.

### DIFF
--- a/crest/Assets/Crest/Crest/Shaders/Ocean.shader
+++ b/crest/Assets/Crest/Crest/Shaders/Ocean.shader
@@ -163,6 +163,14 @@ Shader "Crest/Ocean"
 		// underwater effect is being used.
 		[Enum(CullMode)] _CullMode("Cull Mode", Int) = 2
 
+		[Header(Underwater Sun Occlusion)]
+		// Whether to occlude light scattering for the volume.
+		[Toggle] _UnderwaterSunOcclusion("Enable", Float) = 1
+		// Strength of the occlusion through the underwater fog.
+		_UnderwaterSunOcclusionDistance("Sun Occlusion Distance Strength", Range(0.0, 1000.0)) = 100.0
+		// Fall-off to affect directionality.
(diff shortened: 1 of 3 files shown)


huwb:
Thanks. Could you speak more to what it does and how? I did not understand what it does so far.

huwb:
Could the "distance strength" benefit from a different name. Might just be my understanding is missing, but just recording my initial thought as others may have the same

daleeidd:
Here are two more images which make the effect more obvious:

Screenshot (20)
Before

Screenshot (21)
After

> Could you speak more to what it does and how? I did not understand what it does so far.

It reduces the SSS colour added when looking towards the sun by using the depth texture. The issue was that there could be something a foot in front of the camera, but this SSS value would be present which looked strange. This solution isn't perfect but it seems to do a good enough job and adds some extra detail to underwater.

> Could the "distance strength" benefit from a different name. Might just be my understanding is missing, but just recording my initial thought as others may have the same

Sure. Not sure what a good name would be. The higher the value, the greater the distance from the camera the effect has influence over. All it does is bring the depth value into a more useable range. Maybe _Power_?

PR #707, opened 2020-12-08
Fix underwater normals not flipped with flow

daleeidd:
When rendering underwater, the ocean surface normals were not being flipped if flow was enabled.

This PR solves it by flipping normals only once after they have been computed. _n_geom_ is not flipped because we don't use it since it becomes _n_pixel_. Easier to just flip them once at the end. Happy to change it so that _n_geom_ is flipped.


huwb (approved):
Looks great nice one

PR #712, opened 2020-12-24
Add new Gerstner system

huwb:
**Status**: Ready to merge

This adds a new, leaner and meaner Gerstner system.

It renders the gerstner waves into a 'wave buffer' that is its own cascade independent of the other lod data, more slices but much lower X&Y res (32x32 by default). The wavelengths are snapped in a way that they always repeat spatially, so that the resulting patch can be tiled on the world XZ plane. Since it tiles, the patches are computed near the origin instead of moving with the viewer, which simplifies a few things.

The wave components are compiled in one list and generating the wave buffer is done in one dispatch. Thanks to this and the low resolution i was measuring <10us for the wave generation time on a 1060 gpu.

By default it does not resample the spectrum each frame in play mode. The gerstner sampling cost is non-negligible. This is on a toggle on the component. When this is disabled, the work done on the CPU by the gerstner system should be minimal, which should help a bunch (cf `OceanRenderer.LateUpdate()` cost in #624 , was around 4ms in those tests).

A new input shader `AnimWavesGerster` then samples the repeating tile and adds it to the displacements.

This change is additive and should have minimal impact on the existing systems. It is not added to any scene, i'll leave that for a future PR.

This new gerstner system also incorporates the wave direction fix #562.

**Performance before (main.unity, 1060 gpu)**

OceanRenderer.LateUpdate (CPU): 0.78ms
Wave generation (GPU): 0.330ms

**Performance after**

OceanRenderer.LateUpdate (CPU): 0.5ms
Wave generation (GPU): 0.00941ms generation + 0.206ms wave integration = 0.215ms
(The integration pass could be done in a single dispatch. for the global waves case at least.. but its not totally trivial to do this so ill leave this for future work.)

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
+++ b/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
@@ -673,6 +673,7 @@ namespace Crest
             sp_ForceUnderwater = Shader.PropertyToID("_ForceUnderwater");
             sp_perCascadeInstanceData = Shader.PropertyToID("_CrestPerCascadeInstanceData");
             sp_cascadeData = Shader.PropertyToID("_CrestCascadeData");
+            sp_crestTime = Shader.PropertyToID("_CrestTime");
         }
 
         void LateUpdate()
--- a/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
+++ b/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
@@ -290,7 +290,7 @@ namespace Crest
 
         bool _canSkipCulling = false;
 
-        readonly int sp_crestTime = Shader.PropertyToID("_CrestTime");
+        public static int sp_crestTime = Shader.PropertyToID("_CrestTime");
         readonly int sp_texelsPerWave = Shader.PropertyToID("_TexelsPerWave");
         readonly int sp_oceanCenterPosWorld = Shader.PropertyToID("_OceanCenterPosWorld");
         readonly int sp_meshScaleLerp = Shader.PropertyToID("_MeshScaleLerp");
@@ -1147,8 +1147,9 @@ namespace Crest
             }
 
             // ShapeGerstnerBatched
-            var gerstners = FindObjectsOfType<ShapeGerstnerBatched>();
-            if (gerstners.Length == 0)
+            var gerstnerBatchs = FindObjectsOfType<ShapeGerstnerBatched>();
+            var gerstners = FindObjectsOfType<ShapeGerstner>();
+            if (gerstnerBatchs.Length == 0 && gerstners.Length == 0)
             {
                 showMessage
                 (
--- a/crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
@@ -301,7 +301,7 @@ namespace Crest
                     int vi = outputIdx / 4;
                     int ei = outputIdx - vi * 4;
 
-                    _waveData[vi]._twoPiOverWavelength[ei] = 2f * Mathf.PI / _wavelengths[componentIdx];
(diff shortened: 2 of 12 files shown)


crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
```diff
@@ -0,0 +1,619 @@
...
+                ampSum += _amplitudes[i] * _spectrum._chopScales[i / _componentsPerOctave];
+            }
+            OceanRenderer.Instance.ReportMaxDisplacementFromShape(ampSum * _spectrum._chop, ampSum, ampSum);
+        }
+
+        void InitBatches()
+        {
+            var registered = RegisterLodDataInputBase.GetRegistrar(typeof(LodDataMgrAnimWaves));
+
+            //#if UNITY_EDITOR
```
daleeidd, line 483: Does this need to be uncommented?

daleeidd (approved):
Looks great. I didn't experience any issues in testing other than the non-blocking issue of _debugDrawSlicesInEditor option overlaps with the other data when enabled.

crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
```diff
@@ -0,0 +1,619 @@
...
+        [Tooltip("Change to get a different set of waves.")]
+        public int _randomSeed = 0;
+
+        [Tooltip("Resolution to use for wave generation buffers. Low resolutions are more efficient but can result in noticeable patterns in the shape."), Delayed]
+        public int _resolution = 32;
+
+        [Tooltip("In Editor, shows the wave generation buffers on screen."), SerializeField]
+        bool _debugDrawSlicesInEditor = false;
+
+        Mesh _meshForDrawingWaves;
```
daleeidd, line 52: Should this be serialised?

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
```diff
@@ -290,7 +290,7 @@ public enum DefaultClippingState
 
         bool _canSkipCulling = false;
 
-        readonly int sp_crestTime = Shader.PropertyToID("_CrestTime");
+        public static int sp_crestTime = Shader.PropertyToID("_CrestTime");
```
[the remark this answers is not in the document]
huwb: Ah thanks. I think it does indeed need to be re-init'd on domain reload. I think all statics do. I added this initalisation. Now it cant be read only haha. So i addressed this feedback by adding the initialisation. Hope that sounds right!

crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
huwb: No i think it should be removed, which i've done now. I dont really recall why there are ifdefs there. I feel it is best to just remove the old batches every time, whether in editor or standalone, so i am making it do that.

crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
huwb: Well spotted. I didnt explain this.

This is currently internal to the script and not serialised. I have downstream changes to generate the mesh from a spline, in which case the spline would set this value.

This plan does not expose this field to the user, so they won't be able to provide their own meshes, currently. I have the feeling that no one uses this functionality (the required gerstner shader to render waves weighted by vertex colour has never been requested. it was added recently by me after i recommended it to someone, and they did not use it in the end).

The spline stuff that is coming later is quite powerful (can create lots of interesting wave areas). Given this i propose not exposing this to the user on this script - for now at least. Let me know what you think.

daleeidd: Makes perfect sense.

PR #713, opened 2020-12-28
Give hint if combine shader fails to load

huwb:
**Status**: Ready to merge

Minor but I had a case where the combine shader failed (spuriously) to load resulting in a un-human-readable error. This gives a user facing error and hint to resolve it.


daleeidd (approved):
Looks good.

PR #717, opened 2021-01-03
Wave variance

huwb:
**Status**: Ready for review. I have arranged these changes into logical commits so reviewing by-commit is a good option.

**Overview**:
Wave variance is statistical measure of how many waves there are in each slice of the displacement cascade. When sampling a low LOD of the waves, there are a lot of small wavelengths missing. The variance is a measure of the cumulative amount of missing content for a given slice. This measure can be used to compensate for the missing data. The killer use case of this is the foam at wave crests - right now we dont get foam at all in the mid-to-background as the pinchy waves are missing from the lower lods. The variance is used to compensate for this missing data and foam is now generated reliably at all LODs.

The variance is stored in the W channel of the displacement texture. Variance can be spatially varying so its put into the textures rather than factored out. This channel used to be used for SSS. **SSS has been moved away from this channel and put back into the ocean shader which computes it from surface pinch**. This is the way it used to be originally, and the results are very comparable.

Variance can in theory be useful for accounting for missing waves when approximating SSS, and also for driving surface material roughness in the distance. I didnt have a good fail case for the former, so i've left it disabled for now. I had a quick stab at the latter and it did work, but needs more love to make it mergeable, and it wasnt clear that driving it procedurally was a good idea rather than setting it manually, and this experiment is parked in `feature/variance-roughness`.

This only works with the new gerstner system. For the old gerstner system, the behaviour is unchanged from what it was before (the variance measure will just come through as 0 and have no impact).

**Testing**:
I have set up main.unity with the new gerstner component so can be used to test the changes.

### DIFF
--- a/crest/Assets/Crest/Crest/Shaders/OceanHelpersNew.hlsl
+++ b/crest/Assets/Crest/Crest/Shaders/OceanHelpersNew.hlsl
@@ -71,9 +71,8 @@ void SampleDisplacementsNormals(in Texture2DArray i_dispSampler, in float3 i_uv_
 	// SSS - based off pinch
 #if _SUBSURFACESCATTERING_ON
 	{
-		// Don't use variance - yet
-		// float varianceBeforeThisCascade = data.w;
 		float2x2 jacobian = (float2x2(disp_x.xz, disp_z.xz) - disp.xzxz) / i_texelSize;
+		// Determinant is < 1 for pinched, < 0 for overlap/inversion
 		float det = determinant( jacobian );
 		io_sss += i_wt * saturate( 0.60 - det * 0.12 );
 	}
--- a/crest/Assets/Crest/Crest/Shaders/OceanHelpersNew.hlsl
+++ b/crest/Assets/Crest/Crest/Shaders/OceanHelpersNew.hlsl
@@ -71,10 +71,12 @@ void SampleDisplacementsNormals(in Texture2DArray i_dispSampler, in float3 i_uv_
 	// SSS - based off pinch
 #if _SUBSURFACESCATTERING_ON
 	{
-		float2x2 jacobian = (float2x2(disp_x.xz, disp_z.xz) - disp.xzxz) / i_texelSize;
+		const float2x2 jacobian = (float2x2(disp_x.xz, disp_z.xz) - disp.xzxz) / i_texelSize;
 		// Determinant is < 1 for pinched, < 0 for overlap/inversion
-		float det = determinant( jacobian );
-		io_sss += i_wt * saturate( 0.60 - det * 0.12 );
+		const float det = determinant( jacobian );
+		const float sssMax = 0.6;
+		const float sssRange = 0.12;
+		io_sss += i_wt * saturate( sssMax - sssRange * det );
 	}
 #endif // _SUBSURFACESCATTERING_ON
 
--- a/crest/Assets/Crest/Crest/Shaders/OceanHelpersNew.hlsl
+++ b/crest/Assets/Crest/Crest/Shaders/OceanHelpersNew.hlsl
@@ -69,6 +69,7 @@ void SampleDisplacementsNormals(in Texture2DArray i_dispSampler, in float3 i_uv_
 	}
 
 	// SSS - based off pinch
+#if _SUBSURFACESCATTERING_ON
 	{
 		// Don't use variance - yet
 		// float varianceBeforeThisCascade = data.w;
@@ -76,6 +77,7 @@ void SampleDisplacementsNormals(in Texture2DArray i_dispSampler, in float3 i_uv_
 		float det = determinant( jacobian );
 		io_sss += i_wt * saturate( 0.60 - det * 0.12 );
 	}
+#endif // _SUBSURFACESCATTERING_ON
 
 	io_nxz += i_wt * n.xz;
 }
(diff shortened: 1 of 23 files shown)


daleeidd (approved):
Huge improvement. Only minor feedback but can be skipped.

There are some possible changes to SSS pinch but I will discuss on Discord as it is not really what this PR is about.

crest/Assets/Crest/Crest/Shaders/OceanHelpersNew.hlsl
```diff
@@ -32,26 +35,48 @@ float2 IDtoUV(in float2 i_id, in float i_width, in float i_height)
...
+	{
 		n = normalize(cross(disp_z - disp, disp_x - disp));
 	}
+
+	// SSS - based off pinch
+	{
+		// Don't use variance - yet
+		// float varianceBeforeThisCascade = data.w;
+		float4 du = float4(disp_x.xz, disp_z.xz) - disp.xzxz;
+		float det = (du.x * du.w - du.y * du.z) / (i_texelSize * i_texelSize);
```
daleeidd, line 76: We could use the built-in [determinate](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-determinant) function? Code would be:

```shader
float2x2 du = float2x2(disp_x.xz, disp_z.xz) - disp.xzxz;
float det = determinant(du) / (i_texelSize * i_texelSize);
```

Results looked the same.

I don't think there are any disadvantages. My understanding is that built-in function always perform better than custom ones (I read this during on of AMD's performance presentations if I recall correctly). I tried to avoid casting too since that can be another cost. I haven't measured this though.

Let me know what you think.

crest/Assets/Crest/Crest/Shaders/OceanHelpersNew.hlsl
```diff
@@ -32,26 +35,48 @@ float2 IDtoUV(in float2 i_id, in float i_width, in float i_height)
...
+		io_sss += i_wt * saturate( .03 * (20.0 - (det * 4.0 /*- varianceBeforeThisCascade * 20.0*/)) );
+	}
```
daleeidd, line 78: This could be wrapped in *_SUBSURFACESCATTERING_ON* to be explicit. But I think this should be stripped from the shader if that is not true anyway since the value would never be used.

daleeidd (approved):

crest/Assets/Crest/Crest/Shaders/OceanHelpersNew.hlsl
huwb: good shout, thanks! i made it a bit more maths'y as well, and baked in the multipliers to clean it up

crest/Assets/Crest/Crest/Shaders/OceanHelpersNew.hlsl
huwb: makes sense, done

crest/Assets/Crest/Crest/Shaders/Resources/UpdateFoam.compute
```diff
@@ -83,9 +85,9 @@ void UpdateFoam(uint3 id : SV_DispatchThreadID)
 	// < 0: Overlap
 	float4 du = float4(disp_x.xz, disp_z.xz) - disp.xzxz;
 	float det = (du.x * du.w - du.y * du.z) / (cascadeData._texelWidth * cascadeData._texelWidth);
-	foam += 5.0 * _SimDeltaTime * _WaveFoamStrength * saturate(_WaveFoamCoverage - det);
+	foam += 5.0 * _SimDeltaTime * _WaveFoamStrength * saturate( _WaveFoamCoverage - det + foamBase*0.7 );
```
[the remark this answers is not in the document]
huwb: done

crest/Assets/Crest/Crest/Shaders/OceanHelpersNew.hlsl
```diff
@@ -32,26 +35,48 @@ float2 IDtoUV(in float2 i_id, in float i_width, in float i_height)
...
 		n = normalize(cross(disp_z - disp, disp_x - disp));
 	}
+
+	// SSS - based off pinch
+	{
+		// Don't use variance - yet
+		// float varianceBeforeThisCascade = data.w;
+		float4 du = float4(disp_x.xz, disp_z.xz) - disp.xzxz;
+		float det = (du.x * du.w - du.y * du.z) / (i_texelSize * i_texelSize);
+		io_sss += i_wt * saturate( .03 * (20.0 - (det * 4.0 /*- varianceBeforeThisCascade * 20.0*/)) );
```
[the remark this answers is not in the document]
huwb: done

huwb:
@daleeidd  addressed feedback so far

the SSS is different from the old SSS. i tried to chill it out a bit but had limited success. i think your suggestion is probably needed - computing an averaged / lower lod. Perhaps it could be done in the vert shader.

if you agree i think we can merge this. ill be keen to see if the results hold up in URP/HDRP

crest/Assets/Crest/Crest/Shaders/OceanHelpersNew.hlsl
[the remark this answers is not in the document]
daleeidd: Thanks

daleeidd:
Sounds good. And agreed to merging.

PR #719, opened 2021-01-05
Move global shader variables into new OceanInput

daleeidd:
This is to converge with changes introduced downstream. Created OceanInput.hlsl and moved global shader variables into it. Also rearranged the includes on Ocean.shader to match downstream.


crest/Assets/Crest/Crest/Shaders/OceanInput.hlsl
```diff
@@ -0,0 +1,93 @@
...
+half _SubSurfaceBase;
+half _SubSurfaceSun;
+half _SubSurfaceSunFallOff;
+
+half _SubSurfaceDepthMax;
+half _SubSurfaceDepthPower;
+half3 _SubSurfaceShallowCol;
+half3 _SubSurfaceShallowColShadow;
+
+half _CausticsTextureScale;
```
huwb, line 42: should these be behind #ifdef CAUSTICS , and similar questions for other params that are behind shader options? maybe its fine as it is - maybe unity removes unused ones anyway..

huwb (changes requested):
Happy to see this change, good stuff

My only request would be could we find a different name than OceanInput, as this combination of words is heavily used in crest for ocean inputs. I see a file LitData.hlsl in SRP that serves this kind of purpose. Maybe OceanShaderData.hlsl? (If you do use this name then consider it approved from my POV).

Thanks!

crest/Assets/Crest/Crest/Shaders/OceanInput.hlsl
daleeidd: I think they do. I was just matching what we have downstream. I will add them in here and downstream.

daleeidd:
> What are the downstream changes? I think i can make a pretty good guess but just curious.

- Global shader variables are in their own file
- The includes are grouped before the vert function in Ocean.shader
- Less ifdefs (I will add them back in and add them downstream instead)
- Some includes are cleaned up like in Underwater Curtain

> My only request would be could we find a different name than OceanInput, as this combination of words is heavily used in crest for ocean inputs. I see a file LitData.hlsl in SRP that serves this kind of purpose. Maybe OceanShaderData.hlsl? (If you do use this name then consider it approved from my POV).

Can do.

daleeidd:
Done all of those. I have reorganised the file and added some headers.

huwb (approved):
Looks good

daleeidd:
Thank you

Issue #665, opened 2020-10-09
White LOD borders when far from Center StandardRP , Endless ocean not possible ? 


nitroman123:
**Describe the bug**
The LOD border are showed when upgrade to a newer version of crest. As Dale told was caused becasue of big tranform. I think its something else cause in the previous version of crest ocean was always 6000+ on x and the same o z axis and i will prove it with a screenshot in my backup files. 

**Screenshots / video**
![Artifacts](https://user-images.githubusercontent.com/62849221/95625630-71215180-0a81-11eb-9aec-3ef313beb719.png)
**I even try to move ocean very far away and camera again far about 2000 on x,z axis no artifacts ** 
![λήψη](https://user-images.githubusercontent.com/62849221/95625848-dc6b2380-0a81-11eb-8f42-a1788c589ca3.png)
**Previous Version of Crest Ocean position** 
![OceanNoartifactOld](https://user-images.githubusercontent.com/62849221/95626904-dece7d00-0a83-11eb-8bbf-86f670056e9b.png)
**Camera Position from 0,0,0 (ocean position)** 
![CameraPosition](https://user-images.githubusercontent.com/62849221/95626989-10474880-0a84-11eb-867a-d9abe2bbe066.png)

**Versions**
Standard RP Latest 2019(f12)LTS 

**To Reproduce**
Dale has the prefab i send him 

**Platform**
 - Editor
 - Platform Windows

**Hardware**
CPU Ryzen 3700x , RTX 3080

**Additional context**
So after some updates we cant get an endless ocean ? Never had this issue before

daleeidd:
This is caused by the precision fixes for the grid (searchable in source by using keyword _:OceanGridPrecisionErrors_). The epsilon value added to the scale of the ocean chunk was suppose to solve these very fine gaps, but it appears not to work very well once the camera goes very far from centre. Increasing the value to 1.1 does, but that is quite a large tweak.

Another option is to use a floating origin which should solve other possible issues when the camera is that far out. Have you tried this?

nitroman123:
I just tried this , no result still the same problem.

huwb:
Could we get a reproduction case pushed to a branch.

@nitroman123 are you able to reproduce this issue in one of the example scenes, like main.unity? If so could you paste a screenshot of hte camera transform like you did above?

Thanks

nitroman123:
[...]

PR #720, opened 2021-01-07
Fix gaps and cracks issues

huwb:
**Status**: Ready to merge

Fix for two issues, both tracked by #665 

**Gaps**: When ocean placed near a regular position (such as an integral boundary or multiple of 0.333333, etc), precision issues related either to how transforms propagate in the unity hierarchy or how the values arrive to the shader mean different tiles can snap to different locations and leave a gap. This adds a slight nudge to push the ocean away from a bunch of common boundaries). The nudge is very small and will not affect the ocean simulation/rendering as everything snaps to much larger values than the nudge. In the future, if gaps appear, it may be the boundary is not on 1/60. The boundary might be lod resolution / lod 0 width. If we get a report of gaps, we should get the camera transform and oceanrenderer settings, and see if we can compute the boundary.

**Thin cracks**: The scale of each tile is increased by a small epsilon. This should be subpixel, but should be large enough to overcome precision issues. If cracks appear in the future, this value could be increased. It might be that this value is too small when far from the origin, we could potentially scale the epsilon by dist from origin to help combat this if needed.

@daleeidd could you let me know what you think and please give this a run on mac if you could. Thanks!

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
+++ b/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
@@ -812,6 +812,7 @@ namespace Crest
             // Don't land very close to regular positions where things are likely to snap to, because different tiles might
             // land on either side of a snap boundary due to numerical error and snap to the wrong positions. Nudge away from
             // common by using increments of 1/60 which have lots of factors.
+            // :OceanGridPrecisionErrors
             if (Mathf.Abs(pos.x * 60f - Mathf.Round(pos.x * 60f)) < 0.001f)
             {
                 pos.x += 0.002f;
--- a/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
+++ b/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
@@ -809,6 +809,18 @@ namespace Crest
             // maintain y coordinate - sea level
             pos.y = Root.position.y;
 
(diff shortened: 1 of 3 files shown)


crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
```diff
@@ -809,6 +809,18 @@ void LateUpdatePosition()
             // maintain y coordinate - sea level
             pos.y = Root.position.y;
 
+            // Don't land very close to regular positions where things are likely to snap to, because different tiles might
+            // land on either side of a snap boundary due to numerical error and snap to the wrong positions. Nudge away from
+            // common by using increments of 1/60 which have lots of factors.
```
daleeidd, line 814: Should we add `:OceanGridPrecisionErrors` to keep track?

daleeidd (approved):
Works really well from the test cases. I found for the scaling fix a value as low as 1.0000001 worked for the test case.

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
huwb: Done, thanks

huwb:
Great news, thanks!

I updated the descr above about what we might try if we get reports of gaps or cracks in the future.

Merging!

PR #723, opened 2021-01-11
Orthographic Projection

### LINKED ISSUE #426: Ocean not rendering properly using orthographic camera
**Describe the bug**
Using orthographic camera, ocean should show up on top of terrain. But it looks like the terrain renders on top of the ocean instead, even though terrain is positioned below the ocean.

**Screenshots / video**
I'm rendering to a texture here but that's not necessary to reproduce
Perspective camera works as expected:
persp

Here's orthographic. Where is the blue color? Where is the foam?:
ortho

And here's orthographic with the terrain disabled. There's the blue color and foam! It was under the terrain?:
ortho_no_terrain
**Log**
nope

**Versions**
Using Unity 2019.3.1f1 and built-in pipeline

**To Reproduce**
Put terrain in scene
Put Crest in scene positioned above terrain
Change camera to orthographic and look down

**Platform**
 - Editor/Windows

**Hardware**
Mobile gtx 1050

**Additional context**
nope

daleeidd:
Fixes #426

Adds orthographic projection support. By default the ocean shader supports both orthographic and perspective. But users can choose a mode on the material using a keyword enum to save on performance.

Supporting orthographic projection was just a matter of fixing depth when the camera is orthographic.

I also cleaned up some unused depth references in the shader.

### DIFF
--- a/USERGUIDE.md
+++ b/USERGUIDE.md
@@ -70,6 +70,14 @@ This base reflection can then be overridden by dynamic planar reflections. This 
 
 A good strategy for debugging the use of Unity's specular cubemap is to put another reflective/glossy object in the scene near the surface, and verify that it is lit and reflects the scene properly. Crest tries to use the same inputs for lighting/reflections, so if it works for a test object it should work for the water surface as well.
 
+### Orthographic Projection
+
+Crest supports orthographic projection out-of-the-box, but it might require some configuration to get a desired appearance.
+
+Crest uses the camera's position for the LOD system which can be awkward for orthographic which uses the size property on the camera.
+Use the *Viewpoint* property on the *Ocean Renderer* to override the camera's position.
+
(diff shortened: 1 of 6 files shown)


huwb (approved):
Should the option on the material be a multi compile, or a shader feature? If all variants are not needed I would assume a shader feature.

Besides that it looks great, nice one

daleeidd:
> Should the option on the material be a multi compile, or a shader feature? If all variants are not needed I would assume a shader feature.

Forgot to change that one. Thank you.

daleeidd:
Seems I was a bit eager. Forgot to use the new linear eye depth function for pixelZ. Commit cc131803f12 fixes it.

Caustics can distort strangely depending where the camera is, but seems to hold up well enough. I haven't noticed anything else.

Issue #371, opened 2019-12-09
Ocean depth cache does not work with Gaia terrain


huwb:
**Describe the bug**
Ocean depth cache does not work with Gaia terrain

**Screenshots / video**
![MfiUfKE](https://user-images.githubusercontent.com/939901/70420722-3bc26680-1a60-11ea-8f69-626ca75b8836.jpg)

**Log**

**Versions**
Appears to be 2019.2.15. Built-in RP.

**To Reproduce**

**Platform**
Editor / windows

**Hardware**

**Additional context**

huwb:
May be due to GPU instancing of terrain: https://github.com/crest-ocean/crest/issues/158

huwb:
That was the issue.

huwb:
Caught in validation in 3.4

PR #724, opened 2021-01-12
Fix Terrain "Draw Instanced" not working with Ocean Depth Cache

### LINKED ISSUE #158: Terrain ocean depth - draw instanced
I just did a pull from today to test the changes to project since I last looked.

I noticed that if you use Unity terrain and have "Draw Instanced" enabled 
For some reason it does not get the depth mask correctly, it completely ignores the terrain.

Switch off the "Draw Instanced" and it's all fine.
This is the option in the Terrain settings which arrived with Unity 2018.3

It's something to be aware of, not sure how easy it is to work around as I'm not 100% certain why it's affecting it like this.

daleeidd:
Fixes #158 #371

Based on @huwb's branch: [fix/cache-terrain-depths](https://github.com/crest-ocean/crest/tree/fix/cache-terrain-depths)

Brings over fixes from downstream. Copies depth instead of rendering geometry into ocean depth cache. This didn't work before in 2018 but now works in 2019. Tested on MacOS in build and standalone.

These files are now same as downstream with a few exceptions:
- The copy depth shader use CG instead of HLSL, but they are only a few lines that are different.
- This C# file has a few additional lines which will be added downstream. Allows the depth cache to be used in a prefab

The diff is messy so it is easier to compare these files with the downstream files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,6 @@ If you encounter an issue, please search the [Issues page](https://github.com/hu
 There are a few issues worth calling out here:
 
 * Sky solutions such as Azure[Sky] requires some code to be added to the ocean shader for the fogging/scattering to work. This is a requirement of these products which typically come with instructions for what needs to be added. See the [wiki](https://github.com/crest-ocean/crest/wiki) for examples.
-* This built-in render pipeline version of crest requires the *Draw Instanced* option on terrains to be disabled at start time. It can be re-enabled subsequently after the depth cache is populated. See [issue #158](https://github.com/crest-ocean/crest/issues/158).
 * *Crest* does not support OpenGL or WebGL backends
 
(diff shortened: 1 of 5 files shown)


daleeidd:
Tested on Windows in the editor and standalone and it works.

huwb (approved):
Looks great. Is there a shader for populating the depth cache that can be removed as well? I can't remember and I don't have my laptop handy so I might be taking rubbish.

daleeidd:
Doesn't appear to be. There is only three depth cache shaders:
OceanDepths (used for register inputs)
OceanDepthsCache (used with register input on the cache quad)
CopyDepthIntoCache (used to blit camera texture to cache texture)

daleeidd:
And thanks. Merging.

PR #725, opened 2021-01-14
Add scripting define symbols

### LINKED ISSUE #715: Add compile symbol, to know from other packages, that Crest is installed
To improve integration with other packages, it would be useful for CREST to add its own build symbol so that integration with other packages can be managed.

Just add this:

```csharp
using System.Collections.Generic;
using System.Linq;
using UnityEditor;

public class CrestRemoveDefines : UnityEditor.AssetModificationProcessor
{
    static string symbols;

    public static void OnWillCreateAsset(string assetName)
    {
        AddDefineSymbol("CREST");
    }

    public static AssetDeleteResult OnWillDeleteAsset(string assetPath, RemoveAssetOptions rao)
    {
        RemoveDefineSymbol("CREST");

        return AssetDeleteResult.DidNotDelete;
    }

    public static void AddDefineSymbol(string symbol)
    {
        string definesString = PlayerSettings.GetScriptingDefineSymbolsForGroup(EditorUserBuildSettings.selectedBuildTargetGroup);
        List<string> allDefines = definesString.Split(';').ToList();
        allDefines.Add(symbol);
        PlayerSettings.SetScriptingDefineSymbolsForGroup(EditorUserBuildSettings.selectedBuildTargetGroup, string.Join(";", allDefines.ToArray()));
    }

    public static void RemoveDefineSymbol(string symbol)
    {
        string symbols = PlayerSettings.GetScriptingDefineSymbolsForGroup(EditorUserBuildSettings.selectedBuildTargetGroup);

        if (symbols.Contains(symbol))
        {
            symbols = symbols.Replace(symbol + "; ", "");
            symbols = symbols.Replace(symbol, "");
            PlayerSettings.SetScriptingDefineSymbolsForGroup(EditorUserBuildSettings.selectedBuildTargetGroup, symbols);
        }
    }
}
```

Then, from the other scripts, it can be used something like:
```
#if CREST
:
#endif
```

daleeidd:
Fixes #715

Adds global scripting define symbols so other assets know Crest is installed. Symbols are added and removed based on ***this*** file being added and removed make it self-contained.

Tested to be working on both MacOS and Windows.


huwb:
I havent encountered this add/remove pattern, interesting. So the symbol is always added by this file, or not added if the file is not present. And if the file is removed, then it removes the symbol.

is this a pattern you've seen used elsewhere? Will adding the symbol in this way trigger a double-compile, do you think?

daleeidd:
> is this a pattern you've seen used elsewhere?

I haven't.

> Will adding the symbol in this way trigger a double-compile, do you think?

I believe so. But the other pattern would do the same.

daleeidd:
Just adding that I will commit the symbols in the project settings for the project after testing.

huwb:
Hopefully we can remove this stuff if we can one day move to packages. For now, approved :)

huwb (approved):

daleeidd:
Thanks. Oh, I forgot, but did you want CREST_OCEAN_BIRP for built-in? And will downstream be *CREST_OCEAN, CREST_OCEAN_SRP and CREST_OCEAN_\*\*RP*?

huwb:
It just occurred to me that we probably want CREST_OCEAN present for all, as other clients will really need a single way to know if crest is present, I guess?

Then also adding an additional CREST_OCEAN_xRP on top is optional, and maybe useful (maybe for us more than others, maybe it means we can unity code in srp branch).

daleeidd:
> It just occurred to me that we probably want CREST_OCEAN present for all, as other clients will really need a single way to know if crest is present, I guess?

Sorry. I meant CREST_OCEAN_BIRP in addition to CREST_OCEAN for built-in.

> Then also adding an additional CREST_OCEAN_xRP on top is optional, and maybe useful (maybe for us more than others, maybe it means we can unity code in srp branch).

Sure. Sounds good.

patrickp-unity3d:
Hi @daleeidd, I second that this would be very useful indeed. It follows the same design pattern as GAIA (#if GAIA_PRESENT) from Procedural Worlds. Is this already available as part of a version of the Crest asset bundle?
Great work 👍

daleeidd:
> Is this already available as part of a version of the Crest asset bundle?

Not currently. But going to look at merging this in for 4.15.

PR #731, opened 2021-01-19
Wave splines

huwb:
**Status**: Ready to merge.

**Overview**:

image

This change adds a simple spline object to crest, and some utility functions to sample the spline. It also has some UI helpers in the inspector, and first pass validation.

If the spline is attached to a GO that has a ShapeGerstner component, then the gerstner waves will generate a strip of geometry along the spline and use this geometry to place waves.

I have not written proper documentation. I think a later PR should use the new gerstner component and this spline in our default scenes, and update the documentation.

**Testing**:

Add a Spline and a ShapeGersnter component to a GameObject. Help boxes in the inspector should guide the rest of the setup.

Worked in 2019.4.16 win64 editor and standalone.


daleeidd:
This branch is based of *feature/refactored-gerstner* which has duplicate changes that were merged from *feature/refactored-gerstner-2*. I can see changes that are in master which are showing up as new here. Would merging master into this branch help clear it up? Otherwise, I am going over code we already have.

Also, I have tested it and it works. It will be easier to review with the test data gone if you don't need it anymore.

huwb:
Thanks data reverted now

Hm it should be the legit diff - if you see something fishy let me know and ill dig deeper.

huwb:
I have been grappling with getting gizmos that are visible and also easily clickable. It turns out icons are drawn as 2D elements and are pretty clickable, as long as the icon size is turned up in the editor (or 3d gizmos disabled), so i've recruited an unrelated icon for the cause. not perfect but works ok-enough i guess.

crest/Assets/Crest/Crest/Scripts/Spline/SplinePoint.cs
```diff
@@ -0,0 +1,178 @@
...
+                {
+                    var beforeNewPoint = parent.GetChild(thisIdx - 1);
+                    newPoint.transform.position = Vector3.Lerp(beforeNewPoint.position, thisSP.transform.position, 0.5f);
+                }
+                else if (parent.childCount > 2)
+                {
+                    newPoint.transform.position = 2f * parent.GetChild(newIdx + 1).position - parent.GetChild(newIdx + 2).position;
+                }
+
+                Selection.activeObject = newPoint;
```
daleeidd, line 113: Needs to register the created object with edit history so undo removes the new spline point instead of just changing the selection back to the previous point:
```csharp
Undo.RegisterCreatedObjectUndo(Selection.activeObject, "Add Crest Spline Point");
```

crest/Assets/Crest/Crest/Scripts/Spline/SplinePoint.cs
```diff
@@ -0,0 +1,178 @@
...
+            }
+
+            label = (thisIdx == parent.childCount - 1 || parent.childCount == 0) ? "Add after (extend)" : "Add after";
+            if (GUILayout.Button(label))
+            {
+                Selection.activeObject = AddSplinePointAfter(parent, thisIdx);
```
daleeidd, line 119: Needs to register the created object with edit history so undo removes the new spline point instead of just changing the selection back to the previous point:
```csharp
Undo.RegisterCreatedObjectUndo(Selection.activeObject, "Add Crest Spline Point");
```

This could be done in the function that creates it. But works here too.

crest/Assets/Crest/Crest/Scripts/Spline/SplinePoint.cs
```diff
@@ -0,0 +1,178 @@
...
+
+                if (thisIdx > 0)
+                {
+                    var beforeNewPoint = parent.GetChild(thisIdx - 1);
+                    newPoint.transform.position = Vector3.Lerp(beforeNewPoint.position, thisSP.transform.position, 0.5f);
+                }
+                else if (parent.childCount > 2)
+                {
+                    newPoint.transform.position = 2f * parent.GetChild(newIdx + 1).position - parent.GetChild(newIdx + 2).position;
+                }
```
daleeidd, line 111: Missing the case for when the current spline point is index zero. Currently, the transform is set to zero. Taken from the AddSplinePointAfter function:

```csharp
                else if (parent.childCount == 2)
                {
                    newPoint.transform.position = parent.GetChild(thisIdx + 1).position + 10f * -Vector3.forward;
                }
```

crest/Assets/Crest/Crest/Scripts/Spline/Spline.cs
```diff
@@ -0,0 +1,124 @@
...
+    {
+        public override void OnInspectorGUI()
+        {
+            base.OnInspectorGUI();
+
+            var targetSpline = target as Spline;
+
+            if (GUILayout.Button("Add point (extend)"))
+            {
+                SplinePointEditor.AddSplinePointAfter(targetSpline.transform);
```
daleeidd, line 97: This needs to be recorded with Undo.RegisterCreatedObjectUndo. See other review comments for example.

daleeidd (changes requested):
This is a first pass review. I haven't looked at the HLSL file yet. I will do another after changes made. The only ones that are "Request changes" are marked with eyeballs.

crest/Assets/Crest/Crest/Scripts/Spline/Spline.cs
```diff
@@ -0,0 +1,116 @@
...
+    /// </summary>
+    [ExecuteAlways]
+    public partial class Spline : MonoBehaviour
+    {
+        [Tooltip("Connect start and end point to close spline into a loop. Requires at least 3 spline points.")]
+        public bool _closed = false;
+
+        public SplinePoint[] SplinePoints => GetComponentsInChildren<SplinePoint>();
+
+        public void OnDrawGizmosSelected()
```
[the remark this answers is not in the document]
daleeidd: Looks like this one was resolved before publishing review.

daleeidd:
> Thanks data reverted now

Thanks.

> Hm it should be the legit diff - if you see something fishy let me know and ill dig deeper.

It's fixed now. I was seeing the OceanDebugGUI.DrawTextureArray and more in the diff which is already in master. I think GH must of got confused.

crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
```diff
@@ -622,15 +668,13 @@ public bool Validate(OceanRenderer ocean, ValidatedHelper.ShowMessage showMessag
         {
             var isValid = true;
 
-            if (_componentsPerOctave == 0)
+            if (GetComponent<Spline.Spline>() != null && _meshForDrawingWaves == null)
```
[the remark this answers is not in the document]
huwb: took a few attempts, was hard to fix without complicating everything. ended up taking an unorthodox approach, let me know if it makes sense.

crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
```diff
@@ -520,6 +552,15 @@ void InitBatches()
                 }
             }
 
+            var splineForWaves = GetComponent<Spline.Spline>();
```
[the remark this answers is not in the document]
huwb: Oh i did not know that. fixed this case but there are others. should we use the Try variant everywhere then?

crest/Assets/Crest/Crest/Scripts/Spline/Spline.cs
```diff
@@ -0,0 +1,116 @@
...
+    /// <summary>
+    /// Simple spline object. Spline points are child gameobjects.
+    /// </summary>
+    [ExecuteAlways]
+    public partial class Spline : MonoBehaviour
+    {
+        [Tooltip("Connect start and end point to close spline into a loop. Requires at least 3 spline points.")]
+        public bool _closed = false;
+
+        public SplinePoint[] SplinePoints => GetComponentsInChildren<SplinePoint>();
```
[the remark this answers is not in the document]
huwb: thanks. i've decided having it there does more harm than good. its only used once at runtime. having it here suggests to users that it is an efficient API appropriate for calling, which it is not. and the extensibility point you made is also valid. i decided to just remove it as its replacement is just as simple. hope this makes sense.

crest/Assets/Crest/Crest/Scripts/Spline/SplinePoint.cs
huwb: good catch, done

crest/Assets/Crest/Crest/Scripts/Spline/SplinePoint.cs
huwb: done

crest/Assets/Crest/Crest/Scripts/Spline/Spline.cs
huwb: done

crest/Assets/Crest/Crest/Scripts/Spline/SplinePoint.cs
huwb: after being confused by the code, i ended up reworking it to unify it with the "addbefore" function, and added comments to the code for each case

crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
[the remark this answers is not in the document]
daleeidd: Perfect solution. Pretty much a *no op* function

crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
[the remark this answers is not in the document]
daleeidd: We should for new code. I forgot about this myself so I will do the same. Been meaning to do a PR for other code. I'll resolve this one.

crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstnerSplineHandling.cs
```diff
@@ -0,0 +1,238 @@
...
+using UnityEngine;
+
+namespace Crest
+{
+    /// <summary>
+    /// Generates mesh suitable for rendering gerstner waves from a spline
+    /// </summary>
+    public static class ShapeGerstnerSplineHandling
+    {
+        public static bool GenerateMeshFromSpline(Spline.Spline spline, Transform transform, int subdivisions, float radius, int smoothingIterations, ref Mesh mesh)
```
daleeidd, line 15: Just making a note here. If this was able to take an array of points, then we could work with any spline tool that users use (or favour) already. The spline tool you made is great for this purpose so maybe in the future depending on feedback.

crest/Assets/Crest/Crest/Scripts/Spline/SplinePoint.cs
```diff
@@ -0,0 +1,223 @@
...
+    [ExecuteAlways]
+    public class SplinePoint : MonoBehaviour
+    {
+#if UNITY_EDITOR
+        Spline _spline;
+
+        void Awake()
+        {
+            _spline = transform.parent != null ? transform.parent.GetComponent<Spline>() : null;
+        }
```
daleeidd, line 22: This is no longer referenced anywhere so it can be removed.

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/AnimWavesGerstnerGeometry.shader
```diff
@@ -0,0 +1,158 @@
...
+        Pass
+        {
+            CGPROGRAM
+            #pragma vertex vert
+            #pragma fragment frag
+			#pragma enable_d3d11_debug_symbols
+
+            #include "UnityCG.cginc"
+
+			#include "../../OceanGlobals.hlsl"
```
daleeidd, line 49: This file could use a format. Whitespace is inconsistent.

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/AnimWavesGerstnerGeometry.shader
```diff
@@ -0,0 +1,158 @@
...
+				const half depth_wt = saturate(2.0 * depth / _AverageWavelength);
+				const float attenuationAmount = _AttenuationInShallows * _RespectShallowWaterAttenuation;
+				wt *= attenuationAmount * depth_wt + (1.0 - attenuationAmount);
+
+				// Feature at front/back
+				wt *= min( input.distToSplineEnd_invNormDistToShoreline.y / _FeatherWaveStart, 1.0 );
+				if( _FeatherFromSplineEnds > 0.0 ) wt *= saturate( input.distToSplineEnd_invNormDistToShoreline.x / _FeatherFromSplineEnds );
+
+				// Quantize wave direction and interpolate waves
+				float axisHeading = atan2( input.axis.y, input.axis.x ) + 2.0 * 3.141592654;
```
daleeidd, line 122: If I negate input.axis here, the waves move away from the spline rather than always towards it. Do you think this would be an option users will want?

crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
```diff
@@ -241,6 +267,12 @@ void UpdateEditorOnly()
                 _spectrum = ScriptableObject.CreateInstance<OceanWaveSpectrum>();
                 _spectrum.name = "Default Waves (auto)";
             }
+
+            // Unassign mesh
+            if (_meshForDrawingWaves != null && GetComponent<Spline.Spline>() == null)
```
daleeidd, line 272: TryGetComponent could be used here.

crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
```diff
@@ -622,15 +667,14 @@ public bool Validate(OceanRenderer ocean, ValidatedHelper.ShowMessage showMessag
         {
             var isValid = true;
 
-            if (_componentsPerOctave == 0)
+            var spline = GetComponent<Spline.Spline>();
```
daleeidd, line 670: TryGetComponent could be used here.

crest/Assets/Crest/Crest/Scripts/Spline/Spline.cs
```diff
@@ -0,0 +1,124 @@
...
+                    "Closed splines must have at least 3 spline points. See the <i>Closed</i> parameter and tooltip. To add a point click the <i>Add point</i> button in the Inspector, or add a child GameObject and attach <i>SplinePoint</i> component to it.",
+                    ValidatedHelper.MessageType.Error, this
+                );
+
+                isValid = false;
+            }
+
+            for (int i = 0; i < transform.childCount; i++)
+            {
+                if (transform.GetChild(i).GetComponent<SplinePoint>() == null)
```
daleeidd, line 69: TryGetComponent could be used here.

daleeidd (changes requested):
Changes look good. Added eyes to anything for *Requested Changes*. Just some minor cleaning things. I marked the places where TryGetComponent could be used. Hopefully I didn't miss any.

crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstnerSplineHandling.cs
huwb: yeah good point. i explored changing it but ran into a couple of not-nice things and didnt land on something better.

UpdateMesh() can be called with a list of points which sort of ticks this box.

the bigger issue is that our spline is embedded/hardcoded into ShapeGerstner. i think an adaptor pattern could work for this, like timeprovider.

the current stuff could be moved to an adaptor pattern without needing data changes i think, so i dont think we close the door to this option if we merge it as is, so guess its fine for now

crest/Assets/Crest/Crest/Scripts/Spline/SplinePoint.cs
huwb: oh nice, done

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/AnimWavesGerstnerGeometry.shader
huwb: i'm 4 attempts in now and each time it seems to get worse! i manually retabbing in vs, format doc in vs, checking in sublime, manually changing tabs to spaces in sublime. will think about how to tackle this

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/AnimWavesGerstnerGeometry.shader
huwb: the wind direction setting on the ShapeGerstner component was intended to set the direction, relative to the local spline direction. however testing now it seems like its broken. do you think i should fix this up, or do you have thoughts how best to achieve this?

crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
huwb: done

crest/Assets/Crest/Crest/Scripts/Spline/Spline.cs
huwb: done

crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
huwb: done

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/AnimWavesGerstnerGeometry.shader
daleeidd: Looks good with latest.

crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstnerSplineHandling.cs
daleeidd: Sounds good. I'll resolve this one.

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/AnimWavesGerstnerGeometry.shader
daleeidd: I grabbed the code from AnimWavesGerstner and it works well enough using wind (further up in the vertex part):
```shader
o.axis = dot(v.axis, _AxisX) * unity_ObjectToWorld._m00_m20 + dot(v.axis, float2(-_AxisX.y, _AxisX.x)) * unity_ObjectToWorld._m02_m22;
```

Wind is kind of a weird name for this, but wind seems to just be a generic direction property anyway so I think go ahead with it.

huwb: oh yeah that works. ive done that now.

i proposed renaming the 'wind' setting to *Wave Direction Heading Angle*. bit of a mouthful but good to be explicit i guess, if it makes sense to you? lmk

daleeidd: Sounds good

daleeidd (approved):
Tested to be working on MacOS and Windows in editor and build. Really cool feature. And I suspect highly anticipated too.

Gizmos can still be a little difficult since they are not really visible under the ocean surface. I think clamping the gizmos to the surface is a good next step but we would need to take in elevations into account. So something for the future.

I've also made #735 to disable ocean surface pickability which might help with selecting points.

huwb:
Yeah the selection stuff is a pain. It could be any surfaces in front of the spline points, not just water..

I think if I set the gizmos to be 2d, then the icons show clearly through things and are clickable. However these are not the default settings. Did you try these settings?

An option might be to detect what settings are being used and add an info box to the inspector to recommend changing them, might have a look later

daleeidd:
> I think if I set the gizmos to be 2d, then the icons show clearly through things and are clickable. However these are not the default settings. Did you try these settings?

You're right. It's a shame the 2D cannot be forced for these gizmos. Also, I just realised 3D ones can be clicked through the ocean surface; it's just visibility is an issue.

Another option is to use [Handles](https://docs.unity3d.com/ScriptReference/Handles.html) with Gizmos. These always behave like 2D. But I would only use these when the spline or points are selected so that the benefit of 3D gizmos is still there. It looks like Handles don't receive mouse input so they shouldn't be used alone. Just to enhance visibility. Example:

```csharp
protected virtual void OnSceneGUI()
{
    if (Event.current.type == EventType.Repaint)
    {
        Transform transform = ((SplinePoint)target).transform;
        Handles.color = Color.red;
        Handles.DotHandleCap(0, transform.position, transform.rotation, 1f, EventType.Repaint);
    }
}
```

huwb:
Thats weird - for me it all works dandy. I can see and select points below the terrain/water.

image

what version of unity are you testing on? i'm on 2019.4.16. are you testing on Mac?

daleeidd:
> Thats weird - for me it all works dandy. I can see and select points below the terrain/water.

It's the same for me with 2D. I can see and click them.

huwb:
Ah i misread, thanks.

i struggled to get a match between drawicon and the handle. i'm not able to match their sizes - the icon can be all different sizes depending on the view settings and i dont seem to be able to get the size from code.

i tried abandoning icons and drawing a cube as both a handle and a gizmo, but this arbitrarily killed mouse picking which was annoying.

after messing around i struck upon a simple solution that mostly works. Gizmos.DrawWireCube *is* draw through surfaces, and is also pickable. so i draw both a cube and a wire cube and together they unite to form something that is almost good.

a minor not good part is clicking a wireframe is less comfortable than clicking a solid cube would be.

the major not good part is that, unlike with a 2d icon, picking the point respects scene depth order, so you have to make sure you set up the scene to be non pickable (annoying), or patiently click 2 or 3 times until the point gets selected (annoying).

running out of steam tbh and tempted to go with this solution. i've pushed it, but it can easily be reverted.

daleeidd:
> i struggled to get a match between drawicon and the handle. i'm not able to match their sizes - the icon can be all different sizes depending on the view settings and i dont seem to be able to get the size from code.

Ah yes... looks like some have resorted to reflection to get it :(

> i tried abandoning icons and drawing a cube as both a handle and a gizmo, but this arbitrarily killed mouse picking which was annoying.

I haven't experienced this :S Hmmm...

> Gizmos.DrawWireCube is drawn through surfaces (weirdly)

When 3D icons are activated, the wire cube is obscured by the ocean surface for me still.

> the major not good part is that, unlike with a 2d icon, picking the point using the wire cube respects scene depth order, so you have to make sure you set up the scene to be non pickable (annoying), or patiently click 2 or 3 times until the point gets selected (annoying).

I am thinking that the previous 2D icons were better for that reason (apologies). Even when I have 3D icons selected, they are still visible, but barely (not enough really). And I can select them through any surface. Perhaps we should go with this until a better solution is found.

There are some solutions which are more involved. Like listening to the mouse up event and checking if ray intersects a sphere (our spline point). Or implementing Handles with mouse support (still not sure if this handles pickability or not).

> your change ot make the ocean non-pickable may make the depth order a minor issue, or non issue for most cases.

That is true. I guess whether that PR is worthwhile merging depends on which option we go with.

daleeidd:
New icons work much better. Nice find. Looks ready to merge.

huwb:
Thanks. Merging!

PR #733, opened 2021-01-21
Combine changes

huwb:
Two changes related to the combine pass..

Firstly, added target pragma which apparently fixes an issue on Quest.

The second changes defaults to not doing ping-pong passes for the combine. The ping pong approach uses a graphics shader and copies results backwards and forwards from an aux RT. This is here because early DX10 GPUs missed a feature we need to run the combine as a compute shader (typed loads). So we added two paths, and defaulted to the "safe" one.

I feel we should change the default to the efficient compute path now, given that the old gpus are something like 15years old now, and see how many people report issues. The symptom is that it silently fails to combine the cascades and the ocean will appear flat. If we don't hear issues perhaps we can consider switching completely over to the compute path.

### DIFF
--- a/crest/Assets/Crest/Crest/Shaders/Resources/ShapeCombine.shader
+++ b/crest/Assets/Crest/Crest/Shaders/Resources/ShapeCombine.shader
@@ -20,6 +20,7 @@ Shader "Hidden/Crest/Simulation/Combine Animated Wave LODs"
 			CGPROGRAM
 			#pragma vertex Vert
 			#pragma fragment Frag
+			#pragma target 3.5
 
 			#pragma multi_compile __ CREST_DYNAMIC_WAVE_SIM_ON_INTERNAL
 			#pragma multi_compile __ CREST_FLOW_ON_INTERNAL
--- a/crest/Assets/Crest/Crest/Shaders/Resources/ShapeCombine.shader
+++ b/crest/Assets/Crest/Crest/Shaders/Resources/ShapeCombine.shader
@@ -151,6 +151,7 @@ Shader "Hidden/Crest/Simulation/Combine Animated Wave LODs"
 			CGPROGRAM
 			#pragma vertex Vert
 			#pragma fragment Frag
+			#pragma target 3.5
 
 			#include "UnityCG.cginc"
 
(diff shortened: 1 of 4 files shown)


daleeidd (approved):
Tested. Working on editor and standalone for Windows. Not working (shapecombine) for MacBook Pro 2014 iGPU. I don't have a more modern Mac to test on unfortunately. I could do another PR which makes the option part of the AnimWavesSimSetting if you plan on keeping the code around for longer?

huwb:
D'oh - not working on a 2014 macbook pro - doesnt sound promising. this might be a nonstarter. thanks for testing!

i should make sure the setting is documented in the performance notes at least, so its at least on the radar.

daleeidd:
> this might be a nonstarter.

That's unfortunate. Latest MacOS supports late 2013 MacBooks. Should shader level 3.5 be a general requirement?

huwb:
Apparently 3.5 is required for texture arrays, so yes probably :). What are you thinking, go wider with the pragma?

daleeidd:
> Apparently 3.5 is required for texture arrays

It's also a requirement for instancing... yet Unity's guides on instancing have a target of 3.0 which is confusing. Furthermore, the generated URP shader has a target of 2.0 which is lower than what we have currently. HDRP has a target of 4.5.

I'm thinking we should since we already have a case where not doing so has caused an issue.

crest/Assets/Crest/Crest/Shaders/Resources/ShapeCombine.shader
```diff
@@ -20,6 +20,7 @@ Shader "Hidden/Crest/Simulation/Combine Animated Wave LODs"
 			CGPROGRAM
 			#pragma vertex Vert
 			#pragma fragment Frag
+			#pragma target 3.5
```
daleeidd, line 23: Does this need to be added to the other pass below?

daleeidd (approved):

crest/Assets/Crest/Crest/Shaders/Resources/ShapeCombine.shader
huwb: Ah yes good point

huwb: done

huwb:
i expanded the pragma to the other pass in that same shader. we may have other graphics shaders that do compute-y type stuff and need the pragma, i've not done an exhaustive search for now as i dont know quite what the criteria should be. maybe its every graphics shader that uses a texture array - like our ocean shader? i feel a bit weird copy pasting this around without knowing 100% wehre its needed, and not having a test case for it.

i set ping-pong back to true by default, and moved it to the anim wave settings as you suggested.

daleeidd:
I agree. It is kind of murky. Everything looks good. I'll merge this in.

Issue #705, opened 2020-11-28
Foam fade is frame rate dependent


the-mr-matt:
**Describe the bug**
The rate at which foam fades in my build appears to be dependent on the frame rate. When I enable VSync, frame rate is approx 60 fps. In this case, foam fades quickly. When VSync is disabled frame rate is approx 200-300. In this case, foam fades very slowly, and appears almost permanent.

Note, in the video below there are no cuts, the only thing that changed is the frame rate.

**Screenshots / video**
https://www.dropbox.com/s/0x22ww89zw1rqr8/Crest%20Foam%20Frame%20Rate%20Issue.mp4?dl=0

**Log**
[Player.log](https://github.com/crest-ocean/crest/files/5611130/Player.log)

**Versions**
HDRP 8.2.0, Unity 2020.1.13f1

**To Reproduce**
Appears to be reproducible by any method that alters frame rate. Both VSync and Application.targetFrameRate worked for me.

**Platform**
Windows standalone

**Hardware**
RTX 2080Ti
Ryzen 3700X
32gb RAM
SSD

huwb:
I wonder if the problem is the precision of the foam data - its 8bit by default and maybe at high frame rates the delta is less than 1<256 . i'd have to think about how to solve this kind of thing.

We have the precision exposed in the foam settings - could you try creating a sim settings foam asset and attaching it to the ocean component, and increasing the foam texture precision to R16?

the-mr-matt:
I played with different configurations of precision, compression and mip maps but none seemed to have much affect.

holdingjason:
Anything to do with this one?

https://github.com/crest-ocean/crest/issues/408

huwb:
I have a test/experimental PR which locks simulation rate, which should help with this. feedback welcome

https://github.com/crest-ocean/crest/pull/734

huwb:
Will be fixed in 4.9

daleeidd:
4.9 has been released which solves this issue. The simulation may produce different results so tweaking may be required. Let us know if it can be improved.

Issue #522, opened 2020-05-26
Dynamic Waves freak out on low simsteps


klauskobald:
This is between a Bug Report and a Feature request!
The problem is, that on flat water the objectinteraction creates tiny fluctuations which build up to cracy freaking out waves.
I found out, that this can be fixed by simply damping very tiny velocities like so:

in the UpdateDynWaves.compute shader change this line
`   vtp *= 1.0 - min(1.0, _Damping * dt);
`
into
`
if(vtp<0.0005) vtp*=0.8;`
`else vtp *= 1.0 - min(1.0, _Damping * dt);`

The constants 0.0005 and 0.8 could be discussed and maybe the should be exposed as shader parameters. The shown values work even for these extrem settings:
damping=0 
courant=3 
simsteps=1

(Sorry I have no time for finding out how pull request work)

klauskobald:
ok - I think, we should forget my above suggestions! They create another problem: sustaining water bumps.

huwb:
Yeah was wondering about that. What about applying that friction when both velocity is low, and wave height is small? Not sure, just a thought.

klauskobald:
yes, that would help for sure. I did however not figure out, where the height value is stored.

huwb:
It's `ftp` . f is the height and tp means `time positive` - one time step ahead.

huwb:
I have a test/experimental PR which locks simulation rate, which should help with this. feedback welcome

https://github.com/crest-ocean/crest/pull/734

huwb:
Will be fixed in 4.9

daleeidd:
Thanks for reporting this. 4.9 has been released which solves this issue. The simulation may produce different results so tweaking may be required. Let us know if the issue persists.

Issue #408, opened 2020-01-26
Dynamic Waves Appear to be tied to FPS


holdingjason:
**Describe the bug**
The amount of dynamic wave distortion you get ala strength seems to be tied to frames per second.  Decrease the frames and you get less distortion, increase frames more (not sure if the increase is ultimately capped by whatever your settings are and that you only see the correct distortion at a very high framerate ala 100+ etc)

**Screenshots / video**
These are very helpful to understand and diagnose issues.

**Log**
What is in the editor log? Clear before running and then provide any errors/warnings that appear. The log frequently provides the key hints.

**Versions**
Which render pipeline (built-in/LWRP/HDRP)? Which version of LWRP/HDRP? Which version of unity (including minor, i.e. 2019.1.0f2)? 

**To Reproduce**
How can the developers reproduce this bug? Are you able to reproduce the issue in the example content/scenes?  Yes.

Please include steps to reproduce the behaviour, with the example content ideally so we can follow the steps on our side.
Using the boat demo you can easily repro this by just creating a script that changes the overall frame rate, so spit out a bunch of debug statements in a loop for instance but does not matter anything to control the current FPS.

**Platform**
Editor or Build.  PC

**Hardware**
ATI Radeon RX 580

High FPS 88+
![image](https://user-images.githubusercontent.com/42001313/73138355-856a0f00-4016-11ea-9612-6d6539ac0702.png)

Low FPS 30 or so.
![image](https://user-images.githubusercontent.com/42001313/73138329-3cb25600-4016-11ea-8c46-a8bcfffbfa5b.png)

holdingjason:
BTW if you look at the two shots one thing that jumps out is the sim steps are different 3 to 2.

huwb:
Thanks for the comparison, really useful. I'll let you know when we have news.

Btw the step count will be different as a result of the framerate being different, this is expected for fixed timesteps.

holdingjason:
Thanks @huwb let me know if I can help in anyway since this is a pretty nasty one so happy to help.  Still might be something I am doing but since I was able to see it in the demo hoping not :)

iichiversii:
Creating bow and stern dynamic waves is something I cant setup, can someone make a video tutorial on this?

moosichu:
@iichiversii I would ask on the Discord server for questions to that kind of thing: https://discord.gg/g7GpjDC

Epimolophant:
Hi! Any news on this?

huwb:
[...]

PR #734, opened 2021-01-22
Force sims to run at fixed rate

huwb:
**Status**: Test / for feedback

Our sims (foam and dyn waves) run at variable update rates, which works surprisingly badly:

#705 , #522 , #408

Part of the issue may be our integration scheme, but other schemes are more expensive and more involved to implement. We apply dissipation/friction using a simple lerp model, i tried the other approach (exp) and it did not help. I think i've just learnt the hard/slow way why physics engines run at a fixed rate.

This PR forces the foam sim to run at 30hz and the dynamic wave sim to run at 60hz. If the foam sim is at least 33.33ms behind the current unity time, it does a step, then the result is used in the render. Doing rigid steps like this may be susceptable to visible jitter, and if so might require an "interpolation" scheme where we always simulate past the current time, and then interpolate current and previous sim to get the result. This would add extra cost however, and i did not notice jittering in tests so far, so maybe this is good enough. Also if users are using v-sync and rendering at 60hz or 30hz, the result should look perfect.

Changing the dt changes the behaviour of these sims (dyn waves may grow too large, foam may be too much), which is exactly the point of this PR. Once retweaked to suit the new dt, the behaviour should be consistent across any render framerate. I've made a stab at retweaking dyn wave strengths in the example content, as a first stab, in this PR.


daleeidd:
Latest commit works great on Mac. Tested in editor and build. Not sure if you are finished yet, but let me know when you want a review.

huwb:
Thanks, glad to hear that.

I'm now messing with how the interaction forces work to fix issues with generating interactions into multiple sims which is looking promising, but may not apply directly to the boat interaction shaders we currently use, so I should probably move this to a later commit.

Will request review when I've tidied things up.

huwb:
Closing to be replaced by #778

PR #735, opened 2021-01-23
Disable pickability for ocean patches

daleeidd:
This is to make it easier to select things through the ocean surface. Currently, if the ocean surface is clicked, then the OceanRenderer is selected; and if it is clicked again, the ocean patch is selected. But if something important (like a spline point) is beneath the surface, then bypassing the surface is helpful.

The Unity API that I am using here uses the scene visibility and pickability UI on the left margin of the hierarchy so it can be overridden. [More information here](https://docs.unity3d.com/Manual/ScenePicking.html).

I think we could limit this to only the global ocean surface in the future since small patches being clickable might be okay.

Let me know what you think.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/OceanBuilder.cs
+++ b/crest/Assets/Crest/Crest/Scripts/OceanBuilder.cs
@@ -171,6 +171,14 @@ namespace Crest
                 CreateLOD(ocean, tiles, root.transform, i, lodCount, meshInsts, meshBounds, lodDataResolution, geoDownSampleFactor, oceanLayer);
             }
 
+#if UNITY_EDITOR
+            if (OceanRenderer.Instance._hideOceanTileGameObjects)
+            {
+                // This should cover patches too, but doesn't seem to...
+                UnityEditor.SceneVisibilityManager.instance.DisablePicking(root, includeDescendants: true);
+            }
+#endif
+
 #if PROFILE_CONSTRUCTION
             sw.Stop();
             Debug.Log( "Finished generating " + lodCount.ToString() + " LODs, time: " + (1000.0*sw.Elapsed.TotalSeconds).ToString(".000") + "ms" );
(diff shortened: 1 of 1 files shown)


huwb (approved):
I haven't tested it but it looks/sounds good, nice one.

At some point we should test on both 2020 and 2019. Which did you test on so far?

huwb:
I gave it a quick test on 2019.4.16 and it works well. the only slightly odd thing is it shows the hand icon in the Hierarchy window next to the object as if it *is* pickable (not the struck-through non-pickable icon). 

i suppose another consideration is that it overrides the pickable state for the ocean gameobject - the user may want the ocean pickable and it wont respect their choice when they next open the scene. an alternative may be to set the pickable state in the Reset() function which roughly gets called when the component is added, and may allow the user to override it later. im not sure if this is necessary tbh but just throwing it out there.

daleeidd:
> At some point we should test on both 2020 and 2019. Which did you test on so far?

Only 2019.

> I gave it a quick test on 2019.4.16 and it works well. the only slightly odd thing is it shows the hand icon in the Hierarchy window next to the object as if it _is_ pickable (not the struck-through non-pickable icon). feels trivial and not a blocker, but just mentioning in case its easily addressed.
> 
> i suppose another consideration is that it overrides the pickable state for the ocean gameobject - the user may want the ocean pickable and it wont respect their choice when they next open the scene.

Thanks for the feedback. This was mainly to address spline point picking issues, but you solved the issue in the spline PR. So we could close this. The user can disable picking themselves if they really want it and know how to use the feature. What do you think?

> an alternative may be to set the pickable state in the Reset() function which roughly gets called when the component is added, and may allow the user to override it later. im not sure if this is necessary tbh but just throwing it out there.

That's a good idea. We would need a property to manage the state.

huwb:
Yeah i'd also vote for closing this in light of recent changes. Was a good thing to look into though!

daleeidd:
Replaced by #1066

PR #736, opened 2021-01-24
Make time provider settable from code

huwb:
**Status**: Test / for feedback

Stored as an ITimeProvider to support non-monobehaviour objects


daleeidd (approved):
I haven't tested it, but it looks simple and good.

huwb:
Thanks

PR #737, opened 2021-01-24
Inline settings editing

huwb:
**Status**: Test. Not needed for next release.

Tried to import the stuff from cinemachine that has "create" button for SO fields, and displays an inline editor.

It worked well i think, but will mean our settings fields need to go either above or below the rest of the OceanRenderer fields. I guess above..

The HelpURL attribute wont function anymore but we could just add a help button.

The PredicatedField also wont function, we could either expand the whole thing to allow enabling/disabling the feature as well as editing settings, but this would make it less general/resuable for other components. An alternative would be to put this predicate functionality into it (optional) which is probably the way forward.


huwb:
every time im working on something now im wishing this was merged in!

crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
```diff
@@ -262,10 +262,15 @@ public void CrestUpdate(CommandBuffer buf)
 #if UNITY_EDITOR
         void UpdateEditorOnly()
         {
-            if (_spectrum == null)
+            if (_spectrum != null)
+            {
+                _activeSpectrum = _spectrum;
+            }
+            // TODO - dont do this. could init to default spectrum asset in Reset() function
```
daleeidd, line 269: Is this something that should be resolved before merging?

daleeidd (changes requested):
I have merged in the dev branches. There is a TODO which I am not sure you want resolved before merging. Also some test data that needs reverting which I am happy to do.

crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
huwb: yeah i think we need to kill that. we have a "Create" button now to create spectrum assets. so allowing to edit a dynamically created pseudo-asset is evil. do you want to do the honours or shall i?

huwb:
nice! while searching for "todo" in the Files view theres a couple of others, probably inherited from other peoples code. any we can remove (or act upon)?

crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
daleeidd: I don't think users can edit the dynamic one since _activeSpectrum is not serialised. So you are planning on removing the dynamic spectrum completely? Honours is yours :)

daleeidd:
> nice! while searching for "todo" in the Files view theres a couple of others, probably inherited from other peoples code. any we can remove (or act upon)?

I added two which can be ignored. One is based off a comment from the thread but I am not sure what the author was thinking. Left it there so if there is an issue it might hint at something. The other is just removing LINQ. I can do that one after the one flagged issue is resolved.

crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
huwb: Hm i've lost a handle on what i had in mind here. I think i found the dynamically created settings were editable, but as you pointed out .. they're not :). I tested now and the behaviour is fine so lets nuke this todo. ill do this now.

huwb:
From my point of view this ready to merge in. Could you finalise any bits from your side and i'm happy for you to merge it, or I can do it too. Nice one!

daleeidd (approved):

daleeidd:
Thanks. Merged. Great find BTW.

huwb:
This is so good - I'm gonna grab latest and play with it some more!

PR #738, opened 2021-01-24
Add foam online help button link

daleeidd:
Adds a help button link. There should be a way to create an abstraction to reduce boilerplate similar to validated which can be done for release after this one.


huwb (approved):
Looks good

PR #741, opened 2021-01-28
Convert inline SO editor to attribute

daleeidd:
Opening this PR for discussion and possible merging in #737.

Converts the inline _ScriptableObject_ editor to an attribute called _EmbeddedField_. Works for both _OceanRenderer_ and _ShapeGerstner_.

A little bit messy since it was hacked in. Attributes don't support generics yet so I had to dropped generics :(

PredicatedField does not work with this either.

Labels and such are inferred from what is available from the _SerialisedProperty_, but I can add some parameters to the attribute so it can be customised.

Let me know what you think.

screen


huwb:
Is there a downside to this? I'm a bit tired but I scanned the code and it looks good to me! Removing the generic seemed to have little downside that I could see.

Would it support a tooltip on the field do you think? Not critical but just a thought.

daleeidd:
> Is there a downside to this? I'm a bit tired but I scanned the code and it looks good to me! Removing the generic seemed to have little downside that I could see.

Not that I could see or experience.
 
> Would it support a tooltip on the field do you think? Not critical but just a thought.

Tooltips do work.

daleeidd:
I have fixed a few things and did a cleanup. I have also created a separate PR (linked above) which includes predicated field support.

huwb:
i havent had a chance to test it, but im a big thumbs up for this, looks awesome. since its a cross-dev-branch PR feel free to merge it in now, and you can take it all the way to master if you're confident in it, otherwise feel free to point me to any questions or things you'd like me to test

daleeidd:
Thanks. Merged into #737 to be evaluated there.

PR #744, opened 2021-01-30
Fix null ref ex when adding ocean comp to new GO

huwb:
**Status**: Ready to merge

Repro steps are to make a new scene, make a new GO, and add OceanRenderer component. This resulted in a spew of null ref exceptions for me, on 2019.4.16. This is following the setup instructions we ship with crest.

This change defers creation of the surface, and minor changes to avoid other systems trying to get the Root transform before it is created.

Not excited to merge this at the last minute just before release.. however its holding up to my tests so far. I tested with different paths of creating/loading ocean scenes, and tested standalone build.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
@@ -266,7 +266,7 @@ namespace Crest
             // Shader needs sea level to determine water depth. Ocean instance might not be available in prefabs.
             var centerPoint = Vector3.zero;
             centerPoint.y = OceanRenderer.Instance != null
-                ? OceanRenderer.Instance.Root.position.y : transform.position.y;
+                ? OceanRenderer.Instance.SeaLevel : transform.position.y;
 
             _copyDepthMaterial.SetVector("_OceanCenterPosWorld", centerPoint);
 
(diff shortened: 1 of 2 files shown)


daleeidd:
I cannot reproduce. I am using master and I create a new GO using the hierarchy window and add the OceanRenderer component, but no null exceptions.

huwb:
Ha I've been wondering if it's something local. I did make another new scene, and restarted unity, 100% repro. Tried also disabling scene/domain reload but no difference.

Maybe the animated materials option, do you have that enabled? Trying to think of what else

daleeidd:
Also tested on Mac and Windows. Also HDRP.

> Maybe the animated materials option, do you have that enabled?

I do.

So you create a new scene, and it only has a camera and light? Then the OceanRenderer and it throws nulls? I also tried with some of our components that would access Root existing before the OceanRenderer but still no issues.

daleeidd:
> Trying to think of what else

Anything in the project settings should show up in a git diff.

Any prefabs involved? Are you able to reproduce in another project?

huwb:
If i remote access another pc it happens 100% of the time on all three of our projects. I cant repro it locally. I guess i'll put this PR on ice until we hear about it in the wild. Will add an assert to catch it more explicitly to make sure it gets reported.

PR #745, opened 2021-01-30
Add prefabs to make setup easier

huwb:
**Status**: Early review to collect thoughts

I want to start prefabing things to make setup straightforward for new users.

Added 4 ocean prefabs with different quality levels.

Potential next steps:

* We could have different materials where we suggest settings for different quality levels, assigned to these prefabs.
* We could add prefab variants of these for underwater (as they need different material options).
* We could add wave prefabs for a few different wave conditions

### DIFF
--- a/crest/Assets/Crest/Crest/Prefabs.meta
+++ b/crest/Assets/Crest/Crest/Prefabs.meta
@@ -1,0 +1,8 @@
+fileFormatVersion: 2
+guid: c9905498c7b29184b978859fa922b5f8
+folderAsset: yes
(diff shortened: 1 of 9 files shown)


daleeidd:
All good ideas. Prefabs are certainly have the least friction for getting started quickly.

> We could add prefab variants of these for underwater (as they need different material options).

We could also manage underwater keyword and face culling via scripting? We wouldn't need to strip the variant if we couldn't reliably do it as it would be minor cost. It might need #629 which I can finish.

PR #746, opened 2021-02-01
Expand tile epsilon to cover new crack bug

huwb:
Latest post in issue #665 . This change expands tile epsilon to cover new crack bug. I didnt see any negative side effects of this increased epsilon and it's still quite small. The other repro scenes look fine to me with it as well (tested in editor).

### DIFF
--- a/crest/Assets/Crest/Crest/Shaders/Ocean.shader
+++ b/crest/Assets/Crest/Crest/Shaders/Ocean.shader
@@ -317,7 +317,7 @@ Shader "Crest/Ocean"
 				// Scale up by small "epsilon" to solve numerical issues. Expand slightly about tile center.
 				// :OceanGridPrecisionErrors
 				const float2 tileCenterXZ = UNITY_MATRIX_M._m03_m23;
-				o.worldPos.xz = lerp( tileCenterXZ, o.worldPos.xz, 1.0004 );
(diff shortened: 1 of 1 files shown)


daleeidd:
Looks good for the particular case. With these fixes I can reproduce if I go out further (like 30,000). We should determine a maximum distance from zero we are willing to handle. We already said 80,000 units is too far for HDRP at this point.

If the transform reaches 100,000 units from zeros, Unity warns about precision. But when we go almost that far out, the ocean moves around a little with the camera so that certainly too far for Crest at this time. But for completion, it took about 1.01 in scaling to solve a line.

If we go around 40,000 units out I can reproduce a line that requires around 1.00111 in scaling to go away.

huwb:
> We should determine a maximum distance from zero we are willing to handle

Yeah i've been waiting for a case that was unsolvable really.. for example if we have to apply too large a fudge that we're not comfortable with. I think i'm ok with a tile dilation up to ~1%, and what you proposed is 0.1%, so I think its very reasonable and committed it (thanks).

I could simply bump it to 1% now but i would rather we get pushed up to it, as larger tiles means more rendering and the risk of flickering in overlaps. If we get close to 1% or are needing to exceed it, then i think we should stop there and recommend world design/data fixes.

PR #747, opened 2021-02-02
Rename Spline to 'Crest Spline'

huwb:
**Status**: Proposal for discussion

It was suggested to me that the name 'Spline' is not very considerate to other splines that people might have. In the Add Component menu it just shows as 'Spline'. Prefixing with Crest seems like a good idea?

image


daleeidd:
I'm generally in favour of using Crest as a prefix for classes. But we don't have to as there are a few options:

1. We can set the name in the component menu by adding `AddComponentMenu("Scripts/Crest.Spline/Crest Spline")` to the spline class.
2. Alternatively, we can [assign custom icons](https://docs.unity3d.com/Manual/AssigningIcons.html) for those menus to help separate Crest scripts from others.

huwb:
thats cool! lazy question but do you still have the code to add that logo lying around? feel free to merge into this branch if you do.

agreed on renaming in the menu and adding the icon, good stuff

daleeidd:
Should I do the menu and icon thing for all of our components? I'll do them as a separate branch if so.

huwb:
Yeah that would be awesome. Thanks!!

daleeidd:
I'll close this as the issue has been addressed with #789.

PR #748, opened 2021-02-04
Inline settings editing attribute with predicate field support

daleeidd:
Open this PR for discussing and merging into #741 and finally into #737

Idea and code taken from:
https://forum.unity.com/threads/drawing-a-field-using-multiple-property-drawers.479377/

Basically allows attribute stacking. Writing the GUI code is a little different as you are not suppose to reset the GUI code so the next attribute in the stack receives the changes. This allows our predicated field and embedded asset field to work together.

I haven't fully evaluated the code yet to see if there are any drawbacks. But everything appears to be working well. It also simplifies the code as we don't need to define our own property drawers.

Usage:
```csharp
[PredicatedField("_createFoamSim", order = 0), EmbeddedField(order = 1)]
public SimSettingsFoam _simSettingsFoam;
```

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Helpers/Attributes/EmbeddedFieldAttribute.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Helpers/Attributes/EmbeddedFieldAttribute.cs
@@ -11,7 +11,7 @@ namespace Crest
     using UnityEditor;
 #endif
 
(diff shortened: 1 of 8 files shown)


huwb:
i dont see any issue with this from looking at the code, but im not the most familiar with it

daleeidd:
Merged to be evaluated in #741

Issue #617, opened 2020-08-15
Water at horizon edge is black


andrew-sidereal:
**Describe the bug**
It's the right & left edges which show this. If I rotate camera, it rotates the black bands with it.

**Screenshots / video**
1. https://i.imgur.com/QlMkZfV.jpg
2. https://i.imgur.com/dCLTKHG.png

**Log**
https://i.imgur.com/Zua3Wan.png

**Versions**
built-in rendering pipeline, v2019.4.4

**To Reproduce**
Basic ocean setup with some waves.

**Platform**
 - in editor

huwb:
That's pretty weird. Could you experiment with the far clip plane setting on the camera?

Could you posted screenshots showing the ocean renderer component settings and the camera settings?

andrew-sidereal:
Thanks for taking the time to look into this. I took bunch o screenshots of the camera settings with different far plane values, as well as the ocean game object settings. You can see them all here: https://imgur.com/a/5DPo9sd

Please let me know if there's anything else I can do to help solve this. Thank you!

daleeidd:
Do you know how old your version of Crest is? Are you grabbing it from the master branch?

Does it occur in the main example scene? Have you tried to narrow it down by disabling Enviro and Map Magic or anything else in the scene?

andrew-sidereal:
I got Crest from the master branch either Aug 12th or Aug 14th.

Thank you for helping me think through this. The Crest demo scene works wonderfully.

I've narrowed down my issue to the Enviro Sky Manager. Here are a couple before/after shots: https://imgur.com/a/sBR2WAQ

Are there steps required to properly integrate this with Enviro? I'd appreciate any help you can offer.

Thank you,
Andrew

daleeidd:
Did you follow the integration guide provided by Enviro?

We have a copy on our wiki. I am pretty sure it is still up-to-date:
https://github.com/crest-ocean/crest/wiki/How-to-use-Crest-with-Enviro

andrew-sidereal:
I've completed that guide and it enables Enviro fog. It was clear and easy
to follow -- thank you. However, the same issue still exists. I see this
error in the console: "There is no skybox set in the lighting settings
window. Ocean color will be incorrect." Any guidance on how to fix this
when using Enviro?

Here are a number of shots from night-time, when it's most dramatic. It's
like the square around the player/main-cam is properly shaded, but beyond
that it's not.
https://imgur.com/a/x0bJu9I

[...]

PR #749, opened 2021-02-09
Fix: always update cache components in editor

daleeidd:
Fix: #617 (partially)

Fixes an issue where the properties (max terrain height and resolution) on the depth cache or its transform (scale) are changed, the components (camera and textures) are not updated when in the editor when they should be. Users would have to enter and exit play mode which would cause confusion and possibly confusing bug reports. The update method is public so users can call during runtime if they need to change properties on the depth cache.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
@@ -542,7 +542,7 @@ namespace Crest
                     isValid = false;
                 }
 
-                if (transform.lossyScale.x != transform.lossyScale.z)
+                if (!Mathf.Approximately(transform.lossyScale.x, transform.lossyScale.z))
                 {
                     showMessage
                     (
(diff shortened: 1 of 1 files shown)


huwb (approved):
Thanks, good shout.

I wondered if the new code could/should be used during initial setup of the depth cache so that there is only one set of code to set the transform, camera params, etc. However not sure if this would work and I'm fine with merging this.

daleeidd:
Sounds good. Done. I have added an updateComponents param to PopulateCache. If true, then components are updated in InitObjects.

I have also added validation to check if the components are outdated and recommend user populates the cache.

The only thing I haven't handled is the scale. If we use the island scene as an example, and scale the parent object of both the cache and island, if the cache is baked, the cache will scale correctly. But if the cache is realtime, then it will distort. It is possible to correct this, but is tricky when the cache itself is rotated.

huwb:
Could we catch the scale issue with validation?

daleeidd:
Sure. I am now showing a validation error if the lossy scale is not uniform for X and Z when set to realtime. Rotating the cache seems fine though.

Also added switching the quad texture when the cache type is changed.

huwb:
Awesome.

Probably want to use Mathf.ApproximatelyEqual when checking the scale as doing exact comparison would be liable to fail.

daleeidd:
Done.

huwb:
Looks good

daleeidd:
Thanks. Merged.

PR #757, opened 2021-02-17
Fix broken RayTraceHelper

daleeidd:
*_valid* is no longer set to true in *Init* preventing it from running *Trace* fully. I have removed *_valid* completely.


huwb (approved):
Weird. thanks!

PR #760, opened 2021-02-17
Validate if ocean feature enabled, material keyword enabled

huwb:
For each ocean input, checks that feature is enabled on ocean component, and checks that feature is enabled on ocean material.

Hopefully this reduces setup errors.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/LodData/Shadows/LodDataMgrShadow.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/Shadows/LodDataMgrShadow.cs
@@ -22,6 +22,9 @@ namespace Crest
         protected override GraphicsFormat RequestedTextureFormat => GraphicsFormat.R8G8_UNorm;
         protected override bool NeedToReadWriteTextureData { get { return true; } }
 
+        internal const string MATERIAL_KEYWORD = "_SHADOWS_ON";
+        internal const string MATERIAL_KEYWORD_MISSING_ERROR = "Shadowing must be enabled on the ocean material. Tick the <i>Shadowing</i> option in the <i>Scattering</i> parameter section on the material currently assigned to the OceanRenderer component.";
+
         public static bool s_processData = true;
 
         Light _mainLight;
@@ -91,7 +94,7 @@ namespace Crest
             }
 
 #if UNITY_EDITOR
-            if (!OceanRenderer.Instance.OceanMaterial.IsKeywordEnabled("_SHADOWS_ON"))
+            if (!OceanRenderer.Instance.OceanMaterial.IsKeywordEnabled(MATERIAL_KEYWORD))
             {
                 Debug.LogWarning("Shadowing is not enabled on the current ocean material and will not be visible.", _ocean);
             }
--- a/crest/Assets/Crest/Crest/Scripts/LodData/Shadows/LodDataMgrShadow.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/Shadows/LodDataMgrShadow.cs
@@ -96,7 +96,7 @@ namespace Crest
 #if UNITY_EDITOR
             if (!OceanRenderer.Instance.OceanMaterial.IsKeywordEnabled(MATERIAL_KEYWORD))
             {
-                Debug.LogWarning("Shadowing is not enabled on the current ocean material and will not be visible.", _ocean);
+                Debug.LogWarning(MATERIAL_KEYWORD_MISSING_ERROR, _ocean);
             }
 #endif
         }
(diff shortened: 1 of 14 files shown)


daleeidd:
Good idea.

We could define these strings as `internal const` on the *LodDataMgr\** class. They can be referenced like static variables. This would eliminate duplication in the inputs and ocean renderer. For example:

In *LodDataMgrShadow*:
```csharp
internal const string FeatureDisabledErrorMessage = "<i>Create Shadow Data</i>...";
```

In *RegisterShadowInput*:
```csharp
protected override string FeatureDisabledErrorMessage => LodDataMgrShadow.FeatureDisabledErrorMessage;
```

In *OceanRenderer*:
```csharp
if (ocean.CreateShadowData && !ocean.OceanMaterial.IsKeywordEnabled(LodDataMgrShadow.RequiredShaderKeyword))
{
    showMessage
    (
        LodDataMgrShadow.KeywordMissingErrorMessage,
        ValidatedHelper.MessageType.Error, ocean.OceanMaterial
    );

    isValid = false;
}
```

Worth doing?

Also, the case for the material keyword being enabled but the sim being disabled isn't covered in the *OceanRenderer* validation. Was this intentionally left out?

huwb:
Thanks good shout, done.

Good point about the other way around. I cant get that straight in my head because a single ocean material might be shared across multiple scenes in a project, and e.g. clip surface may only be enabled in some of them. It may not be an error, and I'm not sure if it should be a warning either. It could be an info box.

huwb:
Also added a little "jump to" button experiment for your consideration!

crest/Assets/Crest/Crest/Scripts/LodData/Shadows/LodDataMgrShadow.cs
```diff
@@ -91,7 +94,7 @@ public override void Start()
             }
 
 #if UNITY_EDITOR
-            if (!OceanRenderer.Instance.OceanMaterial.IsKeywordEnabled("_SHADOWS_ON"))
+            if (!OceanRenderer.Instance.OceanMaterial.IsKeywordEnabled(MATERIAL_KEYWORD))
             {
                 Debug.LogWarning("Shadowing is not enabled on the current ocean material and will not be visible.", _ocean);
```
daleeidd, line 99: Could use *MATERIAL_KEYWORD_MISSING_ERROR* here.

crest/Assets/Crest/Crest/Scripts/LodData/Shadows/RegisterShadowInput.cs
```diff
@@ -21,5 +21,11 @@ public class RegisterShadowInput : RegisterLodDataInput<LodDataMgrShadow>
         protected override string ShaderPrefix => "Crest/Inputs/Shadows";
 
         protected override bool FollowHorizontalMotion => false;
+
+        protected override bool FeatureEnabled(OceanRenderer ocean) => ocean.CreateShadowData;
+        protected override string FeatureDisabledErrorMessage => "<i>Create Shadow Data</i> must be enabled on the OceanRenderer component.";
+
+        protected override string RequiredShaderKeyword => LodDataMgrShadow.MATERIAL_KEYWORD;
+        protected override string KeywordMissingErrorMessage => LodDataMgrShadow.MATERIAL_KEYWORD_MISSING_ERROR;
```
daleeidd, line 29: Will need to be wrapped in *UNITY_EDITOR* so it can build. Same goes with the others.

daleeidd, line 29: Will need to be wrapped in *UNITY_EDITOR* so it can build. Same goes with the others.

daleeidd (changes requested):
Nice! Jump button is a great idea. Just a few things to fix for builds.

daleeidd:
> Good point about the other way around. I cant get that straight in my head because a single ocean material might be shared across multiple scenes in a project, and e.g. clip surface may only be enabled in some of them. It may not be an error, and I'm not sure if it should be a warning either. It could be an info box.

That's fair. It might be good as an info box as you say to push new users through misconfiguration.

crest/Assets/Crest/Crest/Scripts/LodData/Shadows/LodDataMgrShadow.cs
huwb: Done thanks

crest/Assets/Crest/Crest/Scripts/LodData/Shadows/RegisterShadowInput.cs
huwb: Ah well caught, think i've done it right, seems to build fine now

huwb:
@daleeidd addressed feedback and added info boxes

daleeidd (approved):
Works well. This is a great QoL improvement.

huwb:
Thanks!

PR #764, opened 2021-02-22
Fix PSSL/PS4 shader errors

daleeidd:
Reported by @woistjadefox in issue #624:

> Tried to make a build but I get these old errors:
> image
> [editor_error_log.txt](https://github.com/crest-ocean/crest/files/5379376/editor_error_log.txt)
> 
> I think this happend in a version before already. Should I just switch all half3 and half2 to float3 / float2 in the underwater shader for testing reasons?

Not sure if I got them all. Some of these might be URP specific so I will have to do a pass there too. This also includes a fix for the determinant wanting float4 for PSSL which was reported on Discord.

@woistjadefox If you could test this branch and let me know if you see any more errors and I will try and fix them here.

### DIFF
--- a/crest/Assets/Crest/Crest/Shaders/OceanEmission.hlsl
+++ b/crest/Assets/Crest/Crest/Shaders/OceanEmission.hlsl
@@ -101,7 +101,7 @@ void ApplyCaustics(in const half3 i_view, in const half3 i_lightDir, in const fl
 
 	const float3 scenePosUV = WorldToUV(scenePos.xz, cascadeData1, _LD_SliceIndex + 1);
 
-	half3 disp = 0.;
+	float3 disp = 0.0;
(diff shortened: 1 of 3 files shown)


woistjadefox:
Hi @daleeidd 
I only had to change the cast to float2x2 to a float4 in these lines:
image

Happend on the PS4 for the URP and HDRP version. The HDRP version didn't throw any errors while building but the water was pink. Thankfully I built the URP version first and got the errors. 

So I'm quite sure you're changes will work.

daleeidd:
Thanks for the correction! Should be all fixed now.

There is one missing which I suspect is the real2 for caustics shadows in URP. It is probably being promoted to float2 which causes them mismatch. I'll fix that downstream.

huwb (approved):

PR #765, opened 2021-02-25
Fix ShapeGerstner breaking in build

daleeidd:
Reported by @holdingjason on Discord. The active spectrum isn't set from the serialised spectrum in builds.


huwb (approved):
Good catch

PR #766, opened 2021-02-25
Add embedded material editor to OceanRenderer

daleeidd:
Adds a material editor similar to what the mesh renderers get. The only difference is that this appears right after the component rather than after all other components.

I wasn't able to make the feature an attribute like the SO editor. This is more difficult as the foldout should appear after the component at the very least. It is probably possible but will require more engineering than the other.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
+++ b/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
@@ -762,11 +762,14 @@ namespace Crest
 
         void WritePerFrameMaterialParams()
         {
-            // Hack - due to SV_IsFrontFace occasionally coming through as true for back faces,
-            // add a param here that forces ocean to be in underwater state. I think the root
-            // cause here might be imprecision or numerical issues at ocean tile boundaries, although
-            // i'm not sure why cracks are not visible in this case.
-            OceanMaterial.SetFloat(sp_ForceUnderwater, ViewerHeightAboveWater < -2f ? 1f : 0f);
+            if (OceanMaterial != null)
+            {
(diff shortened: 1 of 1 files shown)


crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
```diff
@@ -1324,6 +1347,45 @@ public override void OnInspectorGUI()
...
+                {
+                    // Free the memory used by the previous MaterialEditor.
+                    DestroyImmediate(_materialEditor);
+                }
+
+                if (_target._material != null)
+                {
+                    // Create a new instance of the default MaterialEditor.
+                    _materialEditor = (MaterialEditor)CreateEditor(_target._material);
+
```
huwb, line 1371: whitespace can be removed

huwb:
Wow nice! I think this is fine as is.

I dont know if we ever supported this but i tried clearing the material reference, got one null ref exception which i pushed a commit to fix, then it seems to null ref except in the new code, in a slightly confusing way and i wasnt certain the best fix. Would it be possible to support clearing the material off the component more gracefully?

daleeidd:
Thanks. All fixed.

> I dont know if we ever supported this but i tried clearing the material reference, got one null ref exception which i pushed a commit to fix, then it seems to null ref except in the new code, in a slightly confusing way and i wasnt certain the best fix. Would it be possible to support clearing the material off the component more gracefully?

That was an awful conditional statement. I parted it out to solve the issue.

huwb (approved):
This is super awesome

daleeidd:
Thanks. Merged :^)

PR #767, opened 2021-02-25
Deprecate AssignLayer. Use LayerMask for depth cache

### LINKED ISSUE #752: Undo command reverts AssignLayer layer causing the ocean depth cache to be empty on next populate
**Describe the bug**
The AssignLayer that we use in the examples assigns a layer on Awake. If there is an undo command, the layer change will revert. Then the cache becomes empty on next populate as nothing has the layer anymore. This will persist when entering play mode as Awake is not called again. It fixes itself when exiting play mode.

**Versions**
Ever since the editor feature.

**To Reproduce**
1. Open the main scene
2. Change a property
3. Edit > Undo
4. Click _Populate Cache_ button

**Additional context**
This is mainly limited to our examples since I don't expect many to use the assign layer.

**Possible solutions**
Changing persistent data in the editor is definitely something we should avoid.

One solution is to handle the editor case separately. The AssignLayer component can have a static dictionary which the ocean depth cache can access to use DrawRenderer or something rather than using layers when in the editor.

The other solution is to no longer use the AssignLayer component (extra benefit of removing lots of code). We set the layer as we would normally and use a LayerMask property on the depth cache. The following is what a user would see if no layer was assigned (still works as the layer int value is maintained):

For our examples, we could add a validation component which checks if the layer mask has a _Terrain_ layer and if it doesn't, then warn the user. Probably not necessary since it is limited to our example content anyway.

daleeidd:
Fixes #752 

- Marks *Assign Layer* component as *Obsolete*
- Marks *Layer Names* component as *Obsolete* and hides in inspector
- Adds Layers property (LayerMask)
- Adds fix button in validation to help users migrate
- Sets *Terrain* layer in main scene prefab (some extra data in the diff, sorry)
- Removes *Assign Layer* component from main scene
- Updates documentation

Terrain layer will not exist when importing Crest, but this is fine since it will still work. Since we are setting it in the main scene, I believe there will be no conflicts.

If users access the _layerNames in scripting, they'll see a warning. And the AssignLayer component will have a warning help box.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,6 @@ There is also a getting started video here: https://www.youtube.com/watch?v=qsge
   * The SRP assets on the Asset Store specify the minimum version required.
   * Releases on this GitHub target the built-in render pipeline, and each release specifies which version of Unity it was developed on. Currently Unity 2019.4.8 or later is the minimum version. See further down for 2018.4 support.
 * *Crest* example content:
-  * The content requires a layer named *Terrain* which should be added to your project.
   * The post processing package is used (for aesthetic reasons), if this is not present in your project you will see an unassigned script warning which you can fix by removing the offending script.
 * .NET 4.x runtime
 * [Shader compilation target](https://docs.unity3d.com/Manual/SL-ShaderCompileTargets.html) 4.5 or above
--- a/USERGUIDE.md
+++ b/USERGUIDE.md
@@ -300,7 +300,7 @@ If the *Spline* component is attached to the same GameObject as a *ShapeGerstner
 For this information in video format, see here: https://www.youtube.com/watch?v=jcmqUlboTUk
 
 *Crest* requires water depth information to attenuate large waves in shallow water, to generate foam near shorelines, and to provide shallow water shading.
-The way this information is typically generated is through the *OceanDepthCache* component, which takes one or more layer names, and renders everything in those layers from a top-down orthographic view to generate a heightfield for the seabed.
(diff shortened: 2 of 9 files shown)


huwb:
So instead of relying on the Terrain layer to be present in the project, you just picked a layer number and will assign all our things to that layer, in all RPs? Which number did you pick. I wondered if picking the largest number (31?) would work which may minimize overlaps with users existing layers.

daleeidd:
> So instead of relying on the Terrain layer to be present in the project, you just picked a layer number and will assign all our things to that layer, in all RPs? Which number did you pick.

Layer 9. I just used the one we already been using.

> I wondered if picking the largest number (31?) would work which may minimize overlaps with users existing layers.

I've moved it to 31 as suggested. The only difference it would make is reducing possible confusion as to why we are using a layer that the user defined (why are they using my navigation layer?). But functionally, I don't see there being a problem if the is an overlap.

huwb:
Yeah maybe i am over thinking it. I was wary of having our stuff assigned with some random layer from the users project. I guess it could be a layer they're using in a rendering way and end up breaking the examples. However that might be a tiny edge case thinking about it. In any case this looks good and ill approve it

huwb (approved):

daleeidd:
> I guess it could be a layer they're using in a rendering way and end up breaking the examples. However that might be a tiny edge case thinking about it. In any case this looks good and ill approve it

That's a good point. Luckily, I think SRP has made new layer systems for rendering so hopefully that edge case is limited to BIRP only.

> In any case this looks good and ill approve it

Thanks. Will merge.

PR #768, opened 2021-02-25
Add inline button to fix validation issues

daleeidd:
Adds a "fix" button (similar to the jump button) to solve validation issues.

#767 is a really good case for this which I have provided as a test commit. It will migrate the layer names to the layer mask for the user.

Implementation was a lot simpler than anticipated too which is nice.

I have also added a fix for the jump button. *Selection.activeObject* could be the GO which isn't what we pass through (we pass "this" which is the component). Alternatively, it could be fixed by always passing the GO but the current fix covers both cases.

The icon was chosen from the following [list of icons](https://github.com/halak/unity-editor-icons). I couldn't find a suitable fix or repair icon.

Let me know what you think.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/OceanValidation.cs
+++ b/crest/Assets/Crest/Crest/Scripts/OceanValidation.cs
@@ -32,9 +32,10 @@ namespace Crest
         {
             public string _message;
             public Object _object;
+            public System.Action<SerializedObject> _action;
         }
 
-        // This is a shared resource. It will be cleared before use. It is only used by the HelpBox delegate since we 
+        // This is a shared resource. It will be cleared before use. It is only used by the HelpBox delegate since we
         // want to group them by severity (MessageType). Make sure length matches MessageType length.
         public static readonly List<HelpBoxMessage>[] messages = new[]
         {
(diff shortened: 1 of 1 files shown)


huwb:
This is cool!

It's a shame the icon isn't darker, my first impression of that the button is disabled as it seems greyed out. Not sure if there's a good solution for that.

I'm contemplating how the action is called and then object is dirtied. If the fix action does not change the object for whatever reason, then it will be dirty without actual change. Or as you noted in a comment the fix function might change non-serialised data erroneously. I wonder if we should construct a serialisedobject from the object ref and pass this through to the fix action. This would then (soft) constrain the fix function to operate on serialised data, then we can call apply modified properties on that SO which may already handle dirtying (I don't remember but seems like it might).

daleeidd:
> It's a shame the icon isn't darker, my first impression of that the button is disabled as it seems greyed out. Not sure if there's a good solution for that.

Fixed for both buttons by choosing a darker icon:


> I'm contemplating how the action is called and then object is dirtied. If the fix action does not change the object for whatever reason, then it will be dirty without actual change. Or as you noted in a comment the fix function might change non-serialised data erroneously. I wonder if we should construct a serialisedobject from the object ref and pass this through to the fix action. This would then (soft) constrain the fix function to operate on serialised data, then we can call apply modified properties on that SO which may already handle dirtying (I don't remember but seems like it might).

Good idea. Done. Works as you suspected it would. The *Undo.RecordObject* is no longer necessary either but using it allows us to specify the label. Working with SerializedObject can be more difficult to work with but we can address those difficulties when we get to them.

huwb:
Oh super! That's great. That's a perfect use of the SO system.

Ha that's funny that the existing icon was light. I should probably try the dark icon in the dark theme..

huwb:
If the fix function needed to it could access the target object through the SO fields. However I think we should aim for fixes to be simple and data only like your example.

daleeidd:
> Ha that's funny that the existing icon was light. I should probably try the dark icon in the dark theme..

I tested this out. Looks like an icon with the d_ prefix is the dark theme icon so it is always be light. But without the prefix it will alternate for us depending on the chosen theme. The current solution works for light and dark theme.

huwb:
Ok cool

daleeidd:
Removed the test commits and rebased.

huwb (approved):
Nice

PR #770, opened 2021-02-26
Add tooltip attribute for materials/shaders

daleeidd:
Incomplete. Need to finish all the other built-in attributes like Enum and KeywordEnum. Looking for feedback if you think this is worth pursuing further.

Code is based on (MIT):
https://github.com/supyrb/ConfigurableShaders/blob/master/Scripts/Editor/MaterialSimpleToggleDrawer.cs

Adds our own material/shader attributes so we can add tooltips.

## Usage

Tooltip
```shader
[CrestTooltip(Normal map texture should be set to Normals type in the properties)]
```

Toggle:
```shader
[CrestToggle(_Tooltip Whether to add normal detail from a texture. Can be used to add visual detail to the water surface)]
```

I have prefixed with Crest to avoid conflicts since they are global.

## Issues
- It requires redefining the built-in drawer attributes (like Toggle) as only one drawer per property is allowed
  - Others: ToggleDrawer, EnumDrawer, KeywordEnumDrawer, PowerSliderDrawer, IntRangeDrawer
- For special characters, only whitespace and periods are allowed
- Attributes with multiple parameters require a prefix (eg _Tooltip) to determine what they are intended for

### Custom Shader GUI
A custom shader GUI (custom material inspector in other words). This would require manually rendering properties ourselves so they show up. A custom inspector has its own advantages.

### Decorator
The *Header* attribute is a decorator drawer which can have multiple. We could have a button at the inspector which enabled help boxes. Once enabled, these decorator drawers would activate.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Editor/MaterialTooltipDrawer.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Editor/MaterialTooltipDrawer.cs
@@ -1,0 +1,182 @@
+﻿// Crest Ocean System
+
+// Adapted from: https://raw.githubusercontent.com/supyrb/ConfigurableShaders/master/Scripts/Editor/MaterialTooltipDrawer.cs
+// License: https://github.com/supyrb/ConfigurableShaders/blob/master/LICENSE.md
+
+namespace Crest.Editor
+{
+    using System.Reflection;
+    using UnityEditor;
+    using UnityEngine;
+
+    /// <summary>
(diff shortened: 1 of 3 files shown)


huwb:
Just spotted the "Custom Editor GUI" field in the shadergraph graph properties. not sure if thats relevant here.

daleeidd:
It is. We will go with a custom editor as we need it for shader graph anyway. Closing.

PR #771, opened 2021-02-27
Add Crest icon to components and SOs under Crest/Scripts

daleeidd:
Adds the Crest icon to all MonoBehaviour and ScriptableObject classes which are located under Crest/Scripts.

We can go even further one day and add custom icons tailored to each component.


huwb:
Looks good. Is the crest icon image in the assets folder already?

huwb (approved):

daleeidd:
> Is the crest icon image in the assets folder already?

It is... but in Crest-Examples. I'll have to promote it to Crest.

daleeidd:
I've moved it into a new Crest/Icons folder.

huwb:
Sweet

huwb:
I've ran into an issue unfortunately - the icons are used as gizmos in the scene view so all the crest components seem to have representation in the scene view now. 

image

Maybe to some extent this is ok, but the spline points either need a different icon, or they want to have no icon so that they use the gizmoe we use now. And some other components we may not want to have icons drawn for because their 3d position is not interesting to see, nor is having a quick way to select them useful i dont think.. but let me know what you think.

I'm not sure if the gizmo icon can be disabled for some things, that could be useful.

huwb:
(Having icons on our scriptable objects / settings assets works well though)

daleeidd:
I found a way using reflection. Will do a PR for it tomorrow.

PR #772, opened 2021-02-28
Add fix button for no renderer and for feature disabled

huwb:
While working on something else i could not resist trying the fix-me feature.

You're right it works really well.


daleeidd (approved):
Nice. Works great.

PR #774, opened 2021-03-01
Hide some gizmo icons in the scene view

daleeidd:
Uses reflection in Awake to disable the icon for the scene view. I have only added it to a few for now. If we go with this solution then I will add the rest.

The only downside is that it takes the user's choice away since it will always overwrite. But that should be okay.

Let me know what you think.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Helpers/Editor/EditorHelpers.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Helpers/Editor/EditorHelpers.cs
@@ -6,6 +6,7 @@
 
 using System.Collections;
 using System.Collections.Generic;
+using System.Reflection;
 using UnityEditor;
 using UnityEngine;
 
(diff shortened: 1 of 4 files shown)


huwb:
Thanks for this. I feel a bit conflicted about merging it.

I'm usually not a fan of using reflection to get at stuff they did not expose. However the example you borrowed it from is from unity so i guess its ok.. and if the underlying stuff changes then i guess it fails in a not so bad way.

It worked for me. It seems to hard kill the icons - i went back to the master branch and the icons are gone for that component. I guess deleting the library folder would sort it. and i guess its unlikely we'd want to bring them back.

Adding an awake method to every one of our components is not so nice. and maybe some users do rely on the icons for selecting crest things and wont want us to knock all of them out, maybe?

So the thinking here is that the icons look good/professional and are worth the code additions, and unity's UX kinks? its a shame that this functionality in unity is broken. maybe this should be logged as an editor improvement on their forum now that they have a team for that.

I think you mentioned another option was to add a folder for the script name which could address the 'Spline being a generic name' issue. it would be interesting to see if that resolves this.

daleeidd:
I'm feeling the same. Should I revert the icons PR?

> I think you mentioned another option was to add a folder for the script name which could address the 'Spline being a generic name' issue. it would be interesting to try that and see if that resolves the original issue.

I'll do that next.

huwb:
Yeah probably. That's a shame. I'll contemplate putting feedback on the forum as the icons looked great.

daleeidd:
Done: https://forum.unity.com/threads/assigning-icons-set-default-gizmo-visibility.1068026/

PR #775, opened 2021-03-02
Clip setting enum missing

huwb:
@daleeidd this is an odd one - the "everything/nothing" option seems to get hidden by the PredicatedField attribute. I havent had time to dig deeper. Is this something you have encountered?

On a side note, i'm thinking about moving this option to a new settings data, since they are easy to create now. Which would fix this in a different way. Still i felt it was worth flagging this issue.


daleeidd:
Thank you for flagging this. I missed a crucial step in the system. #776 has the fix.

PR #776, opened 2021-03-02
Fix PredicatedFieldAttribute not rendering

daleeidd:
I forgot to handle the case where the PredicatedFieldAttribute is the only MultiPropertyAttribute and needs to call EditorGUI.PropertyField. The way the system works is that the last MultiPropertyAttribute handles rendering. This wasn't a problem with the embedded attributes because they handled the call and the PredicatedFieldAttribute only "decorated".


huwb (approved):
Thanks

PR #777, opened 2021-03-03
Revert "Add Crest icon to components and SOs under Crest/Scripts"

daleeidd:
Reverts wave-harmonic/crest#771


huwb:
I have an extremely late and probably bad idea - feel free to disregard.

Would it be worth keeping the icons just for the SOs? Would this brand the asset data objects with the icon, but not have the other disadvantages? Not sure if this is inconsistent or etc. Let me know what you think and sorry I only just thought of this..!

huwb:
To be clear the question is, would retaining icons for the sim settings and other SOs be:

* not too inconsistent / weird for the user
* nicely add a crest icon for data assets in the project folder that are ours
* not require any special code to make it work
* be worth doing in your opinion

daleeidd:
It wouldn't require any special code. It looks like the SO instances also get the icons. I don't see why not. I'll do a PR.

PR #778, opened 2021-03-03
Fixed simulation timesteps

huwb:
This PR replaces #734.

**Background**

Our sims (foam and dyn waves) run at variable update rates, which works surprisingly badly:

#705 , #522 , #408

Part of the issue may be our integration scheme, but other schemes are more expensive and more involved to implement. We apply dissipation/friction using a simple lerp model, i tried the other approach (exp) and it did not help. I think i've just learnt the hard/slow way why physics engines run at a fixed rate.

**Solution**

This PR forces the foam sim to run at 30hz and the dynamic wave sim to run at 60hz. If the foam sim is at least 33.33ms behind the current unity time, it does a step, then the result is used in the render. Doing rigid steps like this may be susceptable to visible jitter, and if so might require an "interpolation" scheme where we always simulate past the current time, and then interpolate current and previous sim to get the result. This would add extra cost however, and i did not notice jittering in tests so far, so maybe this is good enough. Also if users are using v-sync and rendering at 60hz or 30hz, the result should look perfect.

It also makes adjustments based on current gravity setting to try to retain behaviour better at different gravity multipliers.

Sphere water interaction: Uses a fancy bandfiltered falloff to get consistent interaction forces independent of camera position relative to object. This could not be applied to the interaction shader that the boats use, so only the spinner benefits from this for now. Later it would be good to move all boats to this component and deprecate/remove the other interaction shader.

**Impact**

This will have slight changes to behaviour but i believe it will be close enough, or require small retweaks. Once retweaked to suit the new dt, the behaviour should be consistent across any render framerate.

Dale i think this could use a bit more testing than a typical PRs, if you're able to run it through its paces that would be great. Thanks!


daleeidd:
Is this part of or related to #734?

huwb:
Good question. 734 has grown unwieldy. This replaces it. I'll close it.

daleeidd:
It looks like foam from the dynamic waves isn't being renderers into all the LODs: [master](https://drive.google.com/file/d/1sNsphsQ9m3HalH5042d4QfffpX2tSeVp/view?usp=sharing) vs [PR](https://drive.google.com/file/d/10YsxK6msaAcOMM6K1mAiBvGLapWBL13T/view?usp=sharing)

I think this is also happening to dynamic waves. Or more likely it is dynamic waves which is the issue hence no foam.

huwb:
@daleeidd the missing foam was because the spinner was not injecting ripples into enough simulation cascades, so spinner ripples were very limited in range, which in turn meant foam was limited in range. it now writes into an additional cascade and i've retweaked to retain behaviour.

would you please give it another test / review? thanks

daleeidd (approved):
I struggled to find any issues with the latest changes. Source looks good too. We could postpone this and get feedback from others? Although, I think @holdingjason has moved onto URP now.

huwb:
I did not get any feedback from anyone i pinged so far. i think the people who reported the issues gave up on using the foam and ripple sims, due to the issues...

daleeidd:
No worries. Vote for merging then :^)

daleeidd:
> @daleeidd the missing foam was because the spinner was not injecting ripples into enough simulation cascades, so spinner ripples were very limited in range, which in turn meant foam was limited in range. it now writes into an additional cascade and i've retweaked to retain behaviour.

@huwb 

I made a mistake and didn't see that this is not 100% solved. They disappear at about 130 height above the spinner for me. Whilst the original could go over 1000 before it disappeared.

daleeidd:
Also noticed that the default gravity (1) looks slow. Tested using the SphereInteraction object by copy and pasting it from boats.unity to another scene. Not sure if intentional.

huwb:
Default gravity setting on teh dynamic wave sim settings? In theory gravity==1 should make it match teh gerstners. It always appears slow to me too, but i think its physical..

I made a separate PR to extend the range, thansk for that

PR #779, opened 2021-03-03
Don't let gerstner waves shader sample displacement

huwb:
This creates a feedback loop of doom and is making the patch of waves in boat.unity flicker for me.


daleeidd (approved):
Tested. I didn't see any issues so good to merge I think.

I was curious. I had a look and it looks like this variable wasn't being populated for the gerstner patch so it was completely unused (might have had trash data?).

huwb:
Thanks

I think this vector is set through Shader.SetGlobalVector for all input types.

PR #781, opened 2021-03-05
Make shader keyword check compatible with Shader Graph shader

daleeidd:
Shader keywords differ with SG. They either have a prefix or are missing. The point of this PR is to reduce diff noise and increase compatibility. Hopefully, this will be a temporary solution.

Adds a *MATERIAL_KEYWORD_PREFIX* constant which is used:
```csharp
internal const string MATERIAL_KEYWORD = MATERIAL_KEYWORD_PREFIX + "_FOAM_ON";
```
For built-in, the prefix is an empty string, but CREST_ will be added in HDRP (diff noise).

MATERIAL_KEYWORD_PROPERTY

The next is skipping any keyword where the *OceanMaterial.HasProperty* is false. This checks the property name. So for foam we check "_Foam" which matches with "FOAM_ON".

So with SG, if we do not have a keyword defined, then it will be skipped.

*MATERIAL_KEYWORD_PROPERTY* has been added which will need to be changed in HDRP (diff noise).

This will give us time to consider #769 more carefully. If that PR is merged only the prefix change needs to be reverted. The HasProperty fix will probably need to stay.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgr.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgr.cs
@@ -30,6 +30,9 @@ namespace Crest
         public const int THREAD_GROUP_SIZE_X = 8;
         public const int THREAD_GROUP_SIZE_Y = 8;
 
+        // NOTE: This is a temporary solution to keywords having prefixes downstream.
+        internal const string MATERIAL_KEYWORD_PREFIX = "";
+
         protected abstract int GetParamIdSampler(bool sourceLod = false);
 
         protected abstract bool NeedToReadWriteTextureData { get; }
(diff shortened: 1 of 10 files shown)


huwb (approved):
I didn't quite get why the has property but it's required? But feel free to merge in the meantime

daleeidd:
It's for when we don't have a keyword in HDRP like for the clip surface. Currently, the checks will trigger the validation messages (unless we disable the code). Unfortunately, there is no "HasKeyword". So we check if the keywords property counterpart exists instead. I will do a few more tests before merging so it is robust. I think there might be a better solution.

huwb:
Oh gotcha

daleeidd:
After checking it seems robust. I'll push this through to get the keyword check working downstream.

PR #786, opened 2021-03-11
Rearrange and rename ocean material properties

daleeidd:
Makes it similar to shader graph. It will help with documentation and referencing.

I chose the names from HDRP where they matched.

I also made two new sections: _Shallow Scattering_ and _Foam Bubbles_. Breaks up the documentation better with these additional headings and properly groups foam bubbles settings together.


huwb (approved):
Great stuff thanks

PR #788, opened 2021-03-16
Change VFACE to SV_IsFrontFace

daleeidd:
It has been reported that VFACE is causing issues on the latest consoles.

According to [Microsoft documentation about VFACE](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-semantics#semantics-supported-only-for-direct3d-10-and-newer):
> This semantic is available in Direct3D 9 Shader Model 3.0. For Direct3D 10 and later, use SV_IsFrontFace instead.

### DIFF
--- a/crest/Assets/Crest/Crest/Shaders/Ocean.shader
+++ b/crest/Assets/Crest/Crest/Shaders/Ocean.shader
@@ -437,16 +437,15 @@ Shader "Crest/Ocean"
 				return lightDir;
 			}
 
-			bool IsUnderwater(const float facing)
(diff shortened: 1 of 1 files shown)


huwb (approved):
Nice one. It would be nice if this fixed the "fire fly" pixels where it seems to wrongly classify pixels in some cases, but i wont get my hopes up!

daleeidd:
> Nice one. It would be nice if this fixed the "fire fly" pixels where it seems to wrongly classify pixels in some cases, but i wont get my hopes up!

Thanks. Sadly, it doesn't make a difference.

PR #789, opened 2021-03-19
Use AddComponentMenu for easier component searching

daleeidd:
Reorganises our scripts in the component menu by adding submenus. Also adds a "Crest" prefix to our scripts so that they all come up if someone searches "Crest". I didn't touch any components that were generic utilities. There is also an additional file for housing constants.

Searching:
Add Component Menu Search

Submenus:
Add Component Menu Sub-Menus


huwb (approved):
Nice one

PR #790, opened 2021-03-20
Scale gap fix by camera position

daleeidd:
Helps with underwater when camera is reasonable (lack of better word) distance from zero. Nothing really solves underwater issues at larger distances though, but the gap fix does work well for above water.

If we find that this is going to be permanent, then we could script the lerp value to save performance cost.


huwb (approved):
Thanks

PR #791, opened 2021-03-21
Fix/sphere water interaction range

huwb:
Thanks for the feedback on #778 . You were right, it LOD'd far too aggressively. Inspecting carefully i think the range could be extended a ton, it should be similar to before now.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Helpers/OceanDebugGUI.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Helpers/OceanDebugGUI.cs
@@ -15,6 +15,8 @@ namespace Crest
         public bool _showOceanData = true;
         public bool _guiVisible = true;
 
+        [SerializeField] bool _drawLodDatasActualSize = false;
+
         [Header("Lod Datas")]
         [SerializeField] bool _drawAnimWaves = true;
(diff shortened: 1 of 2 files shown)


daleeidd:
That fixes it at a distance. I did some more comparisons and I found that they are more powerful than the old approach before fixed time. The following images is at around 10 seconds elapsed game time.

**Before fixed time steps**
111 1 Old

**This PR**
111 2 PR

**After changing both 0.2s to 0.1s in shader**
111 3 PR Adjusted

Not sure if this was intentional. I found comparing the data textures themselves to be difficult since the pattern is a little different.

daleeidd (approved):
Looks good :^)

huwb:
Ah thanks. I did make it a bit stronger. However I will recheck and try 0.1 and may push an update later. For now I'll merge this as a range fix.

huwb:
0.1 seems far too chill in terms of displacement/shape when im close up to the spinner. but i appreciate the difference above is significant - at least in terms of foam? im not quite sure but i think ill leave it as is.

daleeidd:
Yeah, it's probably just the foam. My bad.

Happy to leave it. Thought I'd mention it though.

huwb:
Yes was a useful callout, and may be useful to refer back to later

PR #793, opened 2021-04-02
Fix-me button for material keywords

huwb:
Helper to enable/disable material keywords.

Works functionaly but unfortunately the material editor does not reflect the changed keyword and i could not work out why. It is changing the asset successfully, so i dont know why the editor does not reflect this. Posting in case you can work it out


huwb:
Closing, created new PR with only the relevant commit

PR #794, opened 2021-04-02
Fix-me button for material keywords

huwb:
Helper to enable/disable material keywords.

Works functionaly but unfortunately the material editor does not reflect the changed keyword and i could not work out why. It is changing the asset successfully, so i dont know why the editor does not reflect this. Posting in case you can work it out

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
+++ b/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
@@ -1324,15 +1324,32 @@ namespace Crest
             {
                 if (ocean.CreateClipSurfaceData)
                 {
-                    showMessage(LodDataMgrClipSurface.ERROR_MATERIAL_KEYWORD_MISSING, ValidatedHelper.MessageType.Error, ocean.OceanMaterial);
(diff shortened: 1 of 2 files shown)


crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
```diff
@@ -1324,17 +1324,34 @@ public bool Validate(OceanRenderer ocean, ValidatedHelper.ShowMessage showMessag
...
         }
 
+        internal static void FixSetMaterialKeywordEnabled(SerializedObject so, string keyword, bool enabled)
+        {
+            Material mat = so.targetObject as Material;
+            Undo.RecordObject(mat, $"Enable keyword {keyword}");
+            // Does apply change in viewport, but does not change shared material asset it seems..
+            if (enabled)
+            {
+                mat.EnableKeyword(keyword);
```
daleeidd, line 1347: The checkboxes are based off the float value. So for _CLIPSURFACE_ON keyword, there is also _ClipSurface float. So also changing the float is required:
```csharp
mat.SetFloat("_ClipSurface", 1f);
```

I added *MATERIAL_KEYWORD_PROPERTY* a little while ago which will help.

I don't think there is another way.

huwb: thanks

huwb:
Thanks, rolled into #792

PR #796, opened 2021-04-03
Depth cache fix functions

huwb:
Sorry to hit you with yet another review on fix functions but i didnt want to noise up the other PR.

This proposes a few valdiation/fix functions for the depth cache.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
@@ -467,6 +467,10 @@ namespace Crest
             localScale.y *= ymul;
             localScale.z *= zmul;
 
+            // Try to recover from 0 scale
+            if (localScale.x == 0f) localScale.x = localScale.z;
+            if (localScale.z == 0f) localScale.z = localScale.x;
+
             dc.transform.localScale = localScale;
         }
 
--- a/crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
@@ -482,6 +482,18 @@ namespace Crest
             dc.transform.position = pos;
         }
 
+        void FixRotation(SerializedObject depthCache)
+        {
+            var dc = depthCache.targetObject as OceanDepthCache;
+
+            Undo.RecordObject(dc.transform, "Fix depth cache rotation");
+            EditorUtility.SetDirty(dc.transform);
+
+            var ea = dc.transform.eulerAngles;
+            ea.x = ea.z = 0f;
+            dc.transform.eulerAngles = ea;
+        }
+
         public bool Validate(OceanRenderer ocean, ValidatedHelper.ShowMessage showMessage)
         {
             var isValid = true;
@@ -606,6 +618,19 @@ namespace Crest
                     "Set the Y position to the same height as the ocean object.",
                     ValidatedHelper.MessageType.Warning, this,
                     FixHeight
+                );
+
+                isValid = false;
+            }
+
+            if (!Mathf.Approximately(transform.eulerAngles.x, 0f) || !Mathf.Approximately(transform.eulerAngles.z, 0f))
+            {
+                showMessage
+                (
+                    "The depth cache should have 0 rotation around X and Z (but rotation around Y is allowed).",
+                    "Adjust the rotation on this transform and parents in the hierarchy to eliminate X and Z rotation.",
+                    ValidatedHelper.MessageType.Error, this,
+                    FixRotation
                 );
 
                 isValid = false;
--- a/crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
@@ -472,6 +472,11 @@ namespace Crest
             if (localScale.z == 0f) localScale.z = localScale.x;
 
             dc.transform.localScale = localScale;
+
+            if (dc.Type == OceanDepthCacheType.Realtime)
+            {
+                dc.PopulateCache(true);
+            }
         }
 
         void FixHeight(SerializedObject depthCache)
@@ -484,6 +489,11 @@ namespace Crest
             var pos = dc.transform.position;
             pos.y = OceanRenderer.Instance.transform.position.y;
             dc.transform.position = pos;
+
+            if (dc.Type == OceanDepthCacheType.Realtime)
+            {
+                dc.PopulateCache(true);
+            }
         }
 
         void FixRotation(SerializedObject depthCache)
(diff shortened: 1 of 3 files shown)


crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
```diff
@@ -425,6 +425,79 @@ bool IsCacheOutdated()
...
+        {
+            depthCache.FindProperty("_forceAlwaysUpdateDebug").boolValue = false;
+        }
+
+        void FixScale(SerializedObject depthCache)
+        {
+            // Slightly tricky to set scale as you can't assign world space scale.
+            // This function assumes no rotation on cache object or parents, and
+            // computes the current world scale, and uses that to compute multipliers
+            // to apply to the local scale
```
daleeidd, line 444: I find that it doesn't work if any of the parents have a non-uniform scale.

crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
```diff
@@ -439,7 +512,8 @@ public bool Validate(OceanRenderer ocean, ValidatedHelper.ShowMessage showMessag
                     (
                         "Depth cache is outdated.",
```
daleeidd, line 513: Should we auto-populate the cache after a user clicks a fix function which outdates the cache (and is not baked cache)?

daleeidd (changes requested):
Nice work. Just one issue. The other one is minor but worth considering.

crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
huwb: I put a x scale of 0.5, y scale of 2 and z scale of 0.3 all on different parents in the hierarchy and it worked.

however im not surprised at all that it failed for you. i dont think it will work robustly for all cases.

if i set a x or z rotation (not scale) on a parent, and run the fix, it does appear visually to reset the rotation, however it did not clear the error for me, and i could not figure out why. i think it may be due to lossy-ness in the transform hierarchy.

so i think the fix wont be totally reliable. i thought having it try was slightly better than not having it at all. a more hardline fix may be to reset scale/rotation on parents, but that would be very heavy handed and id be concerned about screwing up the scene for the user..

crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
huwb: good shout, done. i assume i should pass `true` to the populate function?

crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
daleeidd: I must of also had a rotation as it does work as you say. The only issue appears to be when a parent has a non-uniform scale and the depth cache has a Y rotation that is not at a 90° interval. I attempted fixing all of this myself when I added the depth cache fixes a little while ago, but struggled with this part.

> so i think the fix wont be totally reliable. i thought having it try was slightly better than not having it at all. a more hardline fix may be to reset scale/rotation on parents, but that would be very heavy handed and id be concerned about screwing up the scene for the user..

We could not offer the fix for difficult cases like the above if worst comes to worst?

crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
daleeidd: Nice. Correct. That way it updates the changed components.

crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
huwb: I'm not quite sure how to detect such cases. I guess if there is any rotation on a parent. And only offer the fix if not? Not sure the cleanest way to do this.

All of this is probably hinting that we should not use the transform at all, I think we should probably invest time there rather than here, and decide for now if we want a fix that fails or no fix. Maybe a failing fix is a bad idea.

daleeidd: > I'm not quite sure how to detect such cases. I guess if there is any rotation on a parent. And only offer the fix if not? Not sure the cleanest way to do this.

It's only if there is a local rotation on the depth cache and parent has non-uniform scale. We could ignore the 90° interval as it isn't worth checking. This should work:

```csharp
(transform.localRotation != Quaternion.identity && transform.parent.lossyScale != Vector3.zero ? null : (System.Action<SerializedObject>) FixScale)
```

Unity compares vectors approximately already when using ==. So if we want exact equals we have to use [Vector3.Equals](https://docs.unity3d.com/ScriptReference/Vector3.Equals.html).

> All of this is probably hinting that we should not use the transform at all, I think we should probably invest time there rather than here, and decide for now if we want a fix that fails or no fix. Maybe a failing fix is a bad idea.

That makes sense. Out of the two, I'd vote for no fix. There is another minor issue where the cache keeps growing with every fix because we choose the largest. It is an edge case though and would fall under misuse.

huwb: Ahh I think when I had the problem of rotation not getting fixed properly, I had no scale on any of the parents at all. If need to test again to be sure but I think that's what I saw.

> local rotation on the depth cache and parent has non-uniform scale

I guess parent here means any parent/ancestor in the hierarchy? I've been using the term parent loosely, to mean any parent up to the root. 

I'm not sure if it's ok to offer a fix that changes parents, add that could in turn impact their children in non intuitive ways.

daleeidd: > > local rotation on the depth cache and parent has non-uniform scale
> 
> I guess parent here means any parent/ancestor in the hierarchy? I've been using the term parent loosely, to mean any parent up to the root.

They both sound kind of samey. I guess direct ancestors is the right term? So only ancestors that would affect the depth cache's transform.
 
> I'm not sure if it's ok to offer a fix that changes parents, as that could in turn impact their children in non intuitive ways.

Agreed that we certainly shouldn't change anything up the hierarchy.

> I'm contemplating how to proceed. As I mentioned moving the depth cache away from using the transform is tempting.. would need to look at how to transition from the old setup though to not break everyone's projects

That would be the ideal I think as we are fighting the transform currently. But no rush as the depth cache has improved greatly this past month or so.

daleeidd (approved):
Looks good. Scale fix function can go through if you think it is still worth keeping. If you need me to do anything to get this in for 4.10, let me know.

huwb:
Ah thought we gave up on this but I'm happy for you to merge it if youre up for it, thanks

daleeidd:
Lots of other great stuff in here. I have disused the FixScale function due to drawbacks as discussed. But everything else is safe to merge in.

huwb:
Nice one

PR #803, opened 2021-04-13
Add version and RP to HelpURL

daleeidd:
Adds the version and render pipeline to the help URLs.

Currently, it points to 4.10 which doesn't exist yet. A custom 404 page will redirect users to latest until the 4.10 tag is published. This allows us to change the version once (after publishing a tag). Otherwise, we would have to change it from latest to 4.10, and then back to latest.

The constants are tedious looking, but this can be fixed with a custom help URL attribute similar to [Unity's](https://github.com/Unity-Technologies/Graphics/blob/master/com.unity.render-pipelines.core/Runtime/Documentation.cs). We cannot do this now as the HelpURLAttribute is sealed. But it must of been unsealed in 2021 or later.


huwb (approved):
Looks great

daleeidd:
Thanks

PR #804, opened 2021-04-15
Rewrite multi-property attributes as decorated/decorator attributes

daleeidd:
- Fixes embedded field attribute for 2021
- Nicer architecture
- Added range and delay fields that work with predicated
- More robust and possibly performs better (untested)

Unity changed something in 2021 causing the _MultiPropertyDrawer_ to be called twice when a property had two attributes that inherited from _MultiPropertyAttribute_ and one of those called _EditorGUI.PropertyField_ (looking at you _Embedded_).

The fix was to separate _MultiPropertyAttribute_ into decorators and decorated attributes. A property will have only one decorated field which solves the problem of targeting multiple attributes.

_PredicatedField_ has been changed to _Predicated_ and is now a decorator. It cannot be used by itself as decorators do not call _EditorGUI.PropertyField_ so  I have added a general purpose _DecoratedField_ which calls _EditorGUI.PropertyField_. This is the only drawback.

I have also added ranged and delayed decorated fields which work with decorators like predicated. We had some delayed fields which broke predicated and is now fixed.

### DIFF
--- a/crest/Assets/Crest/Crest-Examples/BoatDev/Scripts/BoatAlignNormal.cs
+++ b/crest/Assets/Crest/Crest-Examples/BoatDev/Scripts/BoatAlignNormal.cs
@@ -18,7 +18,7 @@ public class BoatAlignNormal : FloatingObjectBase
     float _buoyancyCoeff = 1.5f;
     [Tooltip("Strength of torque applied to match boat orientation to water normal."), SerializeField]
     float _boyancyTorque = 8f;
-    [Tooltip("Approximate hydrodynamics of 'surfing' down waves."), SerializeField, Range(0, 1)]
+    [Tooltip("Approximate hydrodynamics of 'surfing' down waves."), SerializeField, Crest.Range(0, 1)]
     float _accelerateDownhill = 0f;
 
     [Header("Engine Power")]
--- a/crest/Assets/Crest/Crest-Examples/BoatDev/Scripts/BoatAlignNormal.cs
+++ b/crest/Assets/Crest/Crest-Examples/BoatDev/Scripts/BoatAlignNormal.cs
@@ -34,7 +34,7 @@ public class BoatAlignNormal : FloatingObjectBase
 
     [SerializeField, Tooltip("Computes a separate normal based on boat length to get more accurate orientations, at the cost of an extra collision sample.")]
     bool _useBoatLength = false;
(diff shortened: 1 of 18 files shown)


huwb (approved):
Wow great work

I wondered if DecoratedEmbeddedField should simply be Embedded which would seem to make it consistent with Predicated.

In any case thumbs up

daleeidd:
> Wow great work

Thanks.
 
> I wondered if DecoratedEmbeddedField should simply be Embedded which would seem to make it consistent with Predicated.

I'm happy to rename. Should I do this for all attributes? This means DecoratedRangedField will be just Ranged which will conflict with Unity. We would have to use it as Crest.Ranged, but I don't see that as any more verbose than RangedField.

huwb:
I'm not sure. 

I'd be curious how it works with that conflict. Will it give an ambiguous symbol error, then we have to fully qualify it or add a using statement? That might be better behaviour than if it silently fails if we use unitys one.

On a similar note, what happens if I use our Ranged attribute accidentally on a random field - it should work right?

daleeidd:
> I'd be curious how it works with that conflict. Will it give an ambiguous symbol error, then we have to fully qualify it or add a using statement? That might be better behaviour than if it silently fails if we use unitys one.

It does exactly that. If I go this route I will have to go through and qualify all cases that we have used Unity's one. I would just use ours. Happy to do so.

> On a similar note, what happens if I use our Ranged attribute accidentally on a random field - it should work right?

It does.

huwb:
Cool sounds good

daleeidd:
> If I go this route I will have to go through and qualify all cases that we have used Unity's one

@huwb I was mistaken. It only does the ambiguous error when we use Range outside of the Crest namespace. Otherwise, there is no error. But it will always use ours in our namespace. So I still think it safe enough to go ahead. Thoughts?

daleeidd:
I've gone ahead with it for now. Can renamed them if not ok.

I've also added a validator. It lives in the Development folder so it won't be part of releases. It logs an error to ensure that a decorator attribute (eg Predicated) also has a decorated attribute (eg something that inherits from DecoratedPropertyAttribute like Range or DecoratedField).

huwb:
That sounds great! V nice

Surprised the ambiguous symbol error doesn't fire really weird. But it's fine I guess.

Issue #801, opened 2021-04-07
Wireframe gizmos render over opaque objects with Post-Processing stack


daleeidd:
**Describe the bug**
Wireframe gizmos render over opaque objects when a Post-Processing Layer is added to the camera.

**Screenshots / video**

What we expect:

<img width="783" alt="1111" src="https://user-images.githubusercontent.com/5249806/113932261-43a32f80-97a8-11eb-879d-28819c055661.png">

What we see:

<img width="783" alt="111" src="https://user-images.githubusercontent.com/5249806/113932268-46058980-97a8-11eb-96ff-7bfa56c1f54a.png">

**Versions**
Built-in render pipeline
All Unity versions

**Platform**
All platforms

**Additional context**
[Unity Issue Tracker #1124862](https://issuetracker.unity3d.com/product/unity/issues/guid/1124862). It also affects the grid. We use these a lot. Please upvote to have this fixed.

daleeidd:
For anyone wanting a workaround, please disable Post-Processing for the scene view camera:

<img width="783" alt="PP Fix" src="https://user-images.githubusercontent.com/5249806/114942397-572e4600-9df9-11eb-942c-5852b16e0e95.png">

PR #806, opened 2021-04-16
Validate scene view effects options

daleeidd:
Will help with #801

Validates both _Animated Materials_ and _Post-Processing_ options under the scene view effects menu. The former option is an information help box and the latter is a warning help box. Furthermore, the latter is only validated in built-in and if the main camera has an active layer (ie only validate when the bug occurs). See code for help box copy.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Crest.asmdef
+++ b/crest/Assets/Crest/Crest/Scripts/Crest.asmdef
@@ -1,3 +1,21 @@
-﻿{
-	"name": "Crest"
+{
+    "name": "Crest",
+    "references": [
+        "Unity.Postprocessing.Runtime"
+    ],
+    "includePlatforms": [],
+    "excludePlatforms": [],
+    "allowUnsafeCode": false,
+    "overrideReferences": false,
+    "precompiledReferences": [],
+    "autoReferenced": true,
+    "defineConstraints": [],
+    "versionDefines": [
+        {
+            "name": "com.unity.postprocessing",
+            "expression": "",
+            "define": "UNITY_POSTPROCESSING_ENABLED"
+        }
+    ],
+    "noEngineReferences": false
 }
--- a/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
+++ b/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
@@ -1260,6 +1260,55 @@ namespace Crest
                 );
             }
 
+            // Validate scene view effects options.
+            if (SceneView.lastActiveSceneView != null)
+            {
+                var sceneView = SceneView.lastActiveSceneView;
+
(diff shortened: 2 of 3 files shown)


crest/Assets/Crest/Crest/Scripts/Crest.asmdef
```diff
@@ -1,3 +1,21 @@
...
+    "allowUnsafeCode": false,
+    "overrideReferences": false,
+    "precompiledReferences": [],
+    "autoReferenced": true,
+    "defineConstraints": [],
+    "versionDefines": [
+        {
+            "name": "com.unity.postprocessing",
+            "expression": "",
+            "define": "UNITY_POSTPROCESSING_ENABLED"
```
huwb, line 17: is this intentional? just checking

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
```diff
@@ -1260,6 +1260,52 @@ public bool Validate(OceanRenderer ocean, ValidatedHelper.ShowMessage showMessag
                 );
             }
 
+            if (SceneView.lastActiveSceneView != null)
+            {
+                var sceneView = SceneView.lastActiveSceneView;
+                // TODO: PP check here!
```
huwb, line 1266: is this stale

huwb (approved):
Lovely!

crest/Assets/Crest/Crest/Scripts/Crest.asmdef
daleeidd: The define? It is.

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
daleeidd: It is. Thanks.

crest/Assets/Crest/Crest/Scripts/Crest.asmdef
huwb: do we want to add the postprocessing package?

daleeidd: All this does is create the define symbol if the post-processing package is in the project. It doesn't add any packages.

See: https://docs.unity3d.com/Manual/ScriptCompilationAssemblyDefinitionFiles.html#define-symbols

huwb: Ok so adding it to the references list above does not add it to the project?

huwb: Thumbs up for merging anyway

daleeidd: Correct.

PR #809, opened 2021-04-24
Gerstner upgrade - wind speed and wave pairs

huwb:
**Status**: Ready for testing. Run main.unity and tweak the wind speed on the ocean component.

Calculates gerstner waves in 'pairs' - this mirrors how FFTs work and it the resulting waves vary in amplitude over time. On my 1060 this increased wave generation cost from 12us to 22us.

Introduces wind speed - a global wind speed, set to a high value so it won't negatively impact existing projects, and an optional override wind speed on the ShapeGerstner component. The wind speed is used to attenuate waves that are travelling faster than the wind speed. Once this goes out, after a release or two we could reduce the default global wind speed to a slightly more sane value.

If it looks like it works for you Dale then i can revert the data and then try to make it work well with default data (or change data defaults).


daleeidd:
Works for me. The pairs is a nice improvement.

huwb:
@daleeidd this is ready for Final Review, would you do the honours?

I've reverted data. I want new data defaults. but i think that should be connected with making ShapeGerstnerBatched legacy/deprecated, which i also want to do, but not in this PR. So i'm proposing merge this as-is as a change to ShapeGerstner in preview, then deal with data defaults and etc when we un-preview it.

huwb:
Oh and the default global wind speed is very high. In the future i think we could reduce it. For now i've set it high so it does not impact user's existing wave conditions when they upgrade.

The alternative approach would be to reduce it to something 'normal', and users who had large waves (not that common) would need to increase the wind speed. I could be convinced to go this route tbh.

crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
```diff
@@ -159,6 +170,9 @@ struct GerstnerWaveComponent4
             public Vector4 _omega;
             public Vector4 _phase;
             public Vector4 _chopAmp;
+            public Vector4 _amp2;
+            public Vector4 _chopAmp2;
+            public Vector4 _phase2;
```
daleeidd, line 175: If not too much trouble, adding a note here that these are for wave pairs will make it easier to understand their purpose (at least for me). And maybe some other places where you feel appropriate.

crest/Assets/Crest/Crest/Scripts/Shapes/OceanWaveSpectrum.cs
```diff
@@ -136,6 +146,9 @@ public float GetAmplitude(float wavelength, float componentsPerOctave, out float
             power = hasNextIndex ? Mathf.Lerp(thisPower, nextPower, alpha) : thisPower;
             power = Mathf.Pow(10f, power);
 
+            var c = ComputeWaveSpeed(wavelength);
+            power *= Mathf.Clamp01(Mathf.InverseLerp(windSpeed, windSpeed * 0.95f, c));
```
daleeidd, line 150: I'm finding that even at maximum wind speed, this line will produce some zeros. They are for wavelengths 1152 and 2304. Just checking that this is expected. I am guessing they are too big even for wind speed 150.

daleeidd (approved):
Everything works well. Nice work. Just some minor things but not worth holding this up.

crest/Assets/Crest/Crest/Scripts/Shapes/OceanWaveSpectrum.cs
huwb: Yeah its a struggle with the wind speed slider. We need an exponential slider otherwise its impossible to tweak the low speeds. Any ideas how we might achieve this? I suppose brute force is to make a new editor for the ocean component, then implement the nonlinear range bespokely. but let me know if you have fancier ideas around property drawers or etc.

in any case i propose to address this in a follow up and ill merge this as is.

crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
huwb: thanks, done

crest/Assets/Crest/Crest/Scripts/Shapes/OceanWaveSpectrum.cs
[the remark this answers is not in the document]
huwb: sure done

huwb:
thanks!

crest/Assets/Crest/Crest/Scripts/Shapes/OceanWaveSpectrum.cs
daleeidd: A custom property drawer using the system we have should be achievable. I can look into this.

PR #810, opened 2021-04-25
Use surface gradients and add overall normal scale

huwb:
A couple of previous support issues would have been helped by having an overall normal strength control for the water surface which this PR adds.

Also wrapped up in these changes - there were unnecessary normalize() calls and other cleanup to do for the pixel normal calculation. This change moves to surface gradient style setup - derivatives are added/scaled, then normalize happens once at the end of the calculation.


daleeidd (approved):
Great idea. Just needs the changelog updated. But can merge in if not ready to do that yet.

huwb:
Thanks

PR #811, opened 2021-04-29
Server support

huwb:
## Intro

To support servers, we need to run fine when `-nographics` is given as a commandline option, and when `-batchmode` is given as a commandline option. We also need to avoid unnecessary work in both.

To run a server instance, run the following from the command line

`CrestBuild.exe -batchmode -nographics -logfile log.log`

I have added two toggles that emulate batch mode and emulate running without a GPU on the OceanRenderer component.

## batchmode

* Does not construct a surface mesh
* Only creates sims required for queries, like wave sim

## nographics

In addition to `batchmode`:

* Does not create any loddata
* Errors if other GPU-dependant features are used, like runtime baking of depth caches

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
+++ b/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
@@ -243,6 +243,10 @@ namespace Crest
         bool _followSceneCamera = true;
 #pragma warning restore 414
 
+        [Header("Server Settings")]
+        [SerializeField]
+        bool _forceNoGPU = false;
+
         [Header("Debug Params")]
 
         [Tooltip("Attach debug gui that adds some controls and allows to visualise the ocean data."), SerializeField]
@@ -303,7 +307,7 @@ namespace Crest
         /// <summary>
         /// Is runtime environment without graphics card
         /// </summary>
-        public static bool Headless => SystemInfo.graphicsDeviceID == 0;
+        public static bool RunningWithoutGPU => SystemInfo.graphicsDeviceID == 0 || (Instance != null ? Instance._forceNoGPU : false);
 
         // We are computing these values to be optimal based on the base mesh vertex density.
         float _lodAlphaBlackPointFade;
@@ -500,7 +504,7 @@ namespace Crest
 
         void CreateDestroySubSystems()
         {
-            if (!Headless)
+            if (!RunningWithoutGPU)
             {
                 {
                     if (_lodDataAnimWaves == null)
@@ -649,7 +653,7 @@ namespace Crest
 
         bool VerifyRequirements()
         {
-            if (!Headless)
+            if (!RunningWithoutGPU)
             {
                 if (Application.platform == RuntimePlatform.WebGLPlayer)
                 {
--- a/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
+++ b/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
@@ -245,6 +245,8 @@ namespace Crest
 
         [Header("Server Settings")]
         [SerializeField]
+        bool _forceBatchMode = false;
+        [SerializeField]
         bool _forceNoGPU = false;
 
         [Header("Debug Params")]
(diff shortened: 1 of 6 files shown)


TiToMoskito:
Awesome ! Does this also work on the SRP versions?

huwb:
Yes it should indeed

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
```diff
@@ -300,6 +306,16 @@ public enum DefaultClippingState
 
         public static OceanRenderer Instance { get; private set; }
 
+        /// <summary>
+        /// Is runtime environment without graphics card
+        /// </summary>
+        public static bool RunningWithoutGPU => SystemInfo.graphicsDeviceID == 0 || (Instance != null ? Instance._forceNoGPU : false);
```
daleeidd, line 312: *SystemInfo.graphicsDeviceID* is zero when there is only integrated graphics (tested on MacBook). There is [UnityEngine.Rendering.GraphicsDeviceType](https://docs.unity3d.com/2019.4/Documentation/ScriptReference/Rendering.GraphicsDeviceType.html) which works.

daleeidd:
Are we also going to support server builds? See *Server Build* under build https://docs.unity3d.com/2018.4/Documentation/Manual/BuildSettings.html. It would only take one line change to get the build not to fail, but there is a few exceptions spammed in the log for a server build. That documentation page is for 2018.4, but the option isn't documented in 2019.4, but still appears in the screenshot and available to me in build settings even in 2021.

huwb:
Good point! For some reason i thought the server build option was disabled for me, but it is not, so ill look at that later

huwb:
Hm i dont get a flood of exceptions if i make a server build on windows. I get two errors which i can fix by configuring as directed:

`
Crest: Compute shader queries not supported in headless/batch mode. To resolve, assign an Animated Wave Settings asset to the OceanRenderer component and set the Collision Source to be a CPU option.
UnityEngine.StackTraceUtility:ExtractStackTrace ()
UnityEngine.DebugLogHandler:LogFormat (UnityEngine.LogType,UnityEngine.Object,string,object[])
UnityEngine.Logger:Log (UnityEngine.LogType,object)
UnityEngine.Debug:LogError (object)
Crest.SimSettingsAnimatedWaves:CreateCollisionProvider () (at E:/crest/crest/Assets/Crest/Crest/Scripts/LodData/Settings/SimSettingsAnimatedWaves.cs:59)
Crest.OceanRenderer:CreateDestroySubSystems () (at E:/crest/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs:656)
Crest.OceanRenderer:OnEnable () (at E:/crest/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs:438)
 
(Filename: E:/crest/crest/Assets/Crest/Crest/Scripts/LodData/Settings/SimSettingsAnimatedWaves.cs Line: 59)

Crest: Depth cache will not be populated at runtime when in batched/headless mode. Please pre-bake the cache in the Editor.
UnityEngine.StackTraceUtility:ExtractStackTrace ()
UnityEngine.DebugLogHandler:LogFormat (UnityEngine.LogType,UnityEngine.Object,string,object[])
UnityEngine.Logger:Log (UnityEngine.LogType,object)
UnityEngine.Debug:LogWarning (object)
Crest.OceanDepthCache:PopulateCache (bool) (at E:/crest/crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs:282)
Crest.OceanDepthCache:Start () (at E:/crest/crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs:103)
 
(Filename: E:/crest/crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs Line: 282)
`

Are you seeing other stuff? Maybe i could rope you in to looking at them if they are simple to patch up, but if they look trickier send them my way. Thanks!

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
huwb: damn the internet for being wrong. made a second attempt and it seems to be good now, thanks

daleeidd (approved):
Nice. Works well.

I couldn't reproduce that issue I mentioned earlier. I don't know what happened there.

huwb:
Awesome

TiToMoskito:
Is there any ETA when this will be available on the URP version?
Or can i donwload the latest version from the package manager and replace it with the files from master branch?

PR #813, opened 2021-05-01
Fix build errors for platforms without XR/VR support

daleeidd:
Xbox throws build errors here in more recent Unity versions according to reports. Easiest way to develop guards against this is to just disable the VR and/or XR modules.


huwb (approved):
Looks good nice one

daleeidd:
Thanks!

PR #815, opened 2021-05-04
Fix black "square" on Quest

daleeidd:
Issue

Adding MegaBoomers fixes to the issue above. Thank you for the fix.

I also propagated the type change as PSSL doesn't like out parameters being different.

Review is best done by commit as I keep those two changes separate.


huwb (approved):
Great to get this fixed!

PR #816, opened 2021-05-06
Only show enabled spectrums in debug GUI

daleeidd:
The spectrum part of the debug GUI doesn't take component enabled/disabled into account. I went with the simplest and lowest overhead route of just having a static list on both ShapeGerstners. Only downsides are that there is an overhead (negligible I think) even if OceanDebugGUI is not used.

In 2020 *FindObjectsOfType* can exclude inactive objects, but we would still have to call this every frame or implement on hierarchy change event if we wanted this to be dynamic. But may be worth it.

Let me know what you think.


huwb (approved):
Nice

daleeidd:
Thanks.

PR #817, opened 2021-05-06
Add power to range attribute

daleeidd:
Unity has an internal PowerSlider. So I just grabbed it with reflection. It should be safe to use as I found it in the [C# reference](https://github.com/Unity-Technologies/UnityCsReference/blob/master/Editor/Mono/EditorGUI.cs#L2438) which is for 2020.2. Also, [materials have a PowerSlider attribute](https://docs.unity3d.com/2021.2/Documentation/ScriptReference/MaterialPropertyDrawer.html) available which I believe uses this method too.

I have set the power to 2 for wind speed. I find higher values like 3 has a dead zone at the start. Let me know if you want it changed.


huwb (approved):
interesting! looks great

PR #818, opened 2021-05-06
Cutscene and Networked Time Providers

huwb:
**Cutscene Time Provider**

A time provider that `OnEnable` registers itself and drives the Timeline time into the system. `OnDisable` it unregisters itself so cutscenes can be loaded/unloaded on the fly.

To make this push/pop of TPs work elegantly, the runtime TP storage is a list that behaves loosely as a stack. Pushing a TP always moves it to the top of the stack, and popping a TP removes it from anywhere in the stack. This way TPs always activate when they are pushed, and when removed the system has a memory of the previous TP to fallback to. This applies to user-assigned TPs as well.

**Networked Time Provider**

After years of assuming this was a complicated thing, i've realised this probably just needs a time provider with a configurable offset, so just a few lines of code.

**Custom Time Provider - pause support**

Got a support request for how to implement pausing, and realised that using the Custom TP requires unnecessary effort. I've simply added a "pause" option which will freeze time and probably achieve 99% of use cases. I also put the override time/dt on toggles, as specifying an actual fixed time is likely a uncommon scenario. I turned on the overrides by default for now, which makes me a little bit sad, but will mean that everyones existing custom TPs will continue working for them in the heavy handed mode. I'll queue myself a task to turn off the overrides by default for next release.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Time/TimeProviderCutscene.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Time/TimeProviderCutscene.cs
@@ -12,6 +12,8 @@ namespace Crest
     /// This time provider feeds a Timeline time to the ocean system, using a Playable Director
     /// </summary>
     [ExecuteAlways]
+    [AddComponentMenu(Internal.Constants.MENU_PREFIX_SCRIPTS + "Cutscene Time Provider")]
+    [HelpURL(Internal.Constants.HELP_URL_BASE_USER + "other-features.html" + Internal.Constants.HELP_URL_RP + "#time-providers")]
     public class TimeProviderCutscene : TimeProviderBase
     {
         [Tooltip("Playable Director to take time from"), SerializeField]
--- a/crest/Assets/Crest/Crest/Scripts/Time/TimeProviderCutscene.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Time/TimeProviderCutscene.cs
@@ -1,0 +1,143 @@
+﻿// Crest Ocean System
+
+// This file is subject to the MIT License as seen in the root of this folder structure (LICENSE)
+
+using UnityEditor;
+using UnityEngine;
+using UnityEngine.Playables;
+
+namespace Crest
+{
+    /// <summary>
+    /// This time provider feeds a Timeline time to the ocean system, using a Playable Director
+    /// </summary>
+    [ExecuteAlways]
+    public class TimeProviderCutscene : TimeProviderBase
+    {
+        [Tooltip("Playable Director to take time from"), SerializeField]
+        PlayableDirector _playableDirector;
+
+        [Tooltip("Time offset which will be added to the Timeline time"), SerializeField]
+        float _timeOffset = 0f;
+
+        [Tooltip("Auto-assign this time provider to the ocean system when this component becomes active"), SerializeField]
+        bool _assignToOceanComponentOnEnable = true;
+
+        [Tooltip("Restore whatever time provider was previously assigned to ocean system when this component disables"), SerializeField]
+        bool _restorePreviousTimeProviderOnDisable = true;
+
+        TimeProviderDefault _fallbackTP = new TimeProviderDefault();
+
+        TimeProviderBase _previousTimeProviderComponent;
+        ITimeProvider _previousTimeProvider;
+
+        bool _initialised = false;
+
+        private void OnStart()
+        {
+            Initialise(true);
+        }
+
+        void Initialise(bool showErrors)
+        {
+            if (OceanRenderer.Instance == null)
+            {
+                if (showErrors)
+                {
+                    Debug.LogError("Crest: No ocean present, TimeProviderCutscene will have no effect.", this);
+                }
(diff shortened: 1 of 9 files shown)


daleeidd:
I am trying to test the timeline one. Been following these [instructions](https://docs.unity3d.com/Packages/com.unity.timeline@1.2/manual/wf_rec_anim.html) but I am getting overflow exceptions:
```
StackOverflowException: The requested operation caused a stack overflow.
UnityEngine.Object.GetCachedPtr () <0x15b054810 + 0x00004> in <0dab89ab297e40cca227f0e0fb130392>:0
UnityEngine.Object.IsNativeObjectAlive (UnityEngine.Object o) (at /Users/bokken/buildslave/unity/build/Runtime/Export/Scripting/UnityEngineObject.bindings.cs:156)
UnityEngine.Object.CompareBaseObjects (UnityEngine.Object lhs, UnityEngine.Object rhs) (at /Users/bokken/buildslave/unity/build/Runtime/Export/Scripting/UnityEngineObject.bindings.cs:135)
UnityEngine.Object.op_Inequality (UnityEngine.Object x, UnityEngine.Object y) (at /Users/bokken/buildslave/unity/build/Runtime/Export/Scripting/UnityEngineObject.bindings.cs:396)
Crest.TimeProviderCutscene.get_CurrentTime () (at Assets/Crest/Crest/Scripts/Time/TimeProviderCutscene.cs:118)
Crest.TimeProviderCutscene.get_CurrentTime () (at Assets/Crest/Crest/Scripts/Time/TimeProviderCutscene.cs:134)
Crest.TimeProviderCutscene.get_CurrentTime () (at Assets/Crest/Crest/Scripts/Time/TimeProviderCutscene.cs:134)
etc
```

I am a noob with the timeline so let me know if I am not testing the right thing.

huwb:
Could you perhaps put this year data into main.unity on this branch?

I think I'm probably setting it up differently but I'd be keen to try your setup!

daleeidd:
Looks like it is because I assigned the cutscene time provider to the time provider slot with the default checkboxes enabled on the time provider (assign and restore). On the second enter play mode the stack overflow occurs. If you cannot reproduce then I will add the test data.

huwb:
Yeah i can, sigh. Its a bit of a tangled mess of logic, by the time you account for play/edit mode and the various states. if it was runtime only it would be much easier.

i almost fixed its still failing in some situations and im stuck.

two ideas:

* make TimeProviderCutscene not actually inherit from TimeProviderBase itself, so it cant actually be assigned ot the ocean component. pretty sure this will sort it but will probably confuse people
* try to simplify the override system, which will *probably* address this too. maybe have yet another override slot that is just used for cutscenes and etc that need to shove themselves in and out. i guess the most general data structure is a stack where TPs can be pushed and popped, but not sure going full stack is necessary. maybe ill play with this..

daleeidd:
Why do we have to treat the cutscene one differently? Is there an issue with having it assigned in edit mode?

huwb:
the concept is that i can, at any time during play mode, instantiate a cutscene prefab and it will "just work" - its cutscene time provider will activate and the water will always follow the cutscene time. when the PD finishes playing things should return to "normal", where normal is whatever the user had assigned previously.

i implemented a loose stack structure and it seems to have worked out well. if a TP is pushed, it goes to the top of the stack and becomes active. if it was already in the stack, this always bumps it to the top. if a TP is popped, it is removed from the stack. this means assigning a TP always makes it active, but a "history" of TPs are kept and once the current TP is removed, it falls back to last active. on startup, a default TP is added which will never be removed, so there is always a fallback in the stack.

seems to work. let me know what you think!

crest/Assets/Crest/Crest/Scripts/Time/TimeProviderCutscene.cs
```diff
@@ -0,0 +1,119 @@
...
+        // If there is a playable director which is playing, return its time, otherwise
+        // use whatever TP was being used before this component initialised, else fallback
+        // to a default TP.
+        public override float CurrentTime
+        {
+            get
+            {
+                if (_playableDirector != null
+                    && _playableDirector.isActiveAndEnabled
+                    && (!EditorApplication.isPlaying || _playableDirector.state == PlayState.Playing))
```
daleeidd, line 106: Build error here. Could use _Application.isPlaying_ instead since it does the same thing when reading.

daleeidd (changes requested):
Nice one. Looks good. Testing in editor and it worked. Just a build error needs fixing and release notes.

crest/Assets/Crest/Crest/Scripts/Time/TimeProviderCutscene.cs
huwb: did not know about Application.isPlaying, nice. done

huwb:
Thanks, added release notes. The TPs deserve a section in the docs but i will look at that later once the code is out of the way.

daleeidd (approved):
Nice one.

huwb:
Thanks :)

PR #821, opened 2021-05-09
Make interaction scripts behave gracefully when dyn wave sim disabled

huwb:
Right now sphere water interaction and others spew errors if dyn wave sim is disabled.

This change emits a *warning*, only as a helpbox, if the sim is not found, and behaves otherwise gracefully and does not pollute the log.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Interaction/ObjectWaterInteraction.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Interaction/ObjectWaterInteraction.cs
@@ -56,6 +56,13 @@ namespace Crest
             }
 #endif
 
+            if (OceanRenderer.Instance._lodDataDynWaves == null)
+            {
+                // Don't run without a dyn wave sim
+                enabled = false;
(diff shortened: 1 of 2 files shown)


daleeidd:
Wouldn't these logs prevent possible support requests? I cannot think of why one would have these without dynamic waves enabled.

huwb:
The scenario i had in mind is someone using one of their boat prefabs in a level where the dyn wave sim is disabled.

Another scenario that just came to mind is someone might have a Low quality setting where the dyn wave sim is disabled (as one would do on low-to-mid mobile). In this case the same data (prefab) would need to work whether a dyn wave sim has been created or not, and the lack of a sim should not make the data 'erroneous'.

It's possible though that support requests come from this. RegisterDynWaves still has an error box in the inspector, and in this PR the removed errors are still present as warning boxes, so maybe these are enough?

daleeidd (approved):
Looks good. Nice work.

huwb:
Cheers

PR #822, opened 2021-05-09
Mellow out wave attenuation from wind

huwb:
After more experience and tweaking im finding the way the wind attenuates waves (linear ramp starting at 95%) is quite harsh leading to waves seemingly popping in when wind speed changes, and linear nature of ramp is visible.

This change starts transitioning waves at 75% and uses a ease-in-ease-out curve to soften the transition.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Shapes/OceanWaveSpectrum.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Shapes/OceanWaveSpectrum.cs
@@ -150,8 +150,7 @@ namespace Crest
 
             // Dampen based on wind. Waves travelling faster than wind get dampened. 0.95
             // is arbitrary value to give some 'ramp' in the multiplier.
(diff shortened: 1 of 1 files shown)


crest/Assets/Crest/Crest/Scripts/Shapes/OceanWaveSpectrum.cs
```diff
@@ -150,8 +150,7 @@ public float GetAmplitude(float wavelength, float componentsPerOctave, float win
 
             // Dampen based on wind. Waves travelling faster than wind get dampened. 0.95
```
daleeidd, line 151: 0.95 should be 0.75 here

daleeidd (approved):
Very nice. That is a lot better

crest/Assets/Crest/Crest/Scripts/Shapes/OceanWaveSpectrum.cs
huwb: thanks

PR #826, opened 2021-05-12
Use InputSystem when enabled in build settings

sagron:
Can probably be cleaned up more using variables, but for now a straight drop-in replacement around input code that shouldn't impact existing functionality.

### DIFF
--- a/crest/Assets/Crest/Crest-Examples/BoatDev/Scripts/BoatAlignNormal.cs
+++ b/crest/Assets/Crest/Crest-Examples/BoatDev/Scripts/BoatAlignNormal.cs
@@ -4,6 +4,9 @@
 
 using Crest;
 using UnityEngine;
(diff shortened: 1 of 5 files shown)


daleeidd (changes requested):
Thanks. All of it looks great. You're just missing a couple of files to make this complete:
- _Crest/Crest/Scripts/Interaction/BoatProbes.cs_
- _Development/Scripts/TestCameraChange.cs_
  - When you add the input reference to the Development asmdef, uncheck "use GUID" and ignore the GUID that is already there

daleeidd:
@huwb I think I will just merge this in and finish the rest in master. Is this okay?

huwb:
Sure

daleeidd:
Thanks. And thanks @sagron for the PR.

PR #827, opened 2021-05-12
Fix ocean tiles moving away from camera when far from zero

daleeidd:
Only affects URP when _Enable GPU Instancing_ is enabled. Since _UNITY_MATRIX_M_ is the same as _unity_ObjectToWorld_, and we use it already in _SnapAndTransitionVertLayout_, it is a no brainer to apply the fix in built-in.

What was happening was the gap fix was pushing tiles away from camera. Must be a bug with URP.


huwb (approved):
Nice one

daleeidd:
Thanks

PR #829, opened 2021-05-13
Multiple Oceans: Connect underwater to ocean instance

daleeidd:
This is to be merged into #650.

I had this in my stash. I tested with two ocean instances and two cameras. Appears to work. I have added test data with two camera setup.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Helpers/UnderwaterEffect.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Helpers/UnderwaterEffect.cs
@@ -44,6 +44,8 @@ namespace Crest
         PropertyWrapperMPB _mpb;
         Renderer _rend;
 
+        [SerializeField] Camera _camera;
+
         readonly int sp_HeightOffset = Shader.PropertyToID("_HeightOffset");
 
         SampleHeightHelper _sampleWaterHeight = new SampleHeightHelper();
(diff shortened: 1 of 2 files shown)


crest/Assets/Crest/Crest/Scripts/Helpers/UnderwaterEffect.cs
```diff
@@ -59,12 +61,13 @@ private void Start()
 #endif
             _rend = GetComponent<Renderer>();
 
+            _camera = transform.parent.gameObject.GetComponent<Camera>();
```
huwb, line 64: should this be both in the inspector (its serialised above), and also set from code? i'd expect one of or the other, or setting from code only if its not populated by the user.

it might be better to use GetComponentInParent, although it can work like this too

huwb (approved):
Approving this, even though my comments suggest change. Given this is going into a dev branch and will probably be reviewed again i dont mind too much.

Oh and is main.unity changes intentional? I'm fine with them if they are.

crest/Assets/Crest/Crest/Scripts/Helpers/UnderwaterEffect.cs
daleeidd: I have removed serialisation as validation already enforces parent camera relationship. Also now using GetComponentInParent. Thanks.

daleeidd:
Thanks.

> Oh and is main.unity changes intentional? I'm fine with them if they are.

They were as it sets up two cameras for testing. But a have dropped that commit as I have instead duplicated the scene to an git ignored folder for testing instead. This should make merging master into your branch easier.

PR #830, opened 2021-05-14
Fix "fix" helper exception and fix help box not showing

daleeidd:
Fixed a bad merge which resulted in the "remove mesh renderer" help box not showing when it should be. Also fixes an exception because we pass in either a component or GO to some of the fix functions.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/OceanValidation.cs
+++ b/crest/Assets/Crest/Crest/Scripts/OceanValidation.cs
@@ -74,7 +74,10 @@ namespace Crest
         internal static void FixAttachComponent<ComponentType>(SerializedObject lodInputComponent)
             where ComponentType : Component
         {
-            var gameObject = lodInputComponent.targetObject as GameObject;
+            // We will either get the component or the GameObject it is attached to.
+            var gameObject = lodInputComponent.targetObject is GameObject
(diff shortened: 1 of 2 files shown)


huwb:
Thanks. That's a bit of an unfortunate thing with the SO approach, that this can happen..

I tried to make it explicit via the param name that it should be a component passed in - `lodInputComponent`. I don't know if whatever is passing a GO could be passing a component instead, or if it could be renamed to `gameObject` and everything passes in GOs?

Basically id prefer the argument name to say what is expected for these things, and to perhaps Assert we're getting what we expect. Because the only way to know what to pass is from the argument name.

On the other hand if it needs to support both then I'd vote for leaning into it and renaming the arg to this beauty `lodInputComponentOrGameObject` !

huwb (changes requested):
Seemed to work in a few tests. So just the nitpick comment!

huwb:
Thinking further it seems like some things will want to pass in components and some things GOs to this function, if you agree then we should just rename the input and keep the code you have

daleeidd:
> Thinking further it seems like some things will want to pass in components and some things GOs to this function, if you agree then we should just rename the argument and keep the code you have

Sounds good. Done.

huwb (approved):
Nice one

daleeidd:
Thanks.

PR #831, opened 2021-05-15
Hash generation settings

huwb:
Creates a hash from ocean generation settings and uses it to auto rebuild, removing need for rebuild button.

Much more intuitive to use.

I'm not expert on hashing but i think its doing the right thing.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
+++ b/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
@@ -144,7 +144,7 @@ namespace Crest
         internal Material _material = null;
         public Material OceanMaterial { get { return _material; } }
 
-        [SerializeField]
+        [SerializeField, Delayed]
         string _layerName = "Water";
         public string LayerName { get { return _layerName; } }
 
@@ -764,6 +764,8 @@ namespace Crest
             Hashy.Add(_lodCount, ref settingsHash);
             Hashy.Add(_forceBatchMode, ref settingsHash);
             Hashy.Add(_forceNoGPU, ref settingsHash);
+            Hashy.Add(_hideOceanTileGameObjects, ref settingsHash);
+            Hashy.Add(_layerName, ref settingsHash);
 
             return settingsHash;
         }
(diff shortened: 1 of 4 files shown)


crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
```diff
@@ -750,8 +754,31 @@ void LateUpdate()
...
+        {
+            var settingsHash = Hashy.CreateHash();
+
+            // Add all the settings that require rebuilding..
+            Hashy.Add(_lodDataResolution, ref settingsHash);
+            Hashy.Add(_geometryDownSampleFactor, ref settingsHash);
+            Hashy.Add(_lodCount, ref settingsHash);
+            Hashy.Add(_forceBatchMode, ref settingsHash);
+            Hashy.Add(_forceNoGPU, ref settingsHash);
+
```
daleeidd, line 767: Missing _layerName and _hideOceanTileGameObjects.

daleeidd (changes requested):
Clever way of solving that problem. Nice work. Just missing a couple of properties which require a rebuild. Also just needs change log updated.

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
huwb: thanks, nice catch.

the latter was easy to add.

the former unfortunately has a very unsatisfying solution. string is an object and GetHashCode returns the instance number, nothing to do with the contents of the string. i searched around and didnt see a convenient solution. i could convert the string to bytes then hash each byte or such. im concerned that will allocated, or require some hoop jumping to do in a non-allocating way.

i think returning the instance number is working-enough, as the string changing allocates a new string object i guess. seems to work in tests.

on a slight tangent, i guess this is only a string for reasons related to "AssignLayer" and the things around that. BUT there is a well defined Water layer in Unity as of 2019, which im almost certain everyone uses. maybe i should just change it to be a layer selection field? then it would hash properly..

daleeidd: Agreed on changing to a layer selection field. Been wanting to do that for a while now. We'll go with this and do that in a separate PR? I've added it to Trello.

huwb: cool sounds good to me

PR #832, opened 2021-05-17
Allow set ocean mat

huwb:
From a support request - allow setting ocean material from code.

Changing it at runtime works for me so i think it was probably just waiting for the request ot make it public.

(It was slightly tempting to make the setter apply to all tiles, rather than have the tiles poll the material many times pre frame, but that should probably be a separate change).


daleeidd (approved):
Nice one.

PR #835, opened 2021-05-21
Add FFT waves

huwb:
**Status**: This is a transfer over from our private repos and is ready to merge from my POV

* Adds multi-scale FFT. Wave generation in this setup takes 28us on my 1060
* Generalises debug gui texture view to allow drawing any texture array
* Generalise debug gui texture view to allow scale/bias to data. Useful when data can go negative, or has a large or small range
* Supports spectrum control over FFT. Further, Pierson Moskowitz was used to calibrate our gerstner spectrum with the same spectrum using FFT.
* The calibration changed the gerstner spectrum. I've introduced a new pattern for migrating the spectrum data.

**Testing**:

Add ShapeFFT to a GameObject in the scene and it will add FFT waves.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
@@ -192,7 +192,7 @@ namespace Crest
             UpdateEditorOnly();
 #endif
 
-            if (_compute == null || _waveBuffers == null)
+            if (_compute == null || _waveBuffers == null || _resolution != _waveBuffers.width)
             {
                 InitData();
             }
@@ -202,7 +202,7 @@ namespace Crest
             if (!EditorApplication.isPlaying) updateDataEachFrame = true;
 #endif
             // Ensure batches assigned to correct slots
-            if (_firstUpdate || updateDataEachFrame || (_waveBuffers != null && _resolution != _waveBuffers.width))
+            if (_firstUpdate || updateDataEachFrame || (_waveBuffers != null))
             {
                 InitBatches();
 
--- a/crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
@@ -42,7 +42,7 @@ namespace Crest
 
         [Tooltip("When true, uses the wind speed on this component rather than the wind speed from the Ocean Renderer component.")]
         public bool _overrideGlobalWindSpeed = false;
-        [Tooltip("Wind speed in km/h. Controls wave conditions."), Range(0, 150f), Predicated("_overrideGlobalWindSpeed")]
+        [Tooltip("Wind speed in km/h. Controls wave conditions."), Range(0, 150f, 2f), Predicated("_overrideGlobalWindSpeed")]
         public float _windSpeed = 20f;
 
         [Tooltip("Multiplier for these waves to scale up/down."), Range(0f, 1f)]
--- a/crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
@@ -18,7 +18,7 @@ namespace Crest
     /// </summary>
     [ExecuteAlways]
     [AddComponentMenu(Internal.Constants.MENU_PREFIX_SCRIPTS + "Shape FFT")]
-    [HelpURL(Internal.Constants.HELP_URL_BASE_USER + "wave-conditions.html" + Internal.Constants.HELP_URL_RP + "#shapefft")]
+    [HelpURL(Internal.Constants.HELP_URL_BASE_USER + "wave-conditions.html" + Internal.Constants.HELP_URL_RP + "#shapefft-preview")]
     public partial class ShapeFFT : MonoBehaviour, LodDataMgrAnimWaves.IShapeUpdatable
 #if UNITY_EDITOR
         , IReceiveSplinePointOnDrawGizmosSelectedMessages
--- a/crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
@@ -28,9 +28,6 @@ namespace Crest
         [Tooltip("Impacts how aligned waves are with wind.")]
         [Range(0, 1)]
         public float _windTurbulence = 0.145f;
-        [Tooltip("Choppiness of waves.")]
-        [Range(0, 2.5f)]
-        public float _chop = 1f;
 
         [Tooltip("The spectrum that defines the ocean surface shape. Assign asset of type Crest/Ocean Waves Spectrum."), Embedded]
         public OceanWaveSpectrum _spectrum;
@@ -220,7 +217,7 @@ namespace Crest
             ReportMaxDisplacement();
 
             var windSpeedMPS = (_overrideGlobalWindSpeed ? _windSpeed : OceanRenderer.Instance._globalWindSpeed) / 3.6f;
-            _compute.GenerateDisplacements(buf, _windTurbulence, windSpeedMPS, OceanRenderer.Instance.CurrentTime, _chop, _activeSpectrum, updateDataEachFrame, _waveBuffers);
+            _compute.GenerateDisplacements(buf, _windTurbulence, windSpeedMPS, OceanRenderer.Instance.CurrentTime, _activeSpectrum, updateDataEachFrame, _waveBuffers);
 
             buf.SetGlobalVector(sp_AxisX, PrimaryWaveDirection);
             // Seems to come unbound when editing shaders at runtime, so rebinding here.
--- a/crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
@@ -1,0 +1,422 @@
+﻿// Crest Ocean System
+
+// Copyright 2021 Wave Harmonic Ltd
+
+using UnityEngine;
+using UnityEngine.Rendering;
+using Crest.Spline;
+using UnityEngine.Experimental.Rendering;
+
+#if UNITY_EDITOR
+using UnityEditor;
+#endif
+
+namespace Crest
+{
+    /// <summary>
+    /// FFT ocean wave shape
+    /// </summary>
+    [ExecuteAlways]
+    [AddComponentMenu(Internal.Constants.MENU_PREFIX_SCRIPTS + "Shape FFT")]
+    [HelpURL("https://crest.readthedocs.io/en/latest/user/wave-conditions.html#shapefft")]
+    public partial class ShapeFFT : MonoBehaviour, LodDataMgrAnimWaves.IShapeUpdatable
+#if UNITY_EDITOR
+        , IReceiveSplinePointOnDrawGizmosSelectedMessages
+#endif
+    {
+        [Header("Wave Conditions")]
+        [Tooltip("Impacts how aligned waves are with wind.")]
+        [Range(0, 1)]
+        public float _windTurbulence = 0.145f;
+        [Tooltip("Choppiness of waves.")]
+        [Range(0, 2.5f)]
+        public float _chop = 1f;
+
+        [Tooltip("The spectrum that defines the ocean surface shape. Assign asset of type Crest/Ocean Waves Spectrum."), Embedded]
+        public OceanWaveSpectrum _spectrum;
+        OceanWaveSpectrum _activeSpectrum = null;
+
+        [Tooltip("When true, the wave spectrum is evaluated once on startup in editor play mode and standalone builds, rather than every frame. This is less flexible but reduces the performance cost significantly."), SerializeField]
+        bool _spectrumFixedAtRuntime = true;
+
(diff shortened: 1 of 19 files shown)


crest/Assets/Crest/Crest/Shaders/Resources/FFT/FFTSpectrum.compute
```diff
@@ -0,0 +1,217 @@
...
+	const int2 center = _Size.xx / 2;
+	const int2 coord = id.xy - center;
+
+	uint depth;
+	{
+		uint width, height;
+		_ResultInit.GetDimensions( width, height, depth );
+	}
+
+	if( id.z < (depth - 1) && max( abs( coord.x ), abs( coord.y ) ) < _Size / 4 )
```
daleeidd, line 111: Metal complains here about mismatching signs as it assumes `depth - 1` is signed.

crest/Assets/Crest/Crest/Shaders/Resources/FFT/FFTCompute.compute
```diff
@@ -0,0 +1,161 @@
...
+	else
+	{
+		indexA = coord;
+		indexB = coord + offset;
+	}
+
+	if (passIndex == 0)
+	{
+		indexA = reversebits(indexA) >> (32 - PASSES);
+		indexB = reversebits(indexB) >> (32 - PASSES);
```
daleeidd, line 66: This needs the _reversebits_ fallback for it to work on older MacOS devices.

daleeidd (changes requested):
I have brought over the review items from the private repository. Will do a more thorough review once added.

crest/Assets/Crest/Crest/Shaders/Resources/FFT/FFTSpectrum.compute
huwb: missed that before. i think i've pushed a fix now

crest/Assets/Crest/Crest/Shaders/Resources/FFT/FFTCompute.compute
huwb: sorry i for some reason thought it was already merged in but did not check! brain fog. added now. seems good on windows at least..

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
```diff
@@ -0,0 +1,422 @@
...
+#endif
+
+namespace Crest
+{
+    /// <summary>
+    /// FFT ocean wave shape
+    /// </summary>
+    [ExecuteAlways]
+    [AddComponentMenu(Internal.Constants.MENU_PREFIX_SCRIPTS + "Shape FFT")]
+    [HelpURL("https://crest.readthedocs.io/en/latest/user/wave-conditions.html#shapefft")]
```
daleeidd, line 21: This should be (for versioning and RP selection):
```csharp
[HelpURL(Internal.Constants.HELP_URL_BASE_USER + "wave-conditions.html" + Internal.Constants.HELP_URL_RP + "#shapefft")]
```

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
```diff
@@ -0,0 +1,422 @@
...
+        [Delayed, Tooltip("How many wave components to generate in each octave.")]
+        public int _componentsPerOctave = 8;
+
+        [HideInInspector]
+        [Tooltip("Change to get a different set of waves.")]
+        public int _randomSeed = 0;
+
+        [Header("Generation Settings")]
+        [Tooltip("Resolution to use for wave generation buffers. Low resolutions are more efficient but can result in noticeable patterns in the shape."), Delayed]
+        public int _resolution = 32;
```
daleeidd, line 67: If I change resolution, the shape looks incorrect until I disable/enable the component. It might need something like the auto-rebuild that OceanRenderer has?

crest/Assets/Crest/Crest/Scripts/Shapes/OceanWaveSpectrum.cs
```diff
@@ -16,13 +16,20 @@ namespace Crest
     [CreateAssetMenu(fileName = "OceanWaves", menuName = "Crest/Ocean Wave Spectrum", order = 10000)]
     public class OceanWaveSpectrum : ScriptableObject
     {
+        /// <summary>
+        /// The version of this asset. Can be used to migrate across versions. This value should
+        /// only be changed when the editor upgrades the version.
+        /// </summary>
+        [SerializeField, HideInInspector]
+        int _version = 0;
```
daleeidd, line 24: Just bringing over my reply in private repository about setting the version to 1:
> You could use Reset to set the version to 1? Since that will only be called on new instances or when the use hits reset button which should then auto upgrade anyway?

daleeidd (changes requested):
Great effort pushing FFTs forward. They look fantastic. Just a few minor review things.

I think we still need to add the Apache license in the FFT folder? From what I read, it looks like your copyright header etc is enough regarding state changes.

Last thing is release notes. I am guessing you'll want to add documentation after merging?

crest/Assets/Crest/Crest/Shaders/Resources/FFT/FFTSpectrum.compute
```diff
@@ -0,0 +1,217 @@
...
+	// Spectrum
+	float spectrum = PiersonMoskowitzSpectrum( w );
+
+	float wl = PI2 / kMag;
+	float SMALLEST_WL_POW_2 = -4.0;
+	uint NUM_OCTAVES = 14;
+	float index = log2( wl ) - SMALLEST_WL_POW_2;
+	spectrum = _SpectrumControls.SampleLevel( linear_clamp_sampler, float2((index + 0.5) / NUM_OCTAVES, 0.5), 0.0 )
+		* PiersonMoskowitzWindTerm( w )
+		;
```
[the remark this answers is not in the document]
huwb: oops this whole block of code was half-baked, cleaned up now

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
huwb: oops, thanks

crest/Assets/Crest/Crest/Scripts/Shapes/OceanWaveSpectrum.cs
huwb: hmm, need to think more about it

huwb: this was the missing piece. good shout!

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
huwb: thanks fixed now, was not re-init'ing the res properly

huwb:
Ok i tied up a number of other loose ends and implemented the suggestions, thanks again for those. Let me know what you think.

daleeidd (changes requested):
I added a couple of commits to fix/improve some minor documentation stuff. Not sure if release notes covers all changes. A few fixes needed.

With the spectrum upgrade, will it significantly change the result of already authored spectrums? I noticed that our main example scene is pretty intense now.

huwb:
thanks for help! your changes look great

yeah i struggled to get headspace to tackle the docs. I don't think the "Wave Conditions" section will make any sense at all at the moment. I'll try to do another pass soon, or do a rewrite for next release.

> With the spectrum upgrade, will it significantly change the result of already authored spectrums? I noticed that our main example scene is pretty intense now.

d'oh.. would you have a repro for that? i think i've seen that before but testing now i was unable to make it happen. it's worrying if you're getting that and i'm starting to think the approach to versioning data that i tried here was too bold (complicated). contingency plan is to not change the current data, except add the `_version` field, and migrate the data in a 2 or 3 releases. would be much simpler i believe, but may require retweaks if people jump from an old version to new..

daleeidd:
> d'oh.. would you have a repro for that? i think i've seen that before but testing now i was unable to make it happen. it's worrying if you're getting that and i'm starting to think the approach to versioning data that i tried here was too bold (complicated). contingency plan is to not change the current data, except add the `_version` field, and migrate the data in a 2 or 3 releases. would be much simpler i believe, but may require retweaks if people jump from an old version to new..

I was able to reproduce it. Looks to be an edge case. Steps:
1. Open main scene
2. Save the spectrum so upgraded data is saved
3. Make minor change in scene and save
4. Do a `git reset --hard` or checkout those two files (main.unity and WavesModerate.asset) to revert them
5. Unity will prompt "scene reload" or ignore. Choose scene reload

The waves will look like the following:
Screen Shot 2021-05-24 at 3 05 22 am

I don't know how I incurred the issue earlier and if it is something anyone else will hit. I inspected the spectrum after saving again and the version was set to 1 but the powers were not changed. I added a log line and it was the Upgrade method which increments the version in this edge case (not Reset).

huwb:
thanks! after poking around i found another repro:

* show editor for a spectrum - upgrades it
* save
* revert the spectrum asset in git

if i remove the `HideInInspector` attribute from the `_version` field, i can see it does not go back to 0 when i revert the asset. but restarting unity sorts it. nice job unity!

i guess we ship it with this in mind. hopefully aggressive reverting/reseting like we do is indeed an edge case.

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
```diff
@@ -0,0 +1,419 @@
...
+        [Tooltip("When true, the wave spectrum is evaluated once on startup in editor play mode and standalone builds, rather than every frame. This is less flexible but reduces the performance cost significantly."), SerializeField]
+        bool _spectrumFixedAtRuntime = true;
+
+        [Tooltip("Primary wave direction heading (deg). This is the angle from x axis in degrees that the waves are oriented towards. If a spline is being used to place the waves, this angle is relative ot the spline."), Range(-180, 180)]
+        public float _waveDirectionHeadingAngle = 0f;
+        public Vector2 PrimaryWaveDirection => new Vector2(Mathf.Cos(Mathf.PI * _waveDirectionHeadingAngle / 180f), Mathf.Sin(Mathf.PI * _waveDirectionHeadingAngle / 180f));
+
+        [Tooltip("When true, uses the wind speed on this component rather than the wind speed from the Ocean Renderer component.")]
+        public bool _overrideGlobalWindSpeed = false;
+        [Tooltip("Wind speed in km/h. Controls wave conditions."), Range(0, 150f), Predicated("_overrideGlobalWindSpeed")]
```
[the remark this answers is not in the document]
huwb: oh cool done, much better

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
```diff
@@ -0,0 +1,419 @@
...
+            _matGenerateWaves.SetFloat(sp_FeatherWaveStart, _featherWaveStart);
+            // Seems like shader errors cause this to unbind if I don't set it every frame. Could be an editor only issue.
+            _matGenerateWaves.SetTexture(sp_WaveBuffer, _waveBuffers);
+
+            ReportMaxDisplacement();
+
+            var windSpeedMPS = (_overrideGlobalWindSpeed ? _windSpeed : OceanRenderer.Instance._globalWindSpeed) / 3.6f;
+            _compute.GenerateDisplacements(buf, _windTurbulence, windSpeedMPS, OceanRenderer.Instance.CurrentTime, _activeSpectrum, updateDataEachFrame, _waveBuffers);
+
+            buf.SetGlobalVector(sp_AxisX, PrimaryWaveDirection);
```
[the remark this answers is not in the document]
huwb: ah good catch thanks

huwb:
Ok feedback addressed now i believe

daleeidd:
> thanks! after poking around i found another repro:
> 
> * show editor for a spectrum - upgrades it
> * save
> * revert the spectrum asset in git
> 
> if i remove the `HideInInspector` attribute from the `_version` field, i can see it does not go back to 0 when i revert the asset. but restarting unity sorts it. nice job unity!
> 
> i guess we ship it with this in mind. hopefully aggressive reverting/reseting like we do is indeed an edge case.

I narrowed it down to the fact that Unity does not recognise a property being deleted in an SO file. Must be a bug since Unity does recognise a property being deleted in scenes and materials. I will submit a bug report.

I added GitHub for Unity and tried to discard the changes made to the SO asset file to see if they handle this case, but they don't. Any competent developer will use some sort of VCS so I don't know the likeliness that someone will do all the wrong things to hit this issue. The tail end of the issue path being serialising bad data which would be version=1 and no upgrade applied.

huwb:
ok sounds good. lmk what you want to do with this pr in the meantime

daleeidd (approved):
Nice work. There are no other issues I can see. Considering we cannot fix that one problem there is nothing to change.

daleeidd:
Been trying to create a workaround but have been unsuccessful. I could not get FileSystemWatcher to reliably watch anything (on Mac at least). And since Unity ignores certain changes, the AssetPostprocessor doesn't run either. The idea was to read the file and see if version doesn't exist and add it.

I would normally say we shouldn't proceed with merging because I believe people will hit this. But I am not sure when Unity will fix this issue since the FileSystemWatcher not working as expected makes me concerned that the issue is worse than I reported. So waiting for Unity is not palatable.

I will put a note in the release notes warning people about this issue after it is merged.

huwb:
All makes sense. Merged. Thanks for all the help!

PR #836, opened 2021-05-22
Underwater Pass

daleeidd:
This is a draft and will need a refactor but it is working. I might refactor after merging.

This is an alternative to both _Image Effect_ (#299) and _Post Process_ (#663 + #678). It appears to be better than the other approaches. It will merge into #299.

@huwb Let me know what you think. @skaughtx0r would really like your opinion too since you have been using the post-processing branch.

## _Image Effect_ Problems/Disadvantages
- Renders after PPv2 stack so not compatible

## _Post Process_ Problems/Disadvantages
- Requires PPv2 package
- User setup is more involved
- XR support is hacky/smelly. See #678 for more information
- Higher technical debt
- Underwater effect is not really a post-process effect
- Harder to support transparent objects which also receive post-processing. One user has converted the HDRP PP to a custom pass for this very reason (and we probably should too)

## _Render Pass_ Problems/Disadvantages
- ~~Uses grab pass which makes it a bit more expensive (haven't done overall comparison though)~~
- None?

## Known Issues
- [x] Dynamic resolution breaks the pass. Need to look into this.


huwb:
Nice write up. Glad your handle on this stuff is infinitely better than mine!!

Thumbs up for this, from my side at least.

Couple of thoughts that occurred, but not blockers..

Why is grab pass needed - what is in the frame buffer at this time? Is it just opaques, or opaques plus water, or transparents too, or all of the above depending on when the pass is ran? I'm vaguely curious if the pass can be avoided in the future, and how it interacts with transparents. But the cost of the grab seems justified given the above pros/cons!

There's a procedural full screen tri which doesn't need a mesh lying around somewhere, and it's quite cool, but I can give pointers for that later, and it's a polish thing.

huwb:
(In the future when all the dust has settled we should add your notes above to our wiki)

daleeidd:
> Why is grab pass needed - what is in the frame buffer at this time? Is it just opaques, or opaques plus water, or transparents too, or all of the above depending on when the pass is ran? I'm vaguely curious if the pass can be avoided in the future, and how it interacts with transparents. But the cost of the grab seems justified given the above pros/cons!

I'll have to give it another go but had troubles uses the frame buffer.

> There's a procedural full screen tri which doesn't need a mesh lying around somewhere, and it's quite cool, but I can give pointers for that later, and it's a polish thing.

Sounds good.

> (In the future when all the dust has settled we should add your notes above to our wiki)

Sure thing.

daleeidd:
Just did another pass on this and grab pass has been replaced with CopyTexture/Blit. And the mesh has been replaced with the procedural approach which I grabbed from Unity.

huwb:
Ah cool. This looks good to me.

The reason i was curious about what we wanted to grab is because if we want to grab the same thing as what the surface grabs (opaque scene) then theres a good chance we might be able to get the buffer for 'free' by reusing the same one, just like all ocean tiles share the same grabpass data. But its not the end of the world to do an additional copy.

daleeidd:
Oh right. I did originally reuse the grab pass from the surface by accident which had no transparents (which we need).

huwb:
Oh if we need transps then needs to take a copy then. Ignore me!

skaughtx0r:
I just gave this branch a try and everything seems to work great for me.

PR #837, opened 2021-05-22
Fix bug where all ShapeGerstners used one wind heading

huwb:
Have no idea why this was set globally instead of per-material.


daleeidd (approved):

huwb:
cheers

PR #838, opened 2021-05-22
Make hashy functions type explicit

huwb:
While working on something else and attempting to hash an object instance ID, i found it converting an object ref to a bool (i guess doing an 'is null' operation' and calling `Add(bool)`, and failing to hash properly (!?). Wow.

Rather unfortunate but i could not find an elegant way in C# to avoid implicit conversions. I did find i could pass the first arg to `Add()` by reference, which may sort it, however passing stuff in by reference is probably hackier than doing this.


daleeidd (approved):
That is very obscure/sad. Solution looks good.

huwb:
Yeah.. thanks

PR #839, opened 2021-05-23
Make fields public for floaty things

daleeidd:
Made many fields public for *BoatProbes*, *BoatAlignNormal* and *SimpleFloatingObject*. I went through and made sure they were safe to make public (no state relied on them in Start etc).


huwb (approved):
Good stuff

PR #841, opened 2021-05-25
Update ocean underwater material colours

daleeidd:
Copies over the colours, specular and depth fog density from URP so they are similar. I also changed IOR of air to 1 since it is deprecated.

@huwb Did you want the other ocean material to have the same values?


huwb (approved):
Thanks looks good. I'm not really sure about the other materials. If you think it's better feel free, otherwise I reckon they'll work as is. I can also give them a review before we finalize if useful

daleeidd:
Thanks. No worries. I'll go with just this one for now.

PR #850, opened 2021-06-09
Fix InitBatches running every frame

daleeidd:
Reported by @skaughtx0r on Discord.

There was an extra check which caused a conditional to pretty much always be true resulting in *InitBatches* running every frame for FFTs.


huwb (approved):
Well caught, thanks

PR #852, opened 2021-06-11
Underwater clip pane

daleeidd:
This is only at the prototype stage. Would need to make the correct calculations for fog distance when fog is clipped still. Also still need to clip the ocean surface which could be done with a clip surface shader or using the same approach as this.

This approach uses distance so it converts depth to distance. Otherwise, the plane won't appear stable when camera moves around. It may have a minor affect on fog calculation.

@huwb Let me know what you think.


daleeidd:
Closing in favour of rendering from geometry.

PR #853, opened 2021-06-13
Smooth wave rotation

huwb:
Typically when changing wind direction, waves rotate around world origin which is bad. This gets reported regularly on discord.

The FFT is able to generate using any wind direction and makes smooth results. This change pipes through the wind direction.

When using a spline, alignment is already computed in the gerstner geometry input shader, so nothing was required for that.

As a bonus side effect, this resulted in a big cleanup and optimisation win in the spectrum init which is nice.


daleeidd (approved):
Nice work. Works really well. Just needs release notes.

huwb:
Thanks, added release note and merged

PR #854, opened 2021-06-13
Make FFTCompute own its generated data

huwb:
Right now, if N wave splines are used with some spectrum `Spectrum A`, the FFT will be run N times. The wave gen happens in a 'canonical' frame and if wind dir, speed etc are the same, the result will always be the same, so we should only generate the waves once in this case. I think we should allow many splines to be placed, so I think we should pursue this optimisation.

This is a preparation commit - it makes `FFTCompute` return the wave data.

The next step will be to make the `GenerateDisplacements` function static, and have it hash the input arguments and use that as a key into a generator dictionary. It can then return the data from the right generator. This PR moves the data into the generator as a first step.


daleeidd (approved):
Good idea. Looks good.

PR #857, opened 2021-06-16
Remove a few legacy workaround things

huwb:
Cleans out some legacy-support things:

* Default wind speed was set very high so that existing spectrums would not be impacted. Reducing it now so that any *newly created* OceanRenderer components have a wind speed of 10km/h
* `TimeProviderCustom` time override options (override time, override dt) were set true by default so existing instances of the component would override time just like previous did. I'm not changing it so that the overrides are false by default and are opt in. If not opted in, component still useful to pause time.
* Deleted a very old spectrum upgrade path, when we changed number of spectrum octaves. This should **not** impact the more recent spectrum upgrade/migration we did, and part of the reason to remove this old path is to reduce this confusion. I removed the `UpgradeSpectrum()` functions first, then only deleted further lines based on compile errors.

Future work

* Change the way `OceanWaveSpectrum` uses the `_version` field - simplify it once we're confident most people have a version in their data. Maybe in 2-3 releases.

Can't think if theres any more old workarounds like this, lmk if you think of any.


daleeidd (approved):
Looks good. I couldn't see any problems. The wind speed in the main scene needs to be adjusted if you want it the same as before. And release notes.

Other workarounds is blitting directly to texture arrays not being supported in 2018 in the TextureArrayHelper class when clearing to black. Might be supported by now.

huwb:
Cool good shout, blit is much nicer. Moved back to taht. I've removed the ClearToBlack.compute shader, i think removing rather than gutting is fine here.

Is there anything in particular you think we should broadcast from this to users in the changelog? I added a note that default wind speed is 10m/s, is that what you had in mind? I could try to expand it and say that it affects big wavelengths and so will change the look when new ocean components are added, but its a bit of a mouthful.

huwb:
Added a rel note for custom TP

daleeidd:
That's everything for the release notes. All looks good for merging.

huwb:
Thanks!

PR #858, opened 2021-06-17
Fix compile error building win64 player

huwb:
Fixes compile error building standalone development build in win64, 2020.3.10.

Trivial change, but i have no idea why I hit this in years-old code.

### DIFF
--- a/crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/CopyDepthIntoCache.shader
+++ b/crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/CopyDepthIntoCache.shader
@@ -59,7 +59,7 @@ Shader "Crest/Copy Depth Buffer Into Cache"
 
 			float4 frag(Varyings input) : SV_Target
 			{
-				float deviceDepth = SAMPLE_DEPTH_TEXTURE(_CamDepthBuffer, input.uv);
+				float deviceDepth = tex2D(_CamDepthBuffer, input.uv);
 				float linear01Z = CustomLinear01Depth(deviceDepth);
 
 				float altitude;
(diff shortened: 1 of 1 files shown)


crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/CopyDepthIntoCache.shader
```diff
@@ -59,7 +59,7 @@ Shader "Crest/Copy Depth Buffer Into Cache"
 
 			float4 frag(Varyings input) : SV_Target
 			{
-				float deviceDepth = SAMPLE_DEPTH_TEXTURE(_CamDepthBuffer, input.uv);
+				float deviceDepth = tex2D(_CamDepthBuffer, input.uv);
```
daleeidd, line 62: Should this have a swizzle for the red channel to avoid truncation warnings on some platforms?

daleeidd (changes requested):
Just one thing but I think it might be an issue. Also, do you remember what the error was?

daleeidd:
The error for reference:
> Shader error in 'Crest/Copy Depth Buffer Into Cache': undeclared identifier 'sampler_CamDepthBuffer' at line 62 (on d3d11)

huwb:
Ah thanks. should be good now

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/CopyDepthIntoCache.shader
huwb: good point done

daleeidd (approved):
Great. Nice one.

PR #859, opened 2021-06-17
Create default spectrum for standalone builds

### LINKED ISSUE #847: ShapeFFT flat in standalone build if no spectrum asset assigned
**Describe the bug**
ShapeFFT produces flat water in standalone build if no spectrum asset assigned. Reported by Bartolomeus.

huwb:
Fixes #847 

No default spectrum was being created if none assigned.

Unfortunately the waves appear way too large and i have no idea why. It looks a bit like issues we could get around the spectrum `_version` being wrong, but it happens for FFT too which should not rely on `_version`, so im quite confused. if you have ideas lmk. in any case this PR is required.


daleeidd:
This works for me:

```csharp
#if !UNITY_EDITOR
        void Awake()
        {
            // If it is version zero, and we are in a build, then it must be the default.
            if (_version == 0)
            {
                // Reset is called in the editor only so manually call here for standalone builds.
                Reset();
            }
        }
#endif
```

> It looks a bit like issues we could get around the spectrum _version being wrong, but it happens for FFT too which should not rely on _version

_version starts at zero so FFTs will use the upgrade path too.

I've noticed another case of this issue where switching the spectrum with the drawer closed produces the "too powerful waves" bug. I don't think it executes OnInspectorGUI and thus doesn't upgrade the spectrum correctly until the drawer is opened. Will look into this.

huwb:
I thought _version is only used for upgrading in editor code (not standalone), and _version is only used in run time code for getamplitude which is gerstner only.

However of it worked for you I'll happily go with it. I'll try to give it a test myself tonight. Thank you for jumping in, much appreciated.

daleeidd:
> I thought _version is only used for upgrading in editor code (not standalone)

From what I can see the default [_powerLog values](https://github.com/wave-harmonic/crest/blob/master/crest/Assets/Crest/Crest/Scripts/Shapes/OceanWaveSpectrum.cs#L51-L52) get reduced in editor code, and when creating the default spectrum in standalone they also need to be decreased somehow.

> and _version is only used in run time code for getamplitude which is gerstner only.

[_powerLog is used here for FFT](https://github.com/wave-harmonic/crest/blob/master/crest/Assets/Crest/Crest/Scripts/Shapes/FFT/FFTCompute.cs#L234) which is why it also affects FFTs. It looks like it runs in standalone too.

An alternative is to provide a different default array for standalone but either approach works.

daleeidd (approved):

PR #862, opened 2021-06-20
Clear TAs in a single call

huwb:
"Oops"

The old code is a very bad pattern - synchronously clearing one slice at a time. This change clears all texture array slices in a single call.

Checked clear is working in RD.


daleeidd (approved):
Great find.

huwb:
Its slightly slower lol. 760us on my 1060 vs 740us without. I cant really explain why, it might be the gpu is overlapping workloads effectively, so clearing all slices in one block is actually a touch slower.

In any case while its ugly im not sure i can merge this, and will close it!!

PR #863, opened 2021-06-20
Collision provider baked FFT

huwb:
This is hopefully the start of a new collision approach suggested by user lug on discord. It will generate a patch of waves that loop in time and space, at a set of time steps, and saves it into CPU memory. This could be written to file from editor code and loaded in server builds.

I've modified the spectrum to loop in time (and will already loop in space by construction).

This PR is branched off another branch i'm working called "fft wave generators". These refactor the fft wave gen into a separate object and allows the FFT to be run in a simple/flexible way. Hopefully soon that branch will merge and this PR probably wants to be recreated. In the meantime that dev branch should be functional and should be fine to build off.


huwb:
Replaced by #881

PR #866, opened 2021-07-01
Document Underwater Renderer

daleeidd:
Rearranges and rewrites a few things for the _Underwater Renderer_ being the primary underwater solution.

- Added _Underwater Renderer_ to underwater page
- Deprecated curtain and post-processing
- Changed underwater transparents for _Underwater Renderer_
- Added release notes

Here are the pages:
https://crest.readthedocs.io/en/docs-underwater-renderer/user/underwater.html
https://crest.readthedocs.io/en/docs-underwater-renderer/user/rendering.html
https://crest.readthedocs.io/en/docs-underwater-renderer/about/history.html

Let me know what you think.

### DIFF
--- a/docs/about/history.rst
+++ b/docs/about/history.rst
@@ -37,7 +37,7 @@ Preview
 ^^^^^^^
 .. bullet_list::
 
-   -  Add new Underwater Renderer component which executes a fullscreen pass between transparent and post-processing pass.
+   -  Add new *Underwater Renderer* component which executes a fullscreen pass between transparent and post-processing pass.
 
 Changed
 ^^^^^^^
--- a/docs/about/history.rst
+++ b/docs/about/history.rst
@@ -72,6 +72,10 @@ Performance
 
    -  FFT wave generation factored out so that multiple *ShapeFFT* components sharing the same settings will only run one FFT.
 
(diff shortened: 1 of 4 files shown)


docs/about/history.rst
```diff
@@ -23,12 +23,22 @@ Breaking
...
+
+      -  *Underwater Post-Processing* is disabled by default which means it will be inactive if the *Underwater Volume Override* is not present in the scene. `[HDRP]`
+
    .. only:: urp
 
       -  Remove *Sample Shadows* Render Feature as it is now scripted.
          Unity will raise a missing Render Feature reference error.
          Remove the missing Render Feature to resolve. `[URP]`
 
+Preview
```
huwb, line 36: I think since we're deprecating the old stuff this should be in the Changed section as it's not an optional/for testing preview thing

huwb (approved):
Minor change suggestion but approving so you can get on with it without a review cycle

docs/about/history.rst
daleeidd: Sure. Done.

daleeidd:
Thanks. Suggestion is done and added another release note that I missed.

PR #867, opened 2021-07-09
Underwater Fog Input

magenta404:
**_Status: Experimental_**

Adds an input for underwater fog which renders whatever it is attached to after the underwater effect and applies the fog in a separate pass. I have named the component _RegisterUnderwaterInput_ for consistency but it could be renamed.

The original idea was to use _Blend_ and _BlendOp_ to apply the fog but I wasn't able to work it out. I think using subtract might work if we somehow bake density fog and scatter colour into the fragment colour. Instead I rendered the object offscreen and provided texture to the fog pass so it samples itself to get the screen colour. This way it works like the underwater effect does.

Steps:
1. Register object and disable renderer
2. Sorts objects back to front using distance from camera
3. Enables SH and set SH values using [LightProbeUtility](https://github.com/keijiro/LightProbeUtility)
4. Setup main texture to get alpha channel
5. Copies camera target to temporary texture
6. Renders object into temporary texture
7. Passes temporary texture to fog and renders fog over the top using alpha blending

It didn't work out as nice as I hoped as it exposed us a little bit to render pipeline (SH and _MainTex). We need _MainTex to get the alpha value for blending; an example is the particles will be rendered as squares without it.

Also, I had to cull back faces as the example transparent objects in the test scene didn't like culling disabled (need to look into that), but some transparent objects want culling disable for them to look right.

I made the shader a separate pass rather than a separate file as it saves us another material and code reuse.

I have provided test data in the main scene.

Let me know what you think.


crest/Assets/Crest/Crest/Shaders/Underwater/Resources/UnderwaterEffect.shader
```diff
@@ -112,12 +114,102 @@ Shader "Hidden/Crest/Underwater/Underwater Effect"
...
-					sceneColour = ApplyUnderwaterEffect(scenePos, sceneColour, lightCol, lightDir, rawDepth, sceneZ, view, isOceanSurface);
+					sceneColour = ApplyUnderwaterEffect(_AmbientLighting, scenePos, sceneColour, lightCol, lightDir, rawDepth, sceneZ, view, isOceanSurface);
 				}
 
 				return half4(wt * sceneColour, 1.0);
 			}
 			ENDHLSL
 		}
+
+		Pass
```
huwb, line 125: A comment above each pass might be useful to explain each one

huwb (commented):
The sh code is interesting to see. Is it needed here? I'm aware it's used to light the underwater fog, but had thought it was evaluated at the camera position, and am wondering if we already have it evaluated somewhere else (in oceanrenderer maybe)

huwb:
I wasn't able to add a general comment without submitting the review. Nice one GitHub.

huwb:
Yeah the blending is worth talking through. I had the following in my head but I don't know if it works

* User draws alpha geo and it just draws on top of the underwater i.e. it does not have depth fog applied and looks broken
* We have a component to tag this geo, like you've added, and that renders the geo again but with the depth fog shader which is what I think you have. It alpha blends, with the alpha equal to whatever the depth fog gives (exp(-density*waterRayLength)).

Would that work?

crest/Assets/Crest/Crest/Scripts/Underwater/RegisterUnderwaterInput.cs
```diff
@@ -0,0 +1,43 @@
...
+
+// This file is subject to the MIT License as seen in the root of this folder structure (LICENSE)
+
+namespace Crest
+{
+    using System.Collections.Generic;
+    using UnityEngine;
+
+    // TODO: Maybe rename.
+    public class RegisterUnderwaterInput : MonoBehaviour
```
huwb, line 11: Yeah I think it may want to be distinct from the register input naming as it's quite different. Although that might just be  my bias from knowing the internals. My brain initially went to something like "ApplyUnderwaterEffectToAlpha"

daleeidd:
> The sh code is interesting to see. Is it needed here? I'm aware it's used to light the underwater fog, but had thought it was evaluated at the camera position, and am wondering if we already have it evaluated somewhere else (in oceanrenderer maybe)

I am using it for rendering the transparent object the first time round. For some reason it isn't bound if I render it myself even though I am using _DrawRenderer_. I also have to enable the SH keyword manually. I will look into what I can do to remove it. I should pass the one we already have from the underwater effect for the fog pass to keep it consistent at the very least.

> It alpha blends, with the alpha equal to whatever the depth fog gives (exp(-density*waterRayLength)).

That was my first attempt but density has three components which doesn't look right when converted to a scalar value (tried averaging it or picking only one).

The is what the underwater effect has:
```shader
lerp(sceneColour, scatterCol, saturate(1.0 - exp(-_DepthFogDensity.xyz * sceneZ)));
```
And this is what I tried:
```shader
return (scatterCol, saturate(1.0 - exp(-Average(_DepthFogDensity.xyz) * sceneZ)));
```

Here is a comparison. Current:


Using depth fog as alpha value:


It might be tolerable?

huwb:
Ah interesting! It may be worth trying min or max, but otherwise yeah tolerable I would vote. I don't think we should do anything special to render the original alpha geo. I have expanded my understanding notes above accordingly.

crest/Assets/Crest/Crest/Scripts/Underwater/RegisterUnderwaterInput.cs
daleeidd: I renamed to _ApplyUnderwaterEffectToTransparent_ as that might be more recognisable.

crest/Assets/Crest/Crest/Shaders/Underwater/Resources/UnderwaterEffect.shader
daleeidd: Done

daleeidd:
> Ah interesting! It may be worth trying min or max

BlendOp Min or Max? Neither produced good results.

> but otherwise yeah tolerable I would vote

I have added both for now for evaluating. Or maybe permanently if you think technical debt is not too high.

> User has alpha geo in the scene and unity just draws on top of the underwater i.e. it does not have depth fog applied and looks broken. This uses whatever transparent shader is applied. Unity applies SH and everything else needed to render as normal.

I have thought of another approach which might be closer to this. We render the fog before the transparent pass. And add the fog calculation to the ocean shader again. Then we just duplicate the "tagged" objects and then dilate them slightly. The duplicate will have our material which adds the fog and alpha blends. It should sort correctly and be similar in quality to the lower quality alpha blending above. And requires minimal intervention from us.

huwb:
Oh I meant use min or max density (of the three channels). Not confident it works help but just an idea.

Would you be able to let me know where my notes are falling short or not going to work? I'm not following I'm afraid and am hoping we can use it as a tool to sync on the plan

PR #868, opened 2021-07-10
Revert texture array blit

daleeidd:
Reverts part of  #857 to fix builds for some consoles.


huwb (approved):
Sad face. Approved! 

Could you only add a comment to the code saying why it's still there, and maybe a link to the commit that removed it in case we want to attempt it again in the future? It was non obvious (to me at least) that the code would reduce down that much so may be worth keeping a reference. And mentioning the broken platforms may ensure we don't forget and do the same refactor again in 6 months :). Approving anyway to keep things moving along.

daleeidd:
Thanks. All done.

PR #870, opened 2021-07-13
Convert row boat jpg to png to clear package validation errors

daleeidd:
Also:
- Refactor row boat prefab to be variant of model to make more
robust
- Zero out prefab transform. This is another warning Unity gives and generally good practice
- Use normal texture type for normal map (might have been this before though)

Everything looks like it works as it should.


huwb:
Looks good. I vaguely recall some trickiness with LFS. maybe adding new images requires special treatment downstream in SRP. cant really remember though. thumbs up anyway.

huwb (approved):

daleeidd:
Thanks. Will look out for it.

PR #873, opened 2021-07-18
Add texture format setting to fix precision issues from height input

### LINKED ISSUE #864: Tearing on lake with spline waves above sea level
**Describe the bug**
User eckhart3d reported tearing on the ocean surface. Their setup is a lake which is ~40m above sea level (~251m). It only occurs when a ShapeGerstner is used with a spline so the waves themselves cause tearing.

**Screenshots / video**

**Versions**
- Unity 2021.1.4f1
- HDRP 11
- Crest 4.11

**To Reproduce**
TODO: Need to reproduce it myself first.

**Additional context**
Setting the lake to sea level solved the problem for them. I think also setting the sea level to zero also solves the problem. But investigating the issue may be worth it. Not sure if this HDRP specific.

daleeidd:
Fixes #864

- Refactors settings code to push code up to LodDataMgr
- Can now add settings from SimSettings to settings hash (only needed for changing textures so far I believe)
- Adds texture format option for animated waves to solve precision issues
- Adds validation to height input to warn users if not using higher precision texture
  - only really useful if they are using the default
- Documents animate waves settings and links to it

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgr.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgr.cs
@@ -59,6 +59,25 @@ namespace Crest
 
         protected OceanRenderer _ocean;
 
+        // Implement in any sub-class which supports having an asset file for settings. This is used for polymorphic
+        // operations. A sub-class will also implement an alternative for the specialised type called Settings.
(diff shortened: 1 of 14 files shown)


huwb (approved):
All looks perfect to me, nicely done!

daleeidd:
Thank you.

PR #875, opened 2021-07-18
Fix FFT Game Core shader compile errors

### LINKED ISSUE #856: [GameCore] Shader error in FFTSpectrum
We got this shader error when building for GameCore platforms (XB1 & XSS / XSX). 
image

The building process is not blocked and the ocean looks ok in the builds. We do not yet make use of the new FFT features just for your info. 

Unity 2020.3.9f1, HDRP 10.5, Crest 4.11

daleeidd:
Fixes #856. My guess was that file-level variables cannot be manipulated on Game Core platforms. I haven't confirmed this solves the problem though. FFTs look like they are working as before.


huwb (approved):
Great stuff, much better pattern

daleeidd:
Thanks.

OkoGoran:
Just tested it on game core and it works, perfect thanks!

Issue #865, opened 2021-06-28
Ocean culling incorrect with Register Height Input


daleeidd:
**Describe the bug**
Ocean culling becomes more incorrect the further the height is changed by _register height input_.

**Screenshots / video**
<img width="1110" alt="123131" src="https://user-images.githubusercontent.com/5249806/123713312-f8d60880-d828-11eb-9771-a0609aea1676.png">
This is a 100x100 quad with "set base water height using geometry" just under 99m above sea level.

**Versions**
All RPs and latest master.

**To Reproduce**
1. Create a quad 100x100
2. Add _Register Height Input_ with appropriate material (set or base both exhibit the issue)
3. Set quad height to 99m above sea level
4. Incorrect culling should occur when close enough

**Platform**
 - Editor
 - Mac and Windows

Haven't tested standalone yet.

**Additional context**
The issue is probably incorrect displacement reporting. Also, any _register height input_ further than 99m above sea level appears to be ignored.

daleeidd:
Turns out it wasn't reporting the height because I didn't set either reporting options. @huwb Should report renderer bounds be on by default?

Also the 100m issue because of:
```
_worldToCameraMatrix[lodIdx] = CalculateWorldToCameraMatrixRHS(_renderData[lodIdx]._posSnapped + Vector3.up * 100f, Quaternion.AngleAxis(90f, Vector3.right));
```

So basically anything over 100 above sea level is getting clipped. I don't know if this is something you plan on solving elsewhere like with water bodies?

daleeidd:
Fixed in 4.13.

PR #876, opened 2021-07-18
Report renderer bounds by default for Animated Waves input

daleeidd:
Helps solve #865

Reports the renderer bounds on Animated Waves input by default. The reasoning is that I think reporting the renderer bounds requires some understanding of the system which most will not know. They may come across this issue and not understand the cause. This is a "working correctly by default and optimisation yourself" approach. Let me know what you think.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/LodData/RegisterHeightInput.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/RegisterHeightInput.cs
@@ -42,10 +42,6 @@ namespace Crest
         [SerializeField, Tooltip("Inform ocean how much this input will displace the ocean surface vertically. This is used to set bounding box heights for the ocean tiles.")]
         float _maxDisplacementVertical = 0f;
 
-        [SerializeField, Tooltip("Use the bounding box of an attached renderer component to determine the max vertical displacement.")]
-        [Predicated(typeof(MeshRenderer)), DecoratedField]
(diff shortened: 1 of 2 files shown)


huwb (changes requested):
Yeah works for me, but if we want it to work by default then we need to set the numbers reported to something non zero.

We could set _maxDisplacementVertical to 2m and maybe the horizontal one to 1m?

daleeidd:
Good point. I thought it worked just by toggle that property but it looks like it only really works for the vertical:

```csharp
// let ocean system know how far from the sea level this shape may displace the surface
if (_reportRendererBoundsToOceanSystem && _renderer != null)
{
    var minY = _renderer.bounds.min.y;
    var maxY = _renderer.bounds.max.y;
    var seaLevel = OceanRenderer.Instance.SeaLevel;
    maxDispVert = Mathf.Max(maxDispVert, Mathf.Abs(seaLevel - minY), Mathf.Abs(seaLevel - maxY));
}
```

Should we only do this for the height input? Then the toggle will be enough. Although we could just remove both settings since it looks solid enough to just do regardless for the height input.

huwb (approved):

huwb:
It's worth noting that this will push up the sea level to where the height input is, but any waves beyond that won't be accounted for. But the fix for that would be a strong notion of sea level which is the last pr for local water bodies. This is still worth merging anyway.

daleeidd:
Thank you

PR #877, opened 2021-07-19
Add LOD data matrix height options

daleeidd:
Helps solve #865

Current the LOD data is rendered from a position of 100m above sea level. I have had reports of people setting the height above this position. So this adds a setting to allow them to change both above and below rendering bounds.

Not sure if you planned on solving this some other way with water bodies, but this seems like an okay solution for now. Let me know what you think.

Will add documentation if solution seems good.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/LodData/LodTransform.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/LodTransform.cs
@@ -12,6 +12,11 @@ namespace Crest
     /// </summary>
     public class LodTransform : IFloatingOrigin
     {
+        // Anything higher (minus 1 for near plane) will be clipped.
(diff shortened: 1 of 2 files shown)


huwb (commented):
Thanks. I don't have something else in mind.

I wonder now though if we should just make them very large, rather than adding more settings to ocean renderer. Some recent investigations has shown me that even ridiculously large near/far plane settings still work. And I can't think of any time we rely on depth test in ocean inputs, so precision is unlikely to be an issue.

If that sounds sensible then I think I'd vote for making the values constants and seeing them to -10000 and 10000 respectively, or something similar.

daleeidd:
Even better. Thanks! Let me know if it looks good.

huwb (approved):
Yeah that looks great

daleeidd:
Thanks.

PR #879, opened 2021-07-23
Add signed distance clip surface shader

daleeidd:
Adds signed distance shapes (Sphere and Box) to the clip surface input. It uses the clip surface input transform to define the shape. It renders the quad using DrawMesh. It gets the material from the mesh renderer.

The only issue is the corners of the box can get cut off or rounded:

The PR has a few commits which adds test data which I can remove before code review.

I considered moving away from using a mesh renderer and using DrawMesh for all inputs. But I didn't want it to hold this up as that is a significant refactor when including migration code.

I was thinking of adding the shape type to the component itself and driving the shape keywords with scripting. Thoughts?

## Issues

- Blending doesn't work on my old Mac but probably because it is old

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgrClipSurface.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgrClipSurface.cs
@@ -8,6 +8,8 @@ using UnityEngine.Rendering;
 
 namespace Crest
 {
+    using SettingsType = SimSettingsClipSurface;
+
     /// <summary>
     /// Drives ocean surface clipping (carving holes). 0-1 values, surface clipped when > 0.5.
     /// </summary>
@@ -15,8 +17,7 @@ namespace Crest
     {
         public override string SimName { get { return "ClipSurface"; } }
 
-        // The clip values only really need 8bits
-        protected override GraphicsFormat RequestedTextureFormat => GraphicsFormat.R8_UNorm;
+        protected override GraphicsFormat RequestedTextureFormat => Settings._renderTextureGraphicsFormat;
         protected override bool NeedToReadWriteTextureData { get { return true; } }
         static Texture2DArray s_nullTexture => TextureArrayHelpers.BlackTextureArray;
         protected override Texture2DArray NullTexture => s_nullTexture;
@@ -29,6 +30,9 @@ namespace Crest
         internal const string ERROR_MATERIAL_KEYWORD_ON_FEATURE_OFF_FIX = "If this is not intentional, either enable the <i>Create Clip Surface Data</i> option on this component to turn it on, or disable the <i>Clipping</i> feature on the ocean material to save performance.";
 
         bool _targetsClear = false;
+
+        public override SimSettingsBase SettingsBase => Settings;
+        public SettingsType Settings => _ocean._simSettingsClipSurface != null ? _ocean._simSettingsClipSurface : GetDefaultSettings<SettingsType>();
 
         public LodDataMgrClipSurface(OceanRenderer ocean) : base(ocean)
         {
--- a/crest/Assets/Crest/Crest/Scripts/LodData/RegisterClipSurfaceInput.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/RegisterClipSurfaceInput.cs
@@ -1,8 +1,10 @@
-﻿// Crest Ocean System
+// Crest Ocean System
 
 // This file is subject to the MIT License as seen in the root of this folder structure (LICENSE)
 
+using UnityEditor;
 using UnityEngine;
+using UnityEngine.Rendering;
 
 namespace Crest
 {
@@ -12,7 +14,7 @@ namespace Crest
     /// </summary>
     [AddComponentMenu(MENU_PREFIX + "Clip Surface Input")]
     [HelpURL(Internal.Constants.HELP_URL_BASE_USER + "ocean-simulation.html" + Internal.Constants.HELP_URL_RP + "#clip-surface")]
-    public class RegisterClipSurfaceInput : RegisterLodDataInput<LodDataMgrClipSurface>
+    public partial class RegisterClipSurfaceInput : RegisterLodDataInput<LodDataMgrClipSurface>
     {
         /// <summary>
         /// The version of this asset. Can be used to migrate across versions. This value should
@@ -20,13 +22,38 @@ namespace Crest
         /// </summary>
         [SerializeField, HideInInspector]
 #pragma warning disable 414
-        int _version = 0;
+        int _version = 1;
 #pragma warning restore 414
+
+        const string k_SignedDistanceShaderPath = "Hidden/Crest/Inputs/Clip Surface/Signed Distance";
+
+        public enum Mode
+        {
+            Geometry,
+            Primitive,
+        }
+
+        // Have this match UnityEngine.PrimitiveType.
+        public enum Primitive
+        {
+            Sphere = 0,
+            Cube = 3,
+        }
 
         bool _enabled = true;
         public override bool Enabled => _enabled;
 
-        [Header("Convex Hull Options")]
+        [Header("Clip Surface Input Options")]
+
+        [Tooltip("Where the source of the clipping will come from.")]
+        [SerializeField]
+        Mode _mode = Mode.Primitive;
+
+        [Tooltip("The primitive to render (signed distance) into the simulation.")]
+        [SerializeField, Predicated("_mode", inverted: true, Mode.Primitive), DecoratedField]
+        Primitive _primitive = Primitive.Cube;
+
+        [Header("3D Clipping Options")]
 
         [Tooltip("Prevents inputs from cancelling each other out when aligned vertically. It is imperfect so custom logic might be needed for your use case.")]
         [SerializeField] bool _disableClipSurfaceWhenTooFarFromSurface = false;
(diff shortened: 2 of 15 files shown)


huwb (changes requested):
Looks nice and clean! Great work! Happy to move away from DrawRenderer too. Changing for all inputs would be good as a follow up pr, agreed.

I was wondering if we should move away from having the material/shader dictate the mode, and rather expose it directly on the clip component. I think the material setup is confusing for some users.

I added a register height input component and when used with a spline it just works and it's much nicer. Not needing the user to create a material and fiddle with shaders is a big win I think.

What do you think about either:

* Adding a RegisterClipSurfaceInputPrimitive which has a drop down for which primitive shape, and does not require any mesh renderer component, it automatically picks the right shader and just works
* Add a Type drop down box which defaults to primitive and works without any mesh renderer, but can be set to Custom Geometry which then puts it in the old mode.

I like both, maybe leaning towards option 2, although changing the default requires migration to not break existing components. I'd be tempted to assume _version has made it out to the wild and rely on it, so increment it and if version is 0 default to the old shape mode so that old components continue working.

Both options would require moving the new shader to a resources folder if they're not linked statically to a material asset.

Anyway let me know what you think.

huwb:
Oh and cutting corners is expected and is the limit of what the SDF can represent. Dual Contouring is the approach to improve this I guess, but I have not implemented it before and it would take more time and runtime memory. Just mentioning in case it's of interest. I think what you have here is fine as is.

crest/Assets/Crest/Crest/Shaders/Ocean.shader
```diff
@@ -483,6 +483,7 @@ Shader "Crest/Ocean"
 				clipVal = lerp(_CrestClipByDefault, clipVal, wt_smallerLod + wt_biggerLod);
 				// Add 0.5 bias for LOD blending and texel resolution correction. This will help to tighten and smooth clipped edges
 				clip(-clipVal + 0.5);
+				return float4(clipVal, 0, 0, 1);
```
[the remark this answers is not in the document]
daleeidd: Yes. I will remove it.

crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgrClipSurface.cs
```diff
@@ -8,15 +8,16 @@
...
     /// <summary>
     /// Drives ocean surface clipping (carving holes). 0-1 values, surface clipped when > 0.5.
     /// </summary>
     public class LodDataMgrClipSurface : LodDataMgr
     {
         public override string SimName { get { return "ClipSurface"; } }
 
-        // The clip values only really need 8bits
-        protected override GraphicsFormat RequestedTextureFormat => GraphicsFormat.R8_UNorm;
+        protected override GraphicsFormat RequestedTextureFormat => Settings._renderTextureGraphicsFormat;
```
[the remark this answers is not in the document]
daleeidd: Only for it to be sharp. But even with R8, it is still a huge improvement. Goran will probably want R16. It defaults to R8.

1 Convex
Convex Hull

2 SD R8
Signed Distance R8

3 SD R16
Signed Distance R16

daleeidd:
> Looks nice and clean! Great work! Happy to move away from DrawRenderer too. Changing for all inputs would be good as a follow up pr, agreed.

Thank you. Sounds good.

> I was wondering if we should move away from having the material/shader dictate the mode, and rather expose it directly on the clip component. I think the material setup is confusing for some users.
> 
> I added a register height input component and when used with a spline it just works and it's much nicer. Not needing the user to create a material and fiddle with shaders is a big win I think.
> 
> What do you think about either:
> 
> * Adding a RegisterClipSurfaceInputPrimitive which has a drop down for which primitive shape, and does not require any mesh renderer component, it automatically picks the right shader and just works
> * Add a Type drop down box which defaults to primitive and works without any mesh renderer, but can be set to Custom Geometry which then puts it in the old mode.
> 
> I like both, maybe leaning towards option 2, although changing the default requires migration to not break existing components. I'd be tempted to assume _version has made it out to the wild and rely on it, so increment it and if version is 0 default to the old shape mode so that old components continue working.

Perfect. Option 2 sounds great. We could also check for a mesh renderer to help with setting the default.

daleeidd:
@huwb It is now driven by the component and the material is created at runtime. I had to refactor input validation (looking forward to using DrawMesh etc instead). I have rebased down to three commits to make understanding changes easier.

I have updated the original post a known issue. But I think we can ignore it.

crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgrClipSurface.cs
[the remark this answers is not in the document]
huwb: Ah super useful. Thanks. What you have sounds good to me.

huwb:
Thanks I think it's looking good.

I still think we should probably make use of _version. I haven't quite understood the migration here, and it's generally pretty taxing to think about all the cases, and for this reason I think having a standard pattern using _version will help a lot.

The proposed approach so far is to bump version default value to 1 in the code, so any fresh components will have version 1. If _version is set to 0 we know it's old data and should use geometry, like the component used to do. Finally the editor detects version 0 and migrates it forwarded accordingly. Hoping this covers every case. I could prob document this somewhere if it works out so we can deploy it easily going forwards.

huwb:
Oh re the issue, so max blend doesn't work on your Mac? That's somewhat surprising as it should I would have thought, however I'm fine with merging despite this.

daleeidd:
> I still think we should probably make use of _version. I haven't quite understood the migration here, and it's generally pretty taxing to think about all the cases, and for this reason I think having a standard pattern using _version will help a lot.
> 
> The proposed approach so far is to bump version default value to 1 in the code, so any fresh components will have version 1. If _version is set to 0 we know it's old data and should use geometry, like the component used to do. Finally the editor detects version 0 and migrates it forwarded accordingly. Hoping this covers every case. I could prob document this somewhere if it works out so we can deploy it easily going forwards.

Latest commit adds this. Let me know if it is done correctly. It was annoying that SetDirty didn't dirty the scene when called from Awake or OnEnable so I had to call it separately in Start.

> Oh re the issue, so max blend doesn't work on your Mac?

It worked with DrawRenderer, but stopped with DrawMesh. I confirmed that Windows is fine. It might be fixed in newer Macs so I will wait for someone to raise a bug with us first.

huwb:
Thanks. The following may be wrong but this was my hope..

Upgrade should not need to run, and we don't need to add anything else beyond my notes, so bit call to upgrade from start. The code should run happily unupgraded. If the runtime logic finds version is set to 0, it doesn't trust the new bit of data (mode) and assumed geometry mode. So the only thing calling upgrade is the editor.

Hopefully that means we don't need the additional complexity, but let me know. I think it worked for gerstner in a test version, but I did not test exhaustively as I realised version being present could not be relied on at that time.

This is a documentation fail on my side, if this works out I'll get it documented.

daleeidd:
> Upgrade should not absolutely need to run, and we don't need to add anything else beyond my notes, so not need to call upgrade from start. The code should run happily unupgraded. If the runtime logic finds version is set to 0, it doesn't trust the new bit of data (mode) and assumes geometry mode. So the only thing calling upgrade is the editor.

So whenever I check mode I would also be checking version too? I think I will have to make a c# property as mode is used in several places.

huwb:
Yeah that's correct, and that would make sense.

This pattern is a product of trying to avoid unity pitfalls / lack of clean model. If data and runtime state was separate this would be much simpler. For example if data was deserialised by us explicitly then we could upgrade it on load and have up to date data in the runtime at all times. Sadly in unity it's all a big mish mash, serialised data and runtime state are one and the same, authoring and running are mixed together, etc. It's possible to force it to work as you have done, but the result is hard to understand and reason about, so I'd like to try with the property, which isolates the change to one place.

Maybe put a comment on the property with the date. In 3 or 6 months we can likely remove the property over we are confident enough data has been brought forward.

daleeidd:
> Sadly in unity it's all a big mish mash, serialised data and runtime state are one and the same, authoring and running are mixed together, etc. It's possible to force it to work as you have done, but the result is hard to understand and reason about, so I'd like to try with the property, which isolates the change to one place.

I've added a commit which uses _ISerializationCallbackReceiver.OnAfterDeserialize_. Seems to work pretty well. It works in editor and build. I am not sure about all of the drawbacks. I think it can be called multiple times, but looks like it is safe.

crest/Assets/Crest/Crest/Scripts/LodData/RegisterClipSurfaceInput.cs
```diff
@@ -98,9 +235,80 @@ private void LateUpdate()
...
+            }
+        }
 #endif
     }
+
+    public partial class RegisterClipSurfaceInput : ISerializationCallbackReceiver
+    {
+        public void OnAfterDeserialize()
+        {
+            if (_version == 0)
```
huwb, line 301: this looks great!

contemplating if this should dirty the component. i think dirtying would push the version forward. on the flipside, in production scenarios, an artist finding the scene is dirtied for unknown reasons causes confusion. i think maybe lets leave it the way you have it.

perhaps its worth adding a comment above this if() statement with the date and reason (default mode changed from geo to primitive). seems useful to have this logged in an easily readable way, and makes decisions about removing migration code in the future easy.

huwb (approved):
if this is working then this is a great find. feels like it should work. its not intrusive at all and its clear so big thumbs up from me

crest/Assets/Crest/Crest/Scripts/LodData/RegisterClipSurfaceInput.cs
```diff
@@ -98,9 +235,80 @@ private void LateUpdate()
...
+                    break;
+                default:
+                    Debug.LogError("Crest: Not a valid primitive type!");
+                    break;
+            }
+        }
 #endif
     }
+
+    public partial class RegisterClipSurfaceInput : ISerializationCallbackReceiver
```
huwb, line 297: oh it was probably already your plan but may be worth adding a comment above this just saying 

`// Version handling - perform data migration after data loaded` 

or whatever makes sense here

crest/Assets/Crest/Crest/Scripts/LodData/RegisterClipSurfaceInput.cs
daleeidd: > artist finding the scene is dirtied for unknown reasons causes confusion

I dislike this too. Also, calling dirty cannot be done in these methods so I would need to call it in Start.

> perhaps its worth adding a comment

Makes sense. Done.

crest/Assets/Crest/Crest/Scripts/LodData/RegisterClipSurfaceInput.cs
daleeidd: Done.

daleeidd:
@huwb All done. I have also updated the examples. The floating container uses the primitives now. And I have updated the documentation.

PR #880, opened 2021-07-24
Fix disabled input not being cleared if last one

daleeidd:
Flow, ~~clip and depth~~ simulations where not being cleared when the last input is disabled.

### DIFF
--- a/docs/about/history.rst
+++ b/docs/about/history.rst
@@ -27,6 +27,7 @@ Fixed
    -  Fix ocean not rendering on Xbox One and Xbox Series X.
    -  Fix height input (and others) from not working 100m above sea level and 500m below sea level.
    -  Fix FFT shader build errors for Game Core platforms.
+   -  Fix flow simulation sometimes not clearing after disabling last input.
 
    .. only:: hdrp
 
--- a/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgrFlow.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgrFlow.cs
@@ -86,11 +86,8 @@ namespace Crest
                 SubmitDraws(lodIdx, buf);
             }
 
-            // targets have now been cleared, we can early out next time around
-            if (drawList.Count == 0)
-            {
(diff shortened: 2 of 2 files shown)


crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgrClipSurface.cs
```diff
@@ -60,6 +60,8 @@ public override void BuildCommandBuffer(OceanRenderer ocean, CommandBuffer buf)
                 return;
             }
 
+            _targetsClear = false;
```
huwb, line 63: Interesting.. targetsClear is anyway written below the loop, so setting it false here only impacts the loop.. I think.. Is the value used within the loop, hence setting it here fixes it? Just trying to get my head around what is going wrong.

huwb (approved):
Approving as it looks safe enough, but a comment on why it's needed might be useful so we (I) don't deem the extra assignment superfluous and remove it in the future.

crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgrClipSurface.cs
daleeidd: Apologies. I made a mistake. Only flow was the problem. I have rebased with the correct fix for flow.

docs/about/history.rst
```diff
@@ -27,6 +27,7 @@ Fixed
    -  Fix ocean not rendering on Xbox One and Xbox Series X.
    -  Fix height input (and others) from not working 100m above sea level and 500m below sea level.
    -  Fix FFT shader build errors for Game Core platforms.
+   -  Fix flow simulation sometimes not clearing disable input.
```
huwb, line 30: Oh minor gramma .. "..not clearing after disabling last.."

huwb (approved):
Oh yep that's a proper logic fail! Thanks!

huwb:
Logic fail on my side, I meant haha

daleeidd:
All good. Sweet. Thanks for the review.

PR #881, opened 2021-07-25
Baked fft collision

huwb:
**Status**: Ready to commit.

**Overview**

Adds options to the bottom of the ShapeFFT inspector that allow the FFT to be baked to a serialised object.

The anim wave settings has a new coll provider for baked FFT data, it can be pointed at this baked data. This coll provider samples the baked data to service queries.

This allows complex/realistic wave shape to be used on gpu-less / nographics servers.

The data is baked as multi-res slices, so full range of wavelengths supported and data is very compact.

It makes use of unity.mathematics and (optionally) burst. The burst jobs run across frames.

**Performance**

The wall clock time spent in burst jobs for 4096 height queries is ~0.9ms. This time is on spare cores and does not block the main thread.

Added perf notes to header in FFTBakedData.cs.

The memory is currently duplicated - `_framesFlattened` is converted to native array. `_framesFlattened` is set null at runtime and is likely GC'd.

**Limitations**

* It does not support depth caches, that is left for future work.
* It only supports a single FFT. It would be possible to add support for multiple, but this should cover most use cases.
* It does not support the wave spline tech. That would need investigating.

**TODO**

* Figure out what to do about package requirements. Mathematics is required by this code. Burst is optionally but very highly recommended.
* Documentation


daleeidd:
> It makes use of unity.mathematics and (optionally) burst, i have forgotten what the logistics are to allow this Dale, if you could remind me where we landed with this that would be useful, thanks.

There are a few options:

1. We [add the package using scripting API](https://docs.unity3d.com/Manual/upm-api.html) which will work for GH and UAS
2. We can upload dependencies (manifest.json) to UAS using the _Asset Store Tools_. This is awkward though because it takes the entire manifest which will include dependencies users won't want :(
3. Let users handle it themselves (log error if packages not found)

Regardless of which option, we will need to surround burst/math code with assembly version defines.

kpietraszko:
@daleeidd I think our idea was to have Mathematics as hard dependency (float4 is pretty important here), and for Burst dependency, I think just the `[BurstCompile]` could be surrounded with an assembly version define? Correct me if I'm wrong.

daleeidd:
> Regardless of which option, we will need to surround burst/math code with assembly version defines.

I'll rephrase it as "depending on which option".

It might be possible to avoid defines for any of those options. Option 1 requires code to be successfully compiled for it to work. Having that code in its own assembly should solve that problem.

kpietraszko:
side note: baked data is allocated twice in RAM - `float[]` and a `NativeArray` that's required for the Burst jobs. For now we decided that's fine, because the data is usually <20MB. There are options around it obviously, if needed.
another note: if you're wondering why the preview is on `SimSettingsAnimatedWaves`, instead of `FFTBakedData` itself; it's because depending on the settings you use for baking, the ScriptableObject can become large enough for Unity to start choking when it's selected. The inspector keeps deserializing the file all the time, which in some cases drops the Editor fps to <5. That would make the interactive preview not-so interactive.
last note: the performance is not perfect after recent change of approach, which could become even more apparent if shallow attenuation is added.

crest/Assets/Crest/Crest/Scripts/Collision/CollProviderBakedFFT.cs
```diff
@@ -0,0 +1,421 @@
+﻿// Crest Ocean System
+
+// This file is subject to the MIT License as seen in the root of this folder structure (LICENSE)
+
```
daleeidd, line 4: Adding the following to the top of the file (and for any other file using math) will soften the blow of not having the package installed:

```
#if !CREST_MATH
#error Crest requires the Unity Mathematics package (com.unity.mathematics) to be installed.
#endif
```

It will require a package version defines in the assembly definition.

daleeidd:
> I think just the [BurstCompile] could be surrounded with an assembly version define?

If burst is optional and that works that sounds great.

daleeidd:
And just want to say good job both getting this all together. Massive PR :^)

daleeidd:
I have edited the package options. If we don't want users to install packages manually, then it looks like we need to use package version defines to surround the new classes that use mathematics. One issue is that one of these classes looks to be  serialised which may be problematic if conditionally compiled out.

daleeidd:
And this PR replaces #863?

I had trouble getting it to work on my Mac (not surprised). Will try Windows tomorrow.

huwb:
> And this PR replaces #863?

Oh yep, closed that one thanks

@daleeidd regarding mathematics, so we could add CREST_MATH and wrap perhaps the body of the functions in FFTBakedData. Regarding serialisation the main issue would be this i guess:

`half[] _framesFlattened;`

`half` is a wrapper around a `ushort`, i wonder if the serialisation will hold up if we define this as a ushort array when mathematics is not present.

huwb:
@kpietraszko thanks for notes. regarding the memory duplication, what was the potential fix again, was it to set `_framesFlattened` to `null` when in play mode (after conversion to NA)? As long as it doesnt stay null when it comes back out that seems reasonable.

kpietraszko:
@huwb yeah that should be enough, but possibly you might also need to forcefully trigger GC. Hopefully that doesn't freeze, Mono seems funky in that regard

kpietraszko:
> `half` is a wrapper around a `ushort`, i wonder if the serialisation will hold up if we define this as a ushort array when mathematics is not present.

@huwb something like this should be pretty solid if you use NativeArray<ushort>:
`_framesFlattenedNative = new NativeArray<ushort>(_framesFlattened, Allocator.Persistent).Reinterpret<half>()`

daleeidd:
I could not get this to work. I get the following exception:
```
IndexOutOfRangeException: Index was outside the bounds of the array.
Crest.CollProviderBakedFFT.Query (System.Int32 i_ownerHash, System.Single i_minSpatialLength, UnityEngine.Vector3[] i_queryPoints, UnityEngine.Vector3[] o_resultDisps, UnityEngine.Vector3[] o_resultNorms, UnityEngine.Vector3[] o_resultVels) (at Assets/Crest/Crest/Scripts/Collision/CollProviderBakedFFT.cs:51)
Crest.SampleHeightHelper.Sample (System.Single& o_height) (at Assets/Crest/Crest/Scripts/Collision/SamplingHelpers.cs:60)
Crest.OceanRenderer.LateUpdateViewerHeight () (at Assets/Crest/Crest/Scripts/OceanRenderer.cs:1034)
Crest.OceanRenderer.RunUpdate () (at Assets/Crest/Crest/Scripts/OceanRenderer.cs:894)
Crest.OceanRenderer.EditorUpdate () (at Assets/Crest/Crest/Scripts/OceanRenderer.cs:570)
UnityEditor.EditorApplication.Internal_CallUpdateFunctions () (at <c77ced6644e541b7ad3d7315a5e92a0a>:0)
```
I disabled the gerstner in the main scene and added FFT with same spectrum. I clicked the "bake to asset..." button.

kpietraszko:
@daleeidd The one that returns displacement is broken it seems, and it's used more often. The one that returns heights is ok.

huwb:
Yeah thanks, not sure why i didnt see that before. i've been trying to fix it and making progress slowly..

huwb:
I pushed fixes for 2 of the problems now. Think maybe the normal calculation has a bug.

daleeidd:
> @daleeidd regarding mathematics, so we could add CREST_MATH and wrap perhaps the body of the functions in FFTBakedData. Then complain when the functions are called without mathematics present. That would get my vote rather than add an extra technical step to our asset installation (although if we start using mathematics in multiple places then thats what we'll be facing).

Not sure I understand. Are you proposing that we only complain about no mathematics package when someone tries to use the feature?

huwb:
Yeah thats what I'm wondering. What do you think?

Normals fixed now

daleeidd:
> Yeah thats what I'm wondering. What do you think?

Sounds good. Adding a package through the UPM API is pretty simple so we could have a "fix" button on the FFT component if the user enables FFT baking if you want to go that route.
 
> Normals fixed now. It should be good to go. Apologies for those errors..

Thanks. I will have a look.

crest/Assets/Crest/Crest/Shaders/Resources/FFT/FFTBake.compute
```diff
@@ -0,0 +1,31 @@
+// Crest Ocean System
+
+// Copyright 2021 Wave Harmonic Ltd
```
daleeidd, line 3: Should be MIT

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
```diff
@@ -388,6 +401,109 @@ public bool Validate(OceanRenderer ocean, ValidatedHelper.ShowMessage showMessag
...
+        }
+
+        /// <summary>
+        /// Controls & GUI for baking.
+        /// </summary>
+        void OnInspectorGUIBaking()
+        {
+            var bakeAndAssignLabel = "Bake to asset and assign to current settings";
+            var selectCurrentSettingsLabel = "Select current settings";
+            if (OceanRenderer.Instance._simSettingsAnimatedWaves != null)
```
daleeidd, line 471: We also need to check OceanRenderer is present (early exit) otherwise error log spam

daleeidd (changes requested):
First round review. Works really well. Only minor changes needed.

How does "Shape FFT > Enable Baked Collision" and "Animated Waves Settings > Collision Source > Baked FFT" differ or work together to determine if baked data is used?

crest/Assets/Crest/Crest/Scripts/LodData/Settings/SimSettingsAnimatedWaves.cs
```diff
@@ -50,6 +51,9 @@ public enum CollisionSources
         [Tooltip("The render texture format to use for the wave simulation. It should only be changed if you need more precision. See the documentation for information.")]
         public GraphicsFormat _renderTextureGraphicsFormat = GraphicsFormat.R16G16B16A16_SFloat;
 
+        [Predicated("_collisionSource", true, 3), DecoratedField]
+        public FFTBakedData _bakedFFTData;
+
         public override void AddToSettingsHash(ref int settingsHash)
         {
             base.AddToSettingsHash(ref settingsHash);
```
daleeidd, line 59: Might as well add the collision source to the hash now that it is relevant in the editor again with this PR.

kpietraszko:
Having just 5 `ObjectWaterInteractions` in the scene seems to decrease fps in editor from 30 to ~9. In standalone build it's fine. Looks like a large number of NativeArray allocations in `CollProviderBakedFFT` is the culprit, namely DisposeSentinel created for leak tracking. Disabling safety checks suprisingly doesn't help. 
image

skaughtx0r:
This is great and will help to make the simulation on dedicated server match what client sees better.

However, I'm curious, will this also work with flow maps and materials that alter water height? I'm starting to use flow maps to model water currents and I'm worried its going to make the client and server become very out of sync.

crest/Assets/Crest/Crest/Scripts/Collision/CollProviderBakedFFT.cs
```diff
@@ -0,0 +1,469 @@
...
+                {
+                    _queryPointsX = queryPointsX,
+                    _queryPointsZ = queryPointsZ,
+                    _framesFlattened = _data._framesFlattenedNative,
+                    _t = t,
+                    _params = _data._parameters,
+                    _outputX = resultsX,
+                    _outputY = resultsY,
+                    _outputZ = resultsZ,
+                }.Schedule(numQueryQuads, s_jobBatchSize).Complete();
```
kpietraszko, line 81: It would probably be a good performance improvement if instead of `Complete`ing 3 jobs synchronously, they were combined (`CombineDependencies` I think?) into a single JobHandle which would be then `Complete`d and finally assign values to results. There's a lot of waiting done on job threads currently.

huwb:
@skaughtx0r your suspicion is correct - the current stuff does not support flow, nor does it support depth caches which is probably the first extension that will be needed. i'm interested in both of these things but dont have a timeline on them yet.

(a random idea may be to only flow normal maps, not waves. so knock out flow from the wave combine pass. that may visually look like flow without actually flowing the wave displacements. not sure if this is useful or not, just mentioning)

crest/Assets/Crest/Crest/Scripts/Collision/CollProviderBakedFFT.cs
huwb: this is done on latest, thanks

crest/Assets/Crest/Crest/Shaders/Resources/FFT/FFTBake.compute
huwb: thanks, done

crest/Assets/Crest/Crest/Scripts/Collision/VisualiseCollisionArea.cs
```diff
@@ -38,28 +50,70 @@ void Update()
...
+                _resultDisps = new Vector3[_steps * _steps];
+            }
+            if (_resultNorms == null || _resultNorms.Length != _steps * _steps)
+            {
+                _resultNorms = new Vector3[_steps * _steps];
+
+                for (int i = 0; i < _resultNorms.Length; i++)
+                {
+                    _resultNorms[i] = Vector3.up;
+                }
```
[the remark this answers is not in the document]
huwb: fixed, thanks

crest/Assets/Crest/Crest/Scripts/Collision/VisualiseCollisionArea.cs
```diff
@@ -38,28 +50,70 @@ void Update()
...
+                        {
+                            var result = _samplePositions[j * _steps + i];
+                            result.y = _resultHeights[j * _steps + i];
 
-                        DebugDrawCross(result, 1f, Color.green);
+                            DebugDrawCross(result, _resultNorms[j * _steps + i], Mathf.Min(_stepSize / 4f, 1f), Color.green);
+                        }
                     }
                 }
             }
```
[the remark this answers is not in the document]
huwb: i did look at that but they diverge about as much as they have in common. i started sketching it out and it obfuscates the flow a fiar bit to push the branch inwards so i might leave it if thats ok.

crest/Assets/Crest/Crest/Scripts/Collision/CollProviderBakedFFT.cs
```diff
@@ -0,0 +1,469 @@
...
+            return queryStatus == (int)QueryStatus.Success;
+        }
+
+        public void UpdateQueries()
+        {
+        }
+
+        public void CleanUp()
+        {
+        }
```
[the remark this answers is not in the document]
huwb: fixed  - they have been filled!

crest/Assets/Crest/Crest/Scripts/Editor/SimSettingsAnimatedWavesPreview.cs
```diff
@@ -0,0 +1,110 @@
...
+                    rawDataNative[nativeIndex + 0] = Mathf.Clamp01(mappedValueX);
+                    rawDataNative[nativeIndex + 1] = Mathf.Clamp01(mappedValueY);
+                    rawDataNative[nativeIndex + 2] = Mathf.Clamp01(mappedValueZ);
+                    rawDataNative[nativeIndex + 3] = 1f;
+                }
+
+                if (rawDataNative.Length == 0)
+                    return;
+
+                _previewTexture ??= new Texture2D(
```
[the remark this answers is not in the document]
huwb: yeah when i first saw this i thought it wouldnt work, but it seems to. maybe the operator is overridden by unity? i dont recall the details anymore.

crest/Assets/Crest/Crest/Scripts/Editor/SimSettingsAnimatedWavesPreview.cs
```diff
@@ -0,0 +1,110 @@
...
+    /// If the collision source in the animated wave settings is set to baked FFT, shows a preview of the data.
+    /// </summary>
+    [CustomPreview(typeof(SimSettingsAnimatedWaves))]
+    public class SimSettingsAnimatedWavesPreview : ObjectPreview
+    {
+        private Texture2D _previewTexture;
+        private int _frameToPreview = 0;
+        private Object _previousTarget;
+        private SimSettingsAnimatedWaves _targetAnimatedWaves => target as SimSettingsAnimatedWaves;
+        private FFTBakedData _targetBakedData => _targetAnimatedWaves?._bakedFFTData;
```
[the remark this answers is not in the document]
huwb: also seemed to work, but i changed it anyway now

crest/Assets/Crest/Crest/Scripts/LodData/Settings/SimSettingsAnimatedWaves.cs
```diff
@@ -50,6 +51,9 @@ public enum CollisionSources
         [Tooltip("The render texture format to use for the wave simulation. It should only be changed if you need more precision. See the documentation for information.")]
         public GraphicsFormat _renderTextureGraphicsFormat = GraphicsFormat.R16G16B16A16_SFloat;
 
+        [Predicated("_collisionSource", true, 3), DecoratedField]
```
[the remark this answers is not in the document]
huwb: indeed, much nicer, thanks!

crest/Assets/Crest/Crest/Scripts/LodData/Settings/SimSettingsAnimatedWaves.cs
```diff
@@ -148,6 +171,18 @@ internal static void FixSetCollisionSourceToCompute(SerializedObject settingsObj
...
             }
         }
+
+        internal static void FixOceanWindSpeed(SerializedObject settingsObject)
+        {
+            if (OceanRenderer.Instance != null
+                && OceanRenderer.Instance._simSettingsAnimatedWaves != null
+                && OceanRenderer.Instance._simSettingsAnimatedWaves._bakedFFTData != null)
+            {
+                Undo.RecordObject(OceanRenderer.Instance._simSettingsAnimatedWaves, "Set global wind speed to match baked data");
```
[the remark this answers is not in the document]
huwb: good catch thanks

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
huwb: cheers done

crest/Assets/Crest/Crest/Scripts/LodData/Settings/SimSettingsAnimatedWaves.cs
huwb: good shout, done

huwb:
I changed the system to aggregrate all query points into one list, then run the jobs async across frames. This takes the processing off the main thread. it also combines the job handles rather than running separate jobs.

This was inspired by #157 , but ended up diverging quite a lot, as #157 is not up to date with the new query api, and only supported heights. I'm happy with where this landed. However the logic was really tricky and i've not had anywhere near the time/focus that this stuff deserves, so there may be some turbulence. I ran it through a set of tests and it seems to hold up so far..

@kpietraszko Re freeing the raw data (`_framesFlattened`) - done but im not confident to trigger a GC which could cause a big stall depending on external circumstances, so leaving as it is

@daleeidd i havent had a chance to look at the package side of things, if you want to take a swing at that please do, or ill do it later in the week or next weekend.

huwb:
> How does "Shape FFT > Enable Baked Collision" and "Animated Waves Settings > Collision Source > Baked FFT" differ or work together to determine if baked data is used?

Missed this apologies.

"Shape FFT > Enable Baked Collision" has no bearing on whether baked data is used. It only makes the FFT periodic.

So only the latter determines of baked data is used.

Does this make sense?

huwb:
@kpietraszko i had a case where _framesFlattened was null'd out and kept its null value - lost all my collision data. It may have been a c# hot rebuild during play mode, or something similar.

it may be that nulling out that ref in standalone builds is fine if you want to run with this and test it out lmk.

skaughtx0r:
@huwb Making the flow maps only move the normals wouldn't help as I'm using the flow from crest as part of the physics simulation for water current.

crest/Assets/Crest/Crest/Scripts/Editor/SimSettingsAnimatedWavesPreview.cs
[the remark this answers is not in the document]
daleeidd: Thanks. They can work but behaviour is undefined as Unity cannot overload them like the equality operator.

crest/Assets/Crest/Crest/Scripts/Collision/VisualiseCollisionArea.cs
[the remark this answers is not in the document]
daleeidd: True. I had a look again and I missed something which doesn't make it work as well as I thought.

daleeidd:
> @daleeidd i havent had a chance to look at the package side of things, if you want to take a swing at that please do, or ill do it later in the week or next weekend.

Sure. I'll see what I can do.

> "Shape FFT > Enable Baked Collision" has no bearing on whether baked data is used. It only makes the FFT periodic.
> 
> So only the latter determines of baked data is used.
> 
> Does this make sense?

Yes. Thanks.

kpietraszko:
@huwb Yeah I remember having big issues with nulling that out couple weeks back, on the previous iteration. While in theory it should be fine to have this as `#!UNITY_EDITOR` block, it broke both editor and standalone build completely and critically. I'd say nulling out is not crucial and should be left out.

huwb:
I've fixed a few bugs. It's properly async now - does not force a Complete() in the query function.

It seems to workreliably for me. I added multiple VCAs and boats and turned them on and off and everything seems happy.

huwb:
Updated from master

@kpietraszko are you sampling from one owner ID (or sample helper object), and querying with teh same id each frame until a result comes through?

'not ready yet' should only be returned in the first or second frame after querying begins for a owner ID, and is expected in this case as it takes time for the query to be processed. just checking that you're not seeing something that is by design. if you have a new query owner every second or so then that return value is expected.

if this is not the case, then i dont know what would be blocking it. a threading issue/race condition seems possible. is the dots code executing on the main thread? if not it may be worth fudging in a mutex into the query functions as a test to see if it resolves it.

kpietraszko:
@huwb The case on that screenshot is a single script (or DOTS system actually), I call it from regular .net land every frame, and the ownerID is consistent. Interestingly, when I disable Burst in the editor I only get "query failed" on ticks 14 and 25, and then it's fine. This could indicate a threading issue/race condition.

huwb:
Yeah i'm not really sure where such a race condition would come from.

This may benefit from wider testing, so may be worth merging in preview state once we figure out the plan with Mathematics/Burst dependencies. Let me know what you think @daleeidd .

daleeidd:
> This may benefit from wider testing, so may be worth merging in preview state once we figure out the plan with Mathematics/Burst dependencies. Let me know what you think @daleeidd .

Sounds good. I have a PR for one approach: https://github.com/kpietraszko/crest/pull/1

huwb:
Totally missed it. Approved it now @daleeidd

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
```diff
@@ -398,12 +411,132 @@ public bool Validate(OceanRenderer ocean, ValidatedHelper.ShowMessage showMessag
...
+        /// </summary>
+        void OnInspectorGUIBaking()
+        {
+            if (OceanRenderer.Instance == null) return;
+
+            var bakeAndAssignLabel = "Bake to asset and assign to current settings";
+            var selectCurrentSettingsLabel = "Select current settings";
+            if (OceanRenderer.Instance._simSettingsAnimatedWaves != null)
+            {
+                if (GUILayout.Button(bakeAndAssignLabel))
```
daleeidd, line 505: Should we disable this if Burst is not present? From testing it has caused crashing if data is used and wasn't baked using Burst.

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
```diff
@@ -95,6 +97,18 @@ public partial class ShapeFFT : MonoBehaviour, LodDataMgrAnimWaves.IShapeUpdatab
...
         float _maxHorizontalDisplacement = 15f;
 
+        [Header("Collision Data Baking")]
+        [Tooltip("Enable running this FFT with baked data. This makes the FFT periodic (repeating in time).")]
+        public bool _enableBakedCollision = false;
+        [Tooltip("Frames per second of baked data. Larger values generate more frames and increase baked data size."), DecoratedField, Predicated("_enableBakedCollision")]
+        public int _timeResolution = 4;
+        [Tooltip("Smallest wavelength required in collision. To preview disable power sliders in spectrum for smaller values than this number. Smaller values require more resolution and increase baked data size."), DecoratedField, Predicated("_enableBakedCollision")]
+        public float _smallestWavelengthRequired = 2f;
+        [Tooltip("FFT waves will loop with a periodic of this many seconds."), DecoratedField, Predicated("_enableBakedCollision"), Range(4f, 128f), Delayed]
```
daleeidd, line 107: Cannot mix DecoratedField, Range and Delayed together as they are fields attributes rather than decorators (like Predicated). You'll need to choose either Ranged or Delayed.

I should probably make Delayed a decorator.

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/FFTBaker.cs
```diff
@@ -0,0 +1,194 @@
...
+        {
+            smallest = largest = -1;
+
+            for (var i = 0; i < OceanWaveSpectrum.NUM_OCTAVES; i++)
+            {
+                var pow = spectrum._powerDisabled[i] ? 0f : Mathf.Pow(10f, spectrum._powerLog[i]);
+                if (pow > Mathf.Pow(10f, OceanWaveSpectrum.MIN_POWER_LOG))
+                {
+                    var minWL = Mathf.Pow(2f, OceanWaveSpectrum.SMALLEST_WL_POW_2 + i);
+                    if (2f * minWL > minIncludedWavelength && smallest == -1 && minWL >= smallest)
```
daleeidd, line 181: This condition is a little confusing. Why are we checking that `smallest == -1`?

Also, should `2f * minWL > minIncludedWavelength` be inclusive (ie >= instead of >)? Or maybe it is already because of `2 *`? A comment will help here.

daleeidd (changes requested):
Only minor things.

I am getting weirdness with the boat the main scene. I have disabled the island/depth cache and using baked FFT data and the boat is either unresponsive, gets stuck or flies away. The visualize queries looks correct though. Are you getting the same issue? Boat is fine with GPU queries.

kpietraszko:
@daleeidd Sounds very similar to the issue I'm having. Up to you guys if you want to merge as-is.

crest/Assets/Crest/Crest/Scripts/OceanValidation.cs
```diff
@@ -108,6 +109,16 @@ static void FixRemoveRenderer(SerializedObject componentOrGameObject)
...
 
+        public static void FixAddMissingMathPackage(SerializedObject componentOrGameObject)
+        {
+            PackageManagerHelpers.AddMissingPackage("com.unity.mathematics");
+        }
+
+        public static void FixAddMissingBurstPackage(SerializedObject componentOrGameObject)
+        {
+            PackageManagerHelpers.AddMissingPackage("com.unity.burst");
+        }
```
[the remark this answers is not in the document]
huwb: um yeah latest is hopefully fine - if we hardcoded the version then we'd have to constantly keep it up to date

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
huwb: Yeah making a Delayed decorator might be useful as this wants both delayed and range.

I removed delayed and noticed ranged is not working (no slider in gui), can you spot what it needs? I think i've bumped into this before but can never remember what to do to get it working. If this issue could be caught with an error or assert or such that would be helpful.

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/FFTBaker.cs
huwb: you're right, and it had a clause that was totally meaningless! clarified now, lmk if it makes sense

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
huwb: I could repro the crash when i first started testing, but after reinstalling latest burst (1.6.1) the crash may have gone away for me. I've tried various combinations of entering/exiting play mode, baking, with burst in/out and it doesnt seem to crash again.

huwb:
Glitches should be fixed. At high frame rates, queries in FixedUpdate would not be made, and the queries would get lost.

The code now keeps queries around for a time so they wont get lost when FixedUpdate doesnt get called. This same fix exists in QueryBase.cs.

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
daleeidd: Excellent.

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
daleeidd: DecoratedField needs to be removed. Both Range and DecoratedField are fields and there can be only one. DecoratedField exists only to make decorators work. Range is already a decorated field so decorators work if that makes sense.

> If this issue could be caught with an error or assert or such that would be helpful.

Good idea. I'll look into adding it.

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/FFTBaker.cs
daleeidd: Makes sense. Thanks.

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
huwb: ah thanks. fixed now. sounds good

huwb:
I moved the preview to FFTBakedData rather than proxying through SSAW, hope thats ok. That is the data it is previewing, and i found it awkward to preview through the settings (and theres a slow down too).

I believe its just docs left now

docs/about/history.rst
```diff
@@ -29,6 +29,7 @@ Changed
    -  Add buttons to *Spline* inspector to quickly enable water features.
    -  Exposed control over *Spline* ribbon alignment - spline points now define the center of the ribbon by default.
    -  Caustics no longer render in shadows casted from objects underwater.
+   -  Added new CPU-based collision provider - *Baked FFT Data*.
```
huwb, line 32: is this ok adding here? if you already made the new release this may need to be moved

docs/user/collision-shape-for-physics.rst
```diff
@@ -18,25 +18,17 @@ Use of these is demonstrated in the example content.
...
 
-.. sponsor::
+Collision API Usage
+-------------------
 
-   Sponsoring us will help increase our development bandwidth which could work towards using Burst/Jobs for this feature.
+The collision providers built intout our system perform queries asynchronously; queries are offloaded to the GPU or to spare CPU cores for processing.
+This has a few non-trivial impacts on how the query API must be used.
 
-   .. trello:: https://trello.com/c/qUvB1aSO
```
huwb, line 29: i think this trello could probably be completed as we are technically doing this now - burst job queries.

perhaps we should make a new card for overcoming some of the limtations listed below such as not support depth caches (and perhaps allowing varying water level). that might give the opportunity to retain the sponsorship message.

docs/user/collision-shape-for-physics.rst
```diff
@@ -51,8 +43,40 @@ Posting the query and polling for its result are done through the same function.
...
+-  **Time Resolution** - Frames per second of baked data. Larger values may help the collision track the surface closely at the cost of more frames and increase baked data size.
+-  **Smallest Wavelength Required** - Smallest wavelength required in collision. To preview the effect of this, disable power sliders in spectrum for smaller values than this number. Smaller values require more resolution and increase baked data size.
+-  **Time Loop Length** - FFT waves will loop with a period of this many seconds. Smaller values decrease data size but can make waves visibly repetitive.
+
+Next click **Bake to asset and assign to current settings** and select a path and filename for the result.
+After the bake completes the current active *Animated Waves Sim Settings* will be configured to use this data.
+
+There are currently a few key limitations of this approach:
+
+-  Only a single set of waves from one *Shape FFT* component is supported. This collision does not support multiple sets of waves.
```
huwb, line 73: perhaps these should be some kind of formatted box to highlight them more, but lmk what you think

huwb (commented):
i took a swing at docs. added a few comments/questions. besides these i think this is good to merge.

docs/about/history.rst
daleeidd: Ah yes sorry. Will have to be moved into 4.15 which I have setup in master.

docs/user/collision-shape-for-physics.rst
daleeidd: Made this one: https://trello.com/c/EJCQhvsL

LMKWYT

docs/user/collision-shape-for-physics.rst
daleeidd: Good idea. I checked using "important" and it worked well.

docs/user/collision-shape-for-physics.rst
```diff
@@ -51,8 +43,40 @@ Posting the query and polling for its result are done through the same function.
...
+
+This is the default and recommended choice for when a GPU is present.
+Query positions are uploaded to a compute shader which then samples the ocean data and returns the
+desired results.
+The result of the query accurately tracks the height of the surface, including all wave components and depth caches and other Crest features.
+
+
+Baked FFT Data (CPU)
+--------------------
+
```
daleeidd, line 58: If this is going to be a preview feature, then there an admonition for it:
```rst
.. admonition:: Preview
```

daleeidd (changes requested):
Just the exception needs fixing. Otherwise good for merging.

docs/about/history.rst
huwb: done

docs/user/collision-shape-for-physics.rst
huwb: that card looks good. im out of time so i cant change the docs but i could probably do it in future work

docs/user/collision-shape-for-physics.rst
huwb: done

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
```diff
@@ -397,12 +410,133 @@ public bool Validate(OceanRenderer ocean, ValidatedHelper.ShowMessage showMessag
...
+                    var result = FFTBaker.BakeShapeFFT(target as ShapeFFT);
+                    if (result != null)
+                    {
+                        OceanRenderer.Instance._simSettingsAnimatedWaves.CollisionSource = SimSettingsAnimatedWaves.CollisionSources.BakedFFT;
+                        OceanRenderer.Instance._simSettingsAnimatedWaves._bakedFFTData = result;
+                        Selection.activeObject = OceanRenderer.Instance._simSettingsAnimatedWaves;
+
+                        // Rebuild ocean
+                        OceanRenderer.Instance.enabled = false;
+                        OceanRenderer.Instance.enabled = true;
```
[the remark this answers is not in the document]
huwb: hm Instance is checked for null above. is it the reference to the settings that is throwing? that is unchecked and as a quick solution i've now disabled the "bake and assign" if there is no settings avail to assign to.

if it is actually this line and you have a repro, would you perhaps have a fix for it? just checking if a fix is obvious and can easily be made. if its not obvious let me konw if you have repro steps and ill try here.

docs/user/collision-shape-for-physics.rst
huwb: yeah i guess it might be a good shout as the baked data might get breaking changes after more testing?

would you have a suggestion how to use that tag - should the entire section be indented, or should i add a new line inside the section saying that its in preview and use the tag there?

daleeidd: Sure. Content needs to be indented and there needs to be an empty line between tag and content. It is used on the ocean simulation page. Example:
```rst
.. admonition:: Preview

   Some content.
```

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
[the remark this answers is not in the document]
daleeidd: The line before it disables the ocean and which Instance is set to null. We can either add a method on the ocean to handle this since it then won't rely on the static reference, or we could use hashy.

Not sure how to add to this PR myself, but here is the method on OR:

```csharp
        public void Rebuild()
        {
            enabled = false;
            enabled = true;
        }
```

docs/user/collision-shape-for-physics.rst
huwb: Ok so I'll indent the body and put it on the tag, but not the header.

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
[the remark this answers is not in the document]
huwb: Oh I did think about that, but it never happened for me so I dismissed it. Weird. That makes sense though. I'll call that function instead.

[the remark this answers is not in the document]
huwb: added that function to OR, thanks

re adding to pr, in retrospect i should probably create a branch in this depot for us to iterate for cases like this. to push to the branch in this pr, you'd need to ask for write access to the other remote and add the remote to your local git workspace. not sure its worth changing anything now but good to note going forwards.

docs/user/collision-shape-for-physics.rst
huwb: done now

huwb:
All feedback should be addressed

daleeidd:
@kpietraszko if it isn't too much trouble may I have write access to your repository to make changes to this branch?

daleeidd (changes requested):
Nice. I think this will be the last round of changes.

kpietraszko:
> @kpietraszko if it isn't too much trouble may I have write access to your repository to make changes to this branch?

Sorry didn't get a notification. Sent you an invite to the repo.

crest/.vsconfig
```diff
@@ -0,0 +1,6 @@
+﻿{
```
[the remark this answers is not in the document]
daleeidd: Deleted

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
```diff
@@ -397,12 +410,134 @@ public bool Validate(OceanRenderer ocean, ValidatedHelper.ShowMessage showMessag
...
+            var bakeAndAssignLabel = "Bake to asset and assign to current settings";
+            var selectCurrentSettingsLabel = "Select current settings";
+            if (OceanRenderer.Instance._simSettingsAnimatedWaves != null)
+            {
+                if (GUILayout.Button(bakeLabel))
+                {
+                    FFTBaker.BakeShapeFFT(target as ShapeFFT);
+                }
+
+                GUI.enabled = OceanRenderer.Instance._simSettingsAnimatedWaves != null;
```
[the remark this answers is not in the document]
daleeidd: Fixed.

docs/user/collision-shape-for-physics.rst
daleeidd: I've changed it so preview contains just a short note: This feature is in preview and may change in the future.

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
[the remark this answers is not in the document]
daleeidd: Thanks. Looks good.

daleeidd:
> > @kpietraszko if it isn't too much trouble may I have write access to your repository to make changes to this branch?
> 
> Sorry didn't get a notification. Sent you an invite to the repo.

No worries. Thank you for adding me.

docs/user/collision-shape-for-physics.rst
daleeidd: Done.

daleeidd (approved):
Ready to merge when you're ready.

huwb:
Thanks for jumping in on those things, super appreciated. Merging.

PR #882, opened 2021-07-26
Fix FFT material allocations every frame

daleeidd:
Reported to me on Discord (I think) that FFT was creating a material every frame. Fixed it by caching the two material options.


huwb (approved):
Oh crazy. Thanks for fixing that

daleeidd:
Thanks

PR #883, opened 2021-07-27
Add default textures to ocean shader. Update other defaults

daleeidd:
Adds default textures to shader. Also updates the shader default values based on current material and URP. So now when creating a new ocean material it looks similar to the ocean material we provide.


huwb (approved):
Looks good thanks

daleeidd:
Thank you

Issue #800, opened 2021-04-07
MSAA causes outline on objects in front of ocean


metromeo:
**Describe the bug**
msaa not works correctly in hdrp

**Screenshots / video**
https://gfycat.com/damagedpinkladybug

**Versions**
HDRP v. 10.3 10.4
Crest 4.8 4.9, 
Unity 2020.2.7 2020.3.3

**To Reproduce**
Place any mesh with opaque material above water with another mesh with opaque material in background underwater

**Platform**
 - Editor and standalone
 - Win

**Hardware**
GForce GTX 1070, radeon 570

daleeidd:
Thanks. Can I also grab a screenshot of the material for the object in the front?

daleeidd:
Also, which shader are you using for the Terrain?

metromeo:
Terrain material is simple hdrp/terrainlit
I reproduced this in main scene, just added black plane (which also on the first gif, under water with terrain) for difference with white terrain
also i uploaded part of the mesh, that have many edges - https://dropmefiles.com/5TMaZ

![1](https://user-images.githubusercontent.com/5053538/113973356-ddb8b580-9844-11eb-9ba4-ec2da32dae53.png)
![2](https://user-images.githubusercontent.com/5053538/113973363-e0b3a600-9844-11eb-9cd3-5c1e60d5b224.png)
![3](https://user-images.githubusercontent.com/5053538/113973364-e0b3a600-9844-11eb-9f55-d9c6b8d18446.png)
![4](https://user-images.githubusercontent.com/5053538/113973365-e0b3a600-9844-11eb-9c91-f322750aa206.png)

daleeidd:
Thanks. I was able to reproduce. Not sure if I can solve this yet. If you look at the mesh up close you will see a fine outline when MSAA is enabled. And when you zoom out, like you have, it becomes more and more prominent. This is most likely caused by using something that has MSAA applied with something that doesn't have MSAA applied in our shader. It will need further investigation.

daleeidd:
When we sample the depth buffer, choosing the maximum value around -and including- the sample point greatly reduces the issue. It still doesn't look great though and that is five samples (maybe more if using refraction).

I think having access to the MSAA depth might help with this. Although initial attempt was unsuccessful. Disappointing that this sort of thing is still an issue for developers to wrangle.

huwb:
[...]

Issue #529, opened 2020-06-05
White edges around objects [URP]


woistjadefox:
When the camera background color is set to solid and white I get these edges around objects which are in front of the ocean / water. Happens also with the underwater curtain. 
![ezgif-3-0678ba17aa2b](https://user-images.githubusercontent.com/9056181/83850570-6a6c8180-a711-11ea-80a3-fe80903648b0.gif)

Happens with Unity 2019.3.14f1 and the URP Crest v4.2 in the editor game view window and in standalone windows builds. Testing it on a GTX 2060s. 

In the discord channel we talked about this issue and @huwb mentioned that it might has something to do with the "averaged" depth buffer from Unity.

woistjadefox:
What I forgot. It has also something to do with MSAA. It's not happening with MSAA turned of on the camera.

huwb:
Thanks @woistjadefox .

I haven't quite understood the gif above. Those objects are in front of the water are they?

OkoGoran:
Ah sorry, I couldn't show more of the picture. Yes, objects in front of the water on or below the surface.

woistjadefox:
Ah sorry, I couldn't show more of the picture. Yes, objects in front of the water on or below the surface.

daleeidd:
This is likely caused by a discrepancy between the depth texture and the opaque texture.

I couldn't reproduce this with MSAA using Crest 4.6 and Unity 2019.4. I could reproduce this if downsampling was applied to the camera opaque texture. Can you please re-test this and also try setting downsampling to none?

daleeidd:
Fixed MSAA outline in 4.13.

PR #887, opened 2021-08-12
Support MSAA

daleeidd:
Related Issues: #800 #529

Searches for the maximum depth value. It adds four additional samples for the ocean renderer and four additional samples for the underwater renderer. The end result is a dark edge around objects instead of a light edge which is preferable.

1_Broke
Without Fix

1_Fixed
With Fix

The offset amount is driven by scripting as it needs to be "2" for HDRP if refraction is used; otherwise, the issue will reoccur. I cannot work out why this is a problem for HDRP and not BIRP.

S (283)
HDRP refraction with 1 offset (gets worse with higher refraction)

S (284)
HDRP refraction with 2 offset

The approach fails in certain situations where objects are close together:

S (285)

There is a way to sample from the MSAA depth texture for HDRP which produces perfect results, but the texture is not accessible if Render Graph is used which is going to be the default beyond HDRP 10. I made a post asking for assistance: https://forum.unity.com/threads/how-to-sample-from-_depthtexturems-when-render-graph-is-enabled.1155200/

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Helpers/Helpers.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Helpers/Helpers.cs
@@ -1,0 +1,19 @@
+// Crest Ocean System
+
+// This file is subject to the MIT License as seen in the root of this folder structure (LICENSE)
+
+namespace Crest
+{
+    using UnityEngine;
+
+    /// <summary>
+    /// General purpose helpers which, at the moment, do not warrant a seperate file.
+    /// </summary>
+    public static class Helpers
+    {
+        public static bool IsMSAAEnabled(Camera camera)
+        {
+            return camera.allowMSAA && QualitySettings.antiAliasing > 0f;
+        }
+    }
+}
--- a/crest/Assets/Crest/Crest/Shaders/OceanShaderHelpers.hlsl
+++ b/crest/Assets/Crest/Crest/Shaders/OceanShaderHelpers.hlsl
@@ -7,6 +7,22 @@
 #ifndef CREST_OCEAN_SHADER_HELPERS_H
 #define CREST_OCEAN_SHADER_HELPERS_H
 
+// Sample depth macros for all pipelines. Use macros as HDRP depth is a mipchain which can change according to:
+// com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
+#if defined(SHADERGRAPH_SAMPLE_SCENE_DEPTH)
+#define CREST_SAMPLE_SCENE_DEPTH(coordinates) SHADERGRAPH_SAMPLE_SCENE_DEPTH(coordinates)
+#elif defined(TEXTURE2D_X)
+#define CREST_SAMPLE_SCENE_DEPTH(coordinates) SAMPLE_DEPTH_TEXTURE(_CameraDepthTexture, sampler_CameraDepthTexture, coordinates)
+#elif defined(SAMPLE_DEPTH_TEXTURE)
+#define CREST_SAMPLE_SCENE_DEPTH(coordinates) SAMPLE_DEPTH_TEXTURE(_CameraDepthTexture, coordinates)
+#endif
+
+#if UNITY_REVERSED_Z
+#define CREST_DEPTH_COMPARE(depth1, depth2) min(depth1, depth2)
+#else
+#define CREST_DEPTH_COMPARE(depth1, depth2) max(depth1, depth2)
+#endif
+
 // Same as LinearEyeDepth except supports orthographic projection. Use projection keywords to restrict support to either
 // of these modes as an optimisation.
 float CrestLinearEyeDepth(const float i_rawDepth)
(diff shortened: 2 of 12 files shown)


crest/Assets/Crest/Crest/Scripts/Helpers/Helpers.cs
```diff
@@ -0,0 +1,19 @@
+// Crest Ocean System
```
huwb, line 1: I'm guessing there isn't already another helpers file that this could go into? If not this is fine.

Adding/removing files can be a pain with assets so just checking.

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Mask.cs
```diff
@@ -64,6 +66,21 @@ void OnPreRenderOceanMask()
             );
         }
 
+        internal static void AdjustNearPlaneForMSAA(Camera camera, CommandBuffer buffer)
+        {
+            if (Helpers.IsMSAAEnabled(camera))
+            {
+                // @MSAAOutlineFix:
+                // The edge of the ocean surface at the near plane will be MSAA'd leaving a noticeable edge. By rendering
```
huwb, line 74: Interesting, thanks for the note. If this offset is larger, does it cause any issue or just push it back gracefully? Just checking if the offset should be proportional to frustum length or such or if approximate is fine.

crest/Assets/Crest/Crest/Shaders/OceanShaderHelpers.hlsl
```diff
@@ -40,4 +56,24 @@ float CrestLinearEyeDepth(const float i_rawDepth)
...
+// Works for all pipelines.
+float CrestMultiSampleSceneDepth(const float i_rawDepth, const float2 i_positionNDC)
+{
+	float rawDepth = i_rawDepth;
+
+	if (_CrestDepthTextureOffset > 0)
+	{
+		// We could use screen size instead.
+		float2 texelSize = _CameraDepthTexture_TexelSize.xy;
+		int3 offset = int3(-_CrestDepthTextureOffset, 0, _CrestDepthTextureOffset);
```
huwb, line 68: If the offset is destined to be used as a float (exclusively), it may be worth making it all floats. Having said that I was unable to find out how much type conversion costs on GPU so hard to know if it matters. Will go with your judgement.

huwb (approved):
Just minor comments. Looks great nicely done

crest/Assets/Crest/Crest/Shaders/OceanShaderHelpers.hlsl
daleeidd: I'll be using it with Load too (maybe exclusively) once it is in SRP.

huwb: Ok cool

crest/Assets/Crest/Crest/Scripts/Helpers/Helpers.cs
daleeidd: Not that I can see. The plan is to never delete this file even if it becomes empty. It will allow us to add any helper functions that don't warrant a separate file.

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Mask.cs
daleeidd: In certain cases where the near plane intersects a wave, like when looking down, the darkened outline is much thicker so making it larger would make that case worse.

What is the benefit in making it proportional?

huwb: Ah ok. Was just wondering if it would help us to work on a wider range of scenarios. However what you describe wouldn't be helped by my suggestion and I think it's fine as is.

crest/Assets/Crest/Crest/Shaders/Underwater/UnderwaterMaskShared.hlsl
```diff
@@ -101,6 +101,16 @@ Varyings Vert(Attributes v)
...
 {
+#if CREST_HANDLE_XR
+	// @MSAAOutlineFix:
+	// The edge of the ocean surface at the near plane will be MSAA'd leaving a noticeable edge. By rendering the mask
+	// with a slightly further near plane, it exposes the edge to having the underwater fog applied which is much nicer.
+	if (_CrestDepthTextureOffset > 0 && LinearEyeDepth(input.positionCS.z) < (_ProjectionParams.y + 0.001))
+	{
+		discard;
+	}
+#endif
```
daleeidd, line 112: @huwb For XR SPI I am doing it in the shader as XR SPI doesn't accept changing the matrices.

huwb: ok, is it worth adding this note as a comment?

daleeidd: Can do. Actually... I probably should only do it this way. I haven't noticed any issues, but I just realised that changing the near plane in C# will mess with LinearEyeDepth working correctly since it uses _ZBufferParams which should be same as the main depth buffer. Thoughts?

daleeidd: > i feel that _ZBufferParams should be correct if you change the near plane?

It's only for this pass which is rendered into its own depth texture. I was thinking when sampled later on the _ZBufferParams values will be computed from the non modified near far plane.

Another issue illustrated in the following screenshots is the underwater effect rendering the device depth to the screen. One with no near plane offset and one with a 10 near plane offset. The one with the modified near plane produces a depth value which is incorrect in context. It's an extreme case so it might be fine with the very small offset (0.001) I am using.


Ocean mask depth texture rendered (0 near plane offset)


Ocean mask depth texture rendered (10 near plane offset)

Hopefully it makes sense.

daleeidd: Went with doing it in the shader only and removing the C# one due to above issues.

daleeidd:
Thank you.

PR #888, opened 2021-08-13
Scale waves by factor shader

daleeidd:
Finally got around to finishing this off. I have added feathering, texture input and an invert option. Also added documentation and release notes.

I couldn't see a difference with the "Follow Horizontal Motion" option which is strange.


crest/Assets/Crest/Crest/Shaders/OceanInputs/AnimWavesScaleByFactor.shader
```diff
@@ -0,0 +1,123 @@
...
+			half4 Frag( Varyings input ) : SV_Target
+			{
+#if _TEXTURE_ON
+				float scale = tex2D(_MainTex, input.uv).r;
+#else
+				float scale = _Scale;
+#endif
+
+#if _INVERT_ON
+				scale = 1 - scale;
```
huwb, line 111: Minor but 1.0

crest/Assets/Crest/Crest/Shaders/OceanInputs/AnimWavesScaleByFactor.shader
```diff
@@ -0,0 +1,123 @@
...
+		[Toggle] _FeatherAtUVExtents("Feather At UV Extents", Float) = 0
+		// How far from edge to feather.
+		_FeatherWidth("Feather Width", Range(0.001, 0.5)) = 0.1
+	}
+
+	SubShader
+	{
+		Pass
+		{
+			Blend Zero SrcColor
```
huwb, line 34: Is this multiply? I guess it must be of it's working, just my brain not working right now!

crest/Assets/Crest/Crest/Shaders/OceanInputs/AnimWavesScaleByFactor.shader
```diff
@@ -0,0 +1,123 @@
...
+
+// This file is subject to the MIT License as seen in the root of this folder structure (LICENSE)
+
+// This writes straight into the displacement texture and sets the water height to the y value of the geometry.
+
+Shader "Crest/Inputs/Animated Waves/Scale By Factor"
+{
+	Properties
+	{
+		[Enum(ColorWriteMask)] _ColorWriteMask("Color Write Mask", Int) = 15
```
huwb, line 11: This may cause confusion but if you have a use case where only some channels should be scaled then I'm not against it

crest/Assets/Crest/Crest/Shaders/OceanInputs/AnimWavesScaleByFactor.shader
```diff
@@ -0,0 +1,123 @@
...
+
+		// Scale the waves. Zero is no waves and one leaves waves untouched.
+		_Scale("Scale", Range(0, 1)) = 0.35
+
+		// Use the texture instead of the scale value.
+		[Toggle] _Texture("Apply Texture", Float) = 0
+		_MainTex("Texture", 2D) = "black" {}
+
+		// Inverts the scale value.
+		[Toggle] _Invert("Invert", Float) = 0
```
huwb, line 21: Happy for this to go in but would be curious what the use case is? Just checking it's needed.

docs/user/ocean-simulation.rst
```diff
@@ -68,6 +68,8 @@ The following input shaders are provided under *Crest/Inputs/Animated Waves*:
 -  **Add From Texture** allows any kind of shape added to the surface from a texture.
    Can ether be a heightmap texture (1 channel) or a 3 channel XYZ displacement texture.
    Optionally the alpha channel can be used to write to subsurface scattering which increases the amount of light emitted from the water volume, which is useful for approximating light scattering.
+-  **Scale By Factor** scales the waves by a factor where zero is no waves and one leaves waves unchanged.
```
huwb, line 71: Nice one

crest/Assets/Crest/Crest/Shaders/OceanInputs/AnimWavesScaleByFactor.shader
daleeidd: Makes it more flexible with available textures. I was using the standard particle texture for testing which is 1 (no scaling) in the middle and thought it might be useful to have. Happy to remove it if you think it isn't worth keeping.

crest/Assets/Crest/Crest/Shaders/OceanInputs/AnimWavesScaleByFactor.shader
daleeidd: Fixed

crest/Assets/Crest/Crest/Shaders/OceanInputs/AnimWavesScaleByFactor.shader
daleeidd: I cannot. It's now removed.

crest/Assets/Crest/Crest/Shaders/OceanInputs/AnimWavesScaleByFactor.shader
huwb: No that's a fair point, good like that then

crest/Assets/Crest/Crest/Shaders/OceanInputs/AnimWavesScaleByFactor.shader
daleeidd: It is. I have added a comment.

docs/user/ocean-simulation.rst
daleeidd: Thanks :^)

huwb:
I thought for a moment this could be made into a "scale anything" shader that can be used in any input. Like "ocean inputs/all/scale". However that may be less user friendly/direct than what you have

daleeidd:
> I thought for a moment this could be made into a "scale anything" shader that can be used in any input. Like "ocean inputs/all/scale". However that may be less user friendly/direct than what you have

That's not a bad idea. And also true. Hmmm. We could add a material called scale waves to give people an idea of how it can be used?

huwb:
Yeah we could do. I think this would work for advanced-enough users that are comfortable configuring the shader/setting up the geo with the material.

Going off on a massive tangent - this reminds me that there would ideally be a super user friendly path to using this:

* create "register anim waves input" component
* click fix button to add a spline
* click fix button to create 2 spline points
* select "Scale Waves" on the "register anim waves input" component

This would do everything to make some spline geo scale the waves up or down (scale values would be exposed on spline data components).

This will require more change and might not be worth tackling now, but just braindumping. We should build less flexible but more user friendly paths like this wherever possible.

daleeidd:
> Yeah we could do. I think this would work for advanced-enough users that are comfortable configuring the shader/setting up the geo with the material. Which is required to use this shader in this PR anyway. So yeah thumbs up for that if you want to go for it.

Ok. Will do.

> Going off on a massive tangent - this reminds me that there would ideally be a super user friendly path to using this:
> 
> * create "register anim waves input" component
> * click fix button to add a spline
> * click fix button to create 2 spline points
> * select "Scale Waves" on the "register anim waves input" component
> * tweak scales on spline points
> 
> This would do everything to make some spline geo scale the waves up or down (scale values would be exposed on spline data components).
> 
> This will require more change and might not be worth tackling now, but just braindumping. We should build less-flexible-but-more-user-friendly paths like this wherever possible.

Agreed. This could also go with using DrawMesh instead of using mesh renderers which will further simplify things from a user and code perspective.

huwb (approved):
Looks perfect, nice one

daleeidd:
Thanks.

PR #890, opened 2021-08-16
Upgrade spectrums to fix "big wave" issue when reverted

huwb:
I occasionally get the "big waves" issue when reverting the wave assets, so i thought we should just bake in the version number which should sort it (seems to work for me).


daleeidd (approved):
Looks good.

huwb:
Cheers

Issue #878, opened 2021-07-20
ShapeFFT and ShapeGerstner produce lines in foam simulation data


daleeidd:
**Describe the bug**
Reported by @klauskobald on Discord. When using the ShapeFFT, foam simulation data appears to have grid lines.

**Screenshots / video**
![Evidence](https://cdn.discordapp.com/attachments/559872374682681345/867152697610469416/unknown.png)

**Versions**
Latest

**To Reproduce**
1. In the main scene, swap ShapeGerstnerBatched for ShapeFFT.
2. Reduce foam fade rate to something like 0.2 to make it more prominent.

**Platform**
 - Editor and Standalone
 - MacOS and Windows

**Additional context**
I tried disabling the variance to see if it was that but had no effect from what I could see.

huwb:
Thanks for posting. I don't immediately know what works cause this but will post if I figure it out

gamemachine:
This is not specific to fft, it happens with shape gerstner also.

daleeidd:
It also affects ShapeGerstnerBatched, but it is not noticeable with a full spectrum.

@huwb I think ShapeGerstner and ShapeFFT are more susceptible because they are tiled? Rendering the Jacobian shows the issue still. The foam data from the ShapeGerstnerBatched appears more stable. An option could be to add noise or blur the foam data.

huwb:
Would it be possible to see it happening for SGB? With a simplified spectrum would be useful.

daleeidd:
Sure. Test branch here: https://github.com/wave-harmonic/crest/tree/test/foam-grid-issue

Test scenes are under Test folder. There is one scene per shape type.

huwb:
Thanks. Weirdly i suddenly seem to repro it easily by just dropping in an FFT component.

It looks indeed like the wave patches are not tiling. I see lines every 0.5m which is exactly the size of the first wave cascade. If i turn off ~0.5m waves in the spectrum sliders, they go away.

I don't know why the waves are not tiling properly, the FFT was not messed with too much. so it is supposed to be strictly repeating patches (FFT is not able to do non-repeating). Might be worth comparing the spectrum with other similar code, or hacking the spectrum to only have one non-zero value to produce a single wave and try to show that it clearly does not tile.

Will post again if i make progress.

daleeidd:
Fixed with 4.14.

klauskobald:
Has this really been fixed?
I still get lines after updating from 4.13 - 4.15.
[...]

Issue #671, opened 2020-10-20
Static lighting sky and foam bubbles create weird square shapes


CheoCD:
Hello, I'm currently using Crest HDRP in an empty project on Unity 2020.1.8f1 with HDRP 8.2.0. The following issue happens in both the editor and a build, and I'm using Windows and a GTX 1070.

When a static lighting sky is set through an environment profile in the lighting panel, square shapes are created in the ocean as shown by the images below. The strength and color of those shapes are determined by the "Foam bubbles coverage" and "Foam bubbles colour" values in the default "Ocean" and "Ocean-Underwater" materials - setting the foam bubbles coverage value to 0 or the color to black removes the shapes.

![Crest1](https://user-images.githubusercontent.com/50143630/96592564-d29ab900-12e8-11eb-8f1d-67ca9675614b.png)
![Crest2](https://user-images.githubusercontent.com/50143630/96592580-d75f6d00-12e8-11eb-8ac9-61aefd94217e.png)
![Crest3](https://user-images.githubusercontent.com/50143630/96592586-d8909a00-12e8-11eb-8abf-dafff373e455.png)

I hope this bug report will be useful and that you will be able to solve this issue :)

daleeidd:
Thank you for reporting.

I have removed HDRP since it is an issue for all pipelines.

The issue is that foam bubbles requires a higher mesh density (mesh density is affected by the detail params _Lod Data Resolution_ and _Geometry Down Sampling Factor_) to look good at that distance. This shouldn't be the case as we want visual quality to be consistent. This can also happen when close to the surface if the mesh density is low enough.

We will have to investigate further. But for now I recommend scripting the foam bubble coverage (might have a different name for HDRP) so that no foam bubbles are shown when the camera is at that height.

huwb:
It might help to tweak the foam bubble texture scale - the texture itself should ideally help break up the regularity of the mesh triangles.

daleeidd:
Fix will be out for 4.13

daleeidd:
4.13 has been release. Thank you for reporting.

Issue #635, opened 2020-08-27
Waves and foams disappeared


willcatkins:
Hello, I'm using dynamic waves. It seems to be associated with the distance between the camera and the LOD of the ocean. When I get away from it, it will gradually disappear, but I hope it will always exist, even if its LOD is turned off.

But I didn't find a way to deal with this problem. Please tell me how to make the wave (foam) exist or change its LOD method. Thank you very much!

By the way,I have set the Ocean's MaxLOD to -1,because I need a very high altitude view.

daleeidd:
Could you explain your use case? It sounds like the loss of detail at high altitudes is causing issues?

> but I hope it will always exist, even if its LOD is turned off.

It sounds like you want rendering high detail across the entire surface which would invalidate the purpose of the LOD system. Dynamic waves shouldn't really be relied on for gameplay since they are not deterministic. Animated waves will be more persistent.

willcatkins:
Hi,sorry for late.

![Uploading 20200831204544.jpg…]()
![Uploading 20200831204549.jpg…]()
![20200831204552](https://user-images.githubusercontent.com/70247638/91734719-d9a52500-ebdd-11ea-8ec5-d6ff39139b0f.jpg)
![20200831204553](https://user-images.githubusercontent.com/70247638/91734726-dca01580-ebdd-11ea-84ea-fdf1dc86d9a1.jpg)
![20200831204559](https://user-images.githubusercontent.com/70247638/91734750-e6297d80-ebdd-11ea-8d2a-38ba5bc58cd9.jpg)

I don't  want rendering high detail,I just want to display the the wake of ships all the time.

willcatkins:
![20200831204544](https://user-images.githubusercontent.com/70247638/91735830-72887000-ebdf-11ea-92d4-5ea0a1f63229.jpg)
![20200831204549](https://user-images.githubusercontent.com/70247638/91736086-c1360a00-ebdf-11ea-9378-ef44ea22ed54.jpg)

Upload the first two pictures again~~

daleeidd:
Thanks. Can I please grab some more details like how high the camera is, a screenshot of the _Ocean Renderer_ component, and how strong the dynamic waves are for the boat wakes (_Strength_ on the material)?

daleeidd:
4.13 will have a fix which should solve this problem.

daleeidd:
4.13 is out. Please reopen if you still experience this issue.

PR #891, opened 2021-08-17
Sample foam in fragment shader

daleeidd:
Solves #671 and should help with #635.

Samples foam from the fragment shader. The difference is quite significant from a distance.

Vertex (Current):
1_Vert

Fragment:
1_Frag

Shoreline foam is greatly improved and the foam bubbles is no longer pixelated.

The foam bubbles could also be solved using a higher mesh density, but those screenshots already have a decent mesh density (384 resolution / 4 down sample). And adjusting the texture scale can produce passable results.

I have added it as a toggle/keyword for comparison. We could keep it optional. I've moved the camera in the main scene for testing.

Also, this won't look correct with ShapeGerstner or ShapeFFT because of #878. It makes the lines more prominent.

@huwb Let me know what you think.

### DIFF
--- a/crest/Assets/Crest/Crest/Shaders/Ocean.shader
+++ b/crest/Assets/Crest/Crest/Shaders/Ocean.shader
@@ -292,7 +292,7 @@ Shader "Crest/Ocean"
 			{
 				float4 positionCS : SV_POSITION;
 				half4 flow_shadow : TEXCOORD1;
-				half4 foam_screenPosXYW : TEXCOORD4;
+				half3 screenPosXYW : TEXCOORD4;
 				float4 lodAlpha_worldXZUndisplaced_oceanDepth : TEXCOORD5;
 				float3 worldPos : TEXCOORD7;
 				#if _DEBUGVISUALISESHAPESAMPLE_ON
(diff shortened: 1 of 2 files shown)


huwb:
It looks great.

Would getting some perf numbers be possible? I doubt its a huge hit but im just curious.

Thanks for this!

daleeidd:
> Would getting some perf numbers be possible? I doubt its a huge hit but im just curious.

Sure.

Performance on 1660 Ti:
Vertex: 2.93ms
Fragment: 2.99ms

> It looks great.

In the fragment screenshot I was accidentally boosting foam by combing vertex and fragment together which is now fixed. Some new screenshots:

Screenshot (289)
Vertex

Screenshot (290)
Fragment

Screenshot (291)
Vertex

Screenshot (292)
Fragment

huwb:
Oh great yeah lets go for it, looks better to me and small overhead

huwb (approved):
Looks good, nice one!

daleeidd:
Thanks.

crest/Assets/Crest/Crest/Shaders/Ocean.shader
```diff
@@ -540,11 +528,27 @@ Shader "Crest/Ocean"
 				// Foam - underwater bubbles and whitefoam
 				half3 bubbleCol = (half3)0.;
 				#if _FOAM_ON
+				float foam = 0.0;
+				// Data that needs to be sampled at the undisplaced position
+				if (wt_smallerLod > 0.001)
+				{
+					const float3 uv_slice_smallerLod = WorldToUV(input.lodAlpha_worldXZUndisplaced_oceanDepth.yz, cascadeData0, _LD_SliceIndex);
```
huwb, line 535: @daleeidd  Late feedback but it may be worth moving this up into the normal sampling code above. That would mean one set of of statements instead of two, and the UV would be computed once instead of twice.

PR #892, opened 2021-08-17
Make sphere water interaction standalone

huwb:
**Status**: Ready for review

First step in eradicating the Object Water Interaction component is to get Sphere Water Interaction up to scratch. @daleeidd as previously discussed, SWI is currently a pain to set up.

This change makes the component just work by itself with no dependencies on other things like the register component, meshrenderer, floating object. It also doesnt depend on the transform scale as that confuses people, it instead has adirect Radius control.

A remaining problem is that the SWI component doesnt seem to work at different scales. I have it tweaked for **main.unity**, it should look pretty good there. But if you increase the radius then the interaction blows up. I guess theres a scale factor missing somewhere.


daleeidd:
> A remaining problem is that the SWI component doesnt seem to work at different scales. I have it tweaked for main.unity, it should look pretty good there. But if you increase the radius then the interaction blows up. I guess theres a scale factor missing somewhere.

Not sure if I am seeing the problem, but it appears to be that 0.2 number in the shader. But it seems better without it though.

huwb:
The problem i was getting was if i would increase the Radius on the sphere water interaction component in main.unity on this branch, the dynamic waves would get huge.

I think i have incorporated the scale correctly now - the interaction looks good for me at every scale (at least from radius 0.1m to 2m).

Let me know if this makes sense and if you think this looks good to merge

huwb:
I cleaned up data in a latest commit, if you want to run the test in main.unity grab the penultimate-commit

daleeidd (approved):
Looks good. Didn't see any issues with these changes.

huwb:
Great thanks

PR #894, opened 2021-08-22
Render underwater horizon into mask

daleeidd:
Replaces C# horizon line code with a horizon line shader. It renders a quad near the far plane and compares the world position against the ocean level. It passed the test cases I had available and solved a few edge cases.

PR also includes meniscus fixes/improvements.

## Mask

Fixes the following edge cases:
- inverted horizon normal at certain camera transforms
- horizon instability when VR hardware is nearly aligned with horizon (last chapter of [video](https://www.youtube.com/watch?v=pGFjiWDinK0))

### Performance

Measuring the cost of the horizon shader. Measurements taken for the entire mask pass.

No waves and aligned with horizon (worst case scenario)
Old mask: 0.04ms
New mask: 0.14ms

Waves
Old mask: 0.11ms
New mask: 0.11ms

### Issues

I haven't experienced any. Will have to do some more testing with clip surface and underwater fog input branch. I tested against the "underwater from geometry" branch and seemed good.

## Meniscus

Also includes following meniscus fixes and improvements:
- Use _ScreenParams.xy instead of _ScreenParams.y. And apply ocean normal. Makes thickness more consistent
- Now works when there isn't much ocean surface visible

The latter is most prominent when the camera is aligned with horizon:

Before:
1_1_before

After:
1_2_after

This is possible (or simpler to do) now that the mask state is binary and covers the entire screen.

### Performance

No waves and aligned with horizon (worst case scenario)
Old effect: 0.76ms
New effect: 0.91ms

Waves
Old effect: 0.80ms
New effect: 1.60ms

Before the meniscus had a condition to only render when the mask sample contained the ocean surface. But that condition has been removed to improve the meniscus.


huwb (approved):
That's a massive improvement. Great work!

daleeidd:
Thank you

PR #897, opened 2021-08-29
Fix OceanDepthCache disabling in edit mode

daleeidd:
This PR replaces #629.

The OceanDepthCache disabling itself in edit mode is the primary issue this PR addresses.

While #629 is a complete and robust solution, it is quite heavy for the three components that will use it:
- OceanPlanarReflections
- OceanDepthCache
- UnderwaterEnvironmentalLighting

Disabling the component only saves Update or PreRender from running and for these components which normally only have one it should be fine (OceanDepthCache disables after start anyway). Also, OceanDepthCache hard dependent on OceanRenderer as it falls back to the transform height.

Instead this makes it so dependent components execute after the OceanRenderer and leaves managing delayed initialisation to the user if they really need it. This is done with DefaultExecutionOrder attribute. It is the same thing as setting the execution order in the project settings, except they don't show up in the list. It can be overridden by the user if they add their own entry.

Let me know what you think.

### DIFF
--- a/crest/Assets/Crest/Crest-Examples/Shared/Scripts/UnderwaterEnvironmentalLighting.cs
+++ b/crest/Assets/Crest/Crest-Examples/Shared/Scripts/UnderwaterEnvironmentalLighting.cs
@@ -15,9 +15,12 @@ namespace Crest
     /// (directional light, reflections, ambient etc) with the underwater depth. This works with vanilla lighting, but
     /// uncommon or custom lighting will require a custom solution (use this for reference).
     /// </summary>
+    [DefaultExecutionOrder(k_DefaultExecutionOrder)]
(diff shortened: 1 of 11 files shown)


daleeidd:
This was half thought out so converted to draft. Numerous components use the pattern which disables itself in Start if no OceanRenderer is present. So maybe the other PR is also useful still. I'll strip this down to fixing the depth cache only.

huwb:
Fwiw I approved 629, think is be in favour of merging that but you will know best.

daleeidd:
> Fwiw I approved 629, think I'd be in favour of merging that but you will know best.

Thanks. I won't merge it for this release as I would like to add some unit tests for that PR before merging.

PR #898, opened 2021-08-29
Fix Ocean Depth Cache disabling itself in edit mode

daleeidd:
It gets reported every so often that the _Ocean Depth Cache_ disables itself in edit mode. There is no hard requirement on an _Ocean Renderer_ being present so I have removed the check. I have added validation to notify users of the considerations.


huwb (approved):
Looks good

daleeidd:
Thanks.

PR #899, opened 2021-09-03
Clip surface input inverted option

daleeidd:
Adds an "inverted" option to both SDF and convex hull clip surface inputs. Both of these can now remove clip surface data.

With the new order property for SDF, one can create some interesting patterns:

SDFCH
Left: Convex Hull. Right: SDFs


huwb:
Ha thats awesome

huwb (approved):
Ah interesting that the order needs to be exposed. Nice one

daleeidd:
> Ah interesting that the order needs to be exposed. Nice one

Before we got the order from the shader/material render queue. But we don't expose them for SDFs.

huwb:
Yeah cool

PR #900, opened 2021-09-03
Fix no SSS value at extents

daleeidd:
The extents have no SSS factor resulting in popping and extents being noticeably different.

Issue
Extents can be seen in the right view.

Solved by setting SSS factor to 0.5 at last LOD.

### DIFF
--- a/crest/Assets/Crest/Crest/Shaders/Ocean.shader
+++ b/crest/Assets/Crest/Crest/Shaders/Ocean.shader
@@ -524,6 +524,14 @@ Shader "Crest/Ocean"
 					#endif
 				}
 
+#if _SUBSURFACESCATTERING_ON
+				// Extents need the default SSS to avoid popping and not being noticeably different.
+				if (_LD_SliceIndex == (_SliceCount - 1))
+				{
+					sss = CREST_SSS_MAXIMUM - CREST_SSS_RANGE;
+				}
+#endif
+
 				#if _APPLYNORMALMAPPING_ON
 				#if _FLOW_ON
 				ApplyNormalMapsWithFlow(positionXZWSUndisplaced, input.flow_shadow.xy, lodAlpha, cascadeData0, instanceData, n_pixel);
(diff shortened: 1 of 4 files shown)


huwb:
I think due to my historical errors we may be massively over-relying on SSS right now. I think it should probably only show through sharp wave crests in limited situations. Atlas gdc talk is a good ref for this. From that point of view SSS should be mostly 0 everywhere. I think doing that and tweaking a few other things makes it more realistic (perhaps to the point where it needs the full shading model they propose to get more realism).

image

so one solution might be to retweak our materials to reduce dependence on SSS and recommend others do the same, and keep SSS at 0 in background.

but maybe thats rocking the boat too much. if so im happy for this to merge. just be aware that 0.5 may not match every situation. but if you found it works well we could go for that.

huwb:
In case its useful i pushed `test\reduce-sss` which i think is probably more how it should be. the SSS term (pinch jacobian) seems to just flare up everywhere so i sharpened it to pick out crests only

daleeidd:
Thanks. I had a look and we still get popping with the test branch from _SSS Intensity Sun_. Also, increasing the range value exposes a minor issue where depending on the SSS strength the pinch highlights smaller waves rather than larger ones which looks strange:



Although, it isn't so bad once foam and foam bubbles are taken into account.

> just be aware that 0.5 may not match every situation. but if you found it works well we could go for that.

I've updated the code which covers both the current values and the test branch values since I now use the formula. I tested it on the test branch and popping is eliminated and seemed to match up.

> (perhaps to the point where it needs the full shading model they propose to get more realism)
> but maybe thats rocking the boat too much.

Is that the micro facets you're working on? We could swap it out once that arrives. I could also add a property to the shader to adjust the range value for now, but not much point if it will get removed later.

daleeidd:
Merged this in. I did some more testing and it worked well which will keep people happy for the time being. Agreed to replacing SSS approach with something more realistic and better.

Issue #676, opened 2020-10-25
How to display two different oceans with different materials in a scene


dogiracle:
How to display two different oceans with different materials in a scene

daleeidd:
At this stage only a single ocean instance is available. This is being explored here #650. I'm not sure if it would cover this use case though/ For now you can blend materials using [Material.Lerp](https://docs.unity3d.com/ScriptReference/Material.Lerp.html).

huwb:
An alternative which may work is to extend the water body script to allow override materials to be provided, so different water bodies could get different materials.  It probably won't be perfect but may get us some way there. I've been meaning to try it, will post if I do.

huwb:
This is still on the list, it seems to always be one or two items further down than what we have bandwidth to look at, but it is slowly bubbling up..

We're hoping to increase our bandwidth through sponsorship and i wanted to mention that once here in case there is interest in supporting us: https://github.com/sponsors/wave-harmonic

daleeidd:
This feature was added in 4.14 (#902).

PR #902, opened 2021-09-11
Water body material override

huwb:
**Status:** Ready for review

In response to a support request - add a material override field to water bodies. Seems to work fairly straightforwardly (IMO).

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/OceanChunkRenderer.cs
+++ b/crest/Assets/Crest/Crest/Scripts/OceanChunkRenderer.cs
@@ -34,6 +34,7 @@ namespace Crest
         public Renderer Rend { get; private set; }
         PropertyWrapperMPB _mpb;
 
+        public bool MaterialOverridden { get; set; }
 
         // We need to ensure that all ocean data has been bound for the mask to
         // render properly - this is something that needs to happen irrespective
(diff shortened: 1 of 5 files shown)


daleeidd:
Works great. Nice and simple. Should the OceanMaterial return the water body override material for the underwater renderer etc?

huwb:
Ha yeah good point. It's not clear to me at all how it will work for underwater. I would guess the underwater effect should search all water bodies and find any overrides, as a separate process. Or it should do something else entirely. But if you have an idea for a simple way it can be done (I wasnt sure if i understood if OceanMaterial can help) then let me know.

If its going to be an entirely separate thing im tempted to call it future work and see if it is requested in practice. But lmk what you think!

daleeidd:
Yeah, you're right. Separate process or something in the future sounds good.

huwb:
Ok cool. I've updated the status from 'experimental' to ready for review as i think this could merge

huwb:
Oh release notes. Oops.

huwb:
Added release notes and some misc doc stuff

daleeidd (approved):
Works great.

huwb:
Cheers

daleeidd:
Linking #676

PR #903, opened 2021-09-11
Make FFT use 4 samples per wave (min), not 2

huwb:
**Status**: Seems to work here but could use second pair of eyes

I think this fixes both the lines in foam #878 , and also recently asked question on discord about quality of a single band of FFT waves (by gamemachine i think).

Using 2 samples per wave in FFT is too low. This uses 4 samples per wave instead.

It seems like this fixes the lines in foam, which seem to have been down to bad quality of wave displacements. There are still repeating patterns in foam which is to be expected, but not obvious lines.

Also the Texels Per Wave param on the OceanRenderer component seems to need to be set to 2 for best results. I dont think this param is a good user facing feature, so I've committed to removing it to clean up the UI and hardcode what i feel should be a system var (and hard code it to nyquist limit).

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/LodData/LodTransform.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/LodTransform.cs
@@ -129,7 +129,8 @@ namespace Crest
             float oceanBaseScale = OceanRenderer.Instance.Scale;
             float maxDiameter = 4f * oceanBaseScale * Mathf.Pow(2f, lodIdx);
             float maxTexelSize = maxDiameter / OceanRenderer.Instance.LodDataResolution;
-            return 2f * maxTexelSize * OceanRenderer.Instance.MinTexelsPerWave;
+            float texelsPerWave = 2f;
+            return 2f * maxTexelSize * texelsPerWave;
         }
 
         public void SetOrigin(Vector3 newOrigin)
--- a/crest/Assets/Crest/Crest/Scripts/Collision/QueryBase.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Collision/QueryBase.cs
@@ -308,7 +308,8 @@ namespace Crest
             // The smallest wavelengths should repeat no more than twice across the smaller spatial length. Unless we're
             // in the last LOD - then this is the best we can do.
             float minWavelength = i_minSpatialLength / 2f;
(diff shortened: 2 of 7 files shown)


crest/Assets/Crest/Crest/Scripts/LodData/LodTransform.cs
```diff
@@ -129,7 +129,8 @@ public float MaxWavelength(int lodIdx)
             float oceanBaseScale = OceanRenderer.Instance.Scale;
             float maxDiameter = 4f * oceanBaseScale * Mathf.Pow(2f, lodIdx);
             float maxTexelSize = maxDiameter / OceanRenderer.Instance.LodDataResolution;
-            return 2f * maxTexelSize * OceanRenderer.Instance.MinTexelsPerWave;
+            float texelsPerWave = 2f;
```
daleeidd, line 132: ShapeGerstnerBatched looks strange with 2.

Screenshot (311)
With 2

Screenshot (310)
With 3

daleeidd (changes requested):
Looks much better. Can confirm that it reduces the grid lines and are pretty much eliminated.

Other than the other review comment, I noticed that foam exposes the LOD grid above eye level:
Screenshot (316)

At this height increasing the max scale (1024) can help make it look more uniform. But didn't help further down. Adjusting foam settings can help with uniformity too. But both workarounds look a little off.

crest/Assets/Crest/Crest/Scripts/LodData/LodTransform.cs
huwb: Yeah it's always been greater than 2 because it looked better for SGB, and this change degrades the quality. I propose we deprecate. SGB as a next step.

huwb:
>  I noticed that foam exposes the LOD grid above eye level

Do you mean the slightly regular pattern of foam patches in the screenshot? That might need to be addressed with spectrum tweaking.

If you are talking about something else perhaps you could annotate the image to highlight the issue.

daleeidd:
133221707-ec59f063-8170-4a29-bcbd-d81d8034705b

It is more obvious if I disable the lower wavelengths:



I had a look and looks like this behaviour already in master so you can ignore me as not relevant.

daleeidd (approved):

huwb:
Oh I see. Not sure what that's about. Will think about it. Merging this. Thanks!

PR #904, opened 2021-09-12
Store terrain height instead of computing water depth

huwb:
**Status**: Ready to merge

This stores terrain height into the depth data.

This is roughly equivalent to storing water depth, but is more general. It will make things simpler when the water level can vary for lakes and rivers. For example populating a depth cache for a river where water level varies is complicated, as the generation needs to know the water level. With this change, the depth cache just stores terrain height which is simple.

Anything that needs the water depth now calculates it as the difference between the sea level and the terrain height. The local water bodies branch will add another channel to this texture to store the sea level, so the calculation will eventually subtract the two channels rather than using the global sea level.

I searched all usages of `_LD_TexArray_SeaFloorDepth` and fixed them up.

The not great thing about this is it now stores 'terrain heights' in data that is called 'water depth' everywhere, which is confusing. However the local water bodies branch will add water water as a new channel and then this data sort of contains water depth, or at least data related to water depth, so its probably fine.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
+++ b/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
@@ -372,7 +372,7 @@ namespace Crest
 
         public static int sp_crestTime = Shader.PropertyToID("_CrestTime");
         readonly int sp_texelsPerWave = Shader.PropertyToID("_TexelsPerWave");
-        readonly int sp_oceanCenterPosWorld = Shader.PropertyToID("_OceanCenterPosWorld");
+        public static int sp_oceanCenterPosWorld = Shader.PropertyToID("_OceanCenterPosWorld");
         readonly int sp_meshScaleLerp = Shader.PropertyToID("_MeshScaleLerp");
         readonly int sp_sliceCount = Shader.PropertyToID("_SliceCount");
         readonly int sp_clipByDefault = Shader.PropertyToID("_CrestClipByDefault");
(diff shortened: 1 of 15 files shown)


crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
```diff
@@ -372,7 +372,7 @@ public static bool RunningWithoutGPU
 
         public static int sp_crestTime = Shader.PropertyToID("_CrestTime");
         readonly int sp_texelsPerWave = Shader.PropertyToID("_TexelsPerWave");
-        readonly int sp_oceanCenterPosWorld = Shader.PropertyToID("_OceanCenterPosWorld");
+        public static int sp_oceanCenterPosWorld = Shader.PropertyToID("_OceanCenterPosWorld");
```
daleeidd, line 375: Are we able to keep readonly?

daleeidd (approved):
Nice work. I didn't see any issues in review or testing. Just a minor thing but can be ignored.

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
huwb: it can. i tried to make others readonly but some of them are written in InitStatics and some not. i dont understand/remember why its like this. might be worth a comment in InitStatics if we can recall why some need to be written and some not.

huwb:
Thanks, merging.

Just a note @daleeidd - i've remembered that storing terrain height in a half may suffer from precision issues far from y=0. this is why it was stored as a water depth instead before, but its difficult to work with as depth. if issues arise with precision i propose we change the buffer to be 32bit floats.

PR #905, opened 2021-09-15
Fix white texture when sims are disabled with texture quality below full resolution (workaround)

daleeidd:
When "Project Settings > Texture Quality" is not "Full Res" our black texture array becomes white (probably junk). This is a bug with _Graphics.CopyTexture_ when using texture arrays. This can happen in builds when quality settings may be different (ie high in editor or low in build) than in editor which can be hard to debug.

Workaround is to set [_QualitySettings.masterTextureLimit_](https://docs.unity3d.com/ScriptReference/QualitySettings-masterTextureLimit.html) to zero before calling _Graphics.CopyTexture_. I have seen others do this for other bugs for this same method. We can also use SetPixels instead but this looks good enough. I have submitted a bug report to Unity for texture arrays.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Helpers/TextureArrayHelpers.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Helpers/TextureArrayHelpers.cs
@@ -78,7 +78,15 @@ namespace Crest
 
             for (int textureArrayIndex = 0; textureArrayIndex < LodDataMgr.MAX_LOD_COUNT; textureArrayIndex++)
             {
+                // There is a bug using Graphics.CopyTexture with Texture2DArray when "Texture Quality"
+                // (QualitySettings.masterTextureLimit) is not "Full Res" (0) where result is junk (white from what I
+                // have seen). Reported to Unity on 2021.09.15.
+                // https://issuetracker.unity3d.com/product/unity/issues/guid/1365775
+                // https://docs.unity3d.com/ScriptReference/QualitySettings-masterTextureLimit.html
+                var originalMTL = QualitySettings.masterTextureLimit;
+                QualitySettings.masterTextureLimit = 0;
                 Graphics.CopyTexture(texture, 0, 0, array, textureArrayIndex, 0);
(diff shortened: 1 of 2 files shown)


huwb:
Wow thats quite a workaround! Hope that does not cause a hitch at runtime.

Is there a bug number around for this, out of interest?

It occurs to me that an alternative could be to do the copy in one compute dispatch. Not 100% sure if it would work but seemed worth mentioning.

If thats what others are doing then thumbs up. Not sure if it needs a release note but approving anyway.

huwb (approved):

huwb:
(Total aside but it took me a while to figure out/remember what this function does. Is it creating a 'null' texture array to use when sims are disabled? And the copy is there because the input texture array gives the default 'null colour'? I think we could probably clarify the naming a lot as it sells itself as creating a clone of a texture array. Happy to propose something in a separate commit.)

daleeidd:
> Hope that does not cause a hitch at runtime.

I didn't notice anything. I tested it by running it every frame. But after reading it again, maybe it is a little dangerous. Fear is with a project with a lot of textures, it would load up all full resolution textures.

> Is there a bug number around for this, out of interest?

There wasn't. But there is now. I've added it as a comment.
https://issuetracker.unity3d.com/product/unity/issues/guid/1365775

> It occurs to me that an alternative could be to do the copy in one compute dispatch. Not 100% sure if it would work but seemed worth mentioning.

Like clear to black?

daleeidd:
I went with SetPixels as is safer than last approach and simple change. Works in editor and build from testing.

> (Total aside but it took me a while to figure out/remember what this function does. Is it creating a 'null' texture array to use when sims are disabled? And the copy is there because the input texture array gives the default 'null colour'? I think we could probably clarify the naming a lot as it sells itself as creating a clone of a texture array. Happy to propose something in a separate commit.)

Correct. Sure. Sounds good.

huwb:
Ah cool thanks for going the extra mile there, feel better about that

daleeidd:
Thank you.

PR #906, opened 2021-09-16
Fix underwater caustics jittering in XR when far from zero

daleeidd:
Reported on Discord that underwater caustics were jittering in XR. Player was far-ish from zero (650, 0, -5000). Video for reference:
https://www.dropbox.com/s/iry6krhqlptef0i/Caustics%20Underwater.mp4?dl=0

PR follows Unity's lead with reconstructing world-space position in the fragment shader using the depth which has been confirmed by user that fixes it. I also applied it to horizon as Unity does it for all fullscreen shaders in SRP.

I've taken functions from SRP/URP for BIRP.


huwb (approved):
Looks good

This might be just me but i interpreted Legacy as legacy code - something brought along from previous versions of crest, or something to support oldversions of unity. Maybe its fine now that i figured it out, but an alternative might be using the name BIRP or sth. But i'll leave it up to you.

daleeidd:
Thanks. Makes sense. Went with BIRP.

PR #907, opened 2021-09-16
Remove duplicate matrix calculation

daleeidd:
I was calculating this twice. Also, removes this code path from SRP as the IVP matrix is provided by Unity.


huwb (approved):
Nice oen

daleeidd:
Thanks

PR #910, opened 2021-09-19
Add sea floor offset to missing shaders

daleeidd:
Adds sea floor offset to shaders that need it.


huwb (approved):
Oh nice one, missed all of those!

daleeidd:
Thanks

PR #911, opened 2021-09-19
Fix/water level derivs

huwb:
Normally the ocean shader assumes water is horizontally flat - base water normal is (0,1,0). Then gersnter/normal map normals are added on top of this.

This is not the case for rivers and if you look at a shallow angle up a river, the fresnel looks totally wrong due to the missing consideration of the water gradient.

This computes derivatives of the sea level and feeds them to the shader to get correct base slope.

There were bad artifacts when depth target was 16bit, so switching to 32bit.


crest/Assets/Crest/Crest/Shaders/Ocean.shader
```diff
@@ -437,6 +439,8 @@ Shader "Crest/Ocean"
 
 			half4 Frag(const Varyings input, const bool i_isFrontFace : SV_IsFrontFace) : SV_Target
 			{
+				//return half4(abs(input.seaLevelDerivs), 0.,1.);
```
daleeidd, line 442: Forgot to remove

daleeidd (approved):
Nice one. Just one minor thing.

crest/Assets/Crest/Crest/Shaders/Ocean.shader
huwb: thx sorted

PR #913, opened 2021-09-22
Fix underwater high CPU memory usage

daleeidd:
I made a mistake with stripping as I was not stripping variants for the "__" (ie no keyword) case. It was reported on discord that underwater shader was using ~500mb. It is now under 25mb which puts it in line with other shaders.


huwb (approved):
I couldnt see an issue with this, nice one

The build processor is probably the code im least familiar with and i didnt follow everything, but probably would if i studied the code for longer. Maybe in the future it would be worthwhile adding more verbose comments, like maybe examples of things that will/wont be stripped or something. But maybe its clear enough and i just need to spend more time with it.

Approving anyway, nice one.

daleeidd:
Thanks. And I have improved the comments.

PR #914, opened 2021-09-24
Fix clip surface using incorrect slices

daleeidd:
An issue with clip surface not being accurate for large waves was reported and I discovered it was blowing out when there was horizontal displacement. Uncertain if it is the reported issue.

old
Blowout of the sphere which is clamped sample due to incorrect slice index.

new
Fixed.

### DIFF
--- a/crest/Assets/Crest/Crest/Shaders/Ocean.shader
+++ b/crest/Assets/Crest/Crest/Shaders/Ocean.shader
@@ -438,6 +438,36 @@ Shader "Crest/Ocean"
 				// We need this when sampling a screenspace texture.
 				UNITY_SETUP_STEREO_EYE_INDEX_POST_VERTEX(input);
 
+#if _CLIPSURFACE_ON
+				{
+					// Clip surface
+					half clipValue = 0.0;
+
+					uint slice0; uint slice1; float alpha;
+					PosToSliceIndices(input.worldPos.xz, 0.0, _CrestCascadeData[0]._scale, slice0, slice1, alpha);
+
+					const CascadeParams cascadeData0 = _CrestCascadeData[slice0];
+					const CascadeParams cascadeData1 = _CrestCascadeData[slice1];
+					const float weight0 = (1.0 - alpha) * cascadeData0._weight;
+					const float weight1 = (1.0 - weight0) * cascadeData1._weight;
+
+					if (weight0 > 0.001)
+					{
+						const float3 uv = WorldToUV(input.worldPos.xz, cascadeData0, slice0);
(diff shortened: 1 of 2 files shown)


huwb:
thanks for this. what is the fix actually? it was not easy to tell from the diff, i could not spot if any changes to the code itself. if it was simply moving it, could you give a hint about why that fixes it? sorry im sure these are dumb questions :)

daleeidd:
Sorry. That was a bit brief. I'll add some comments too. I now calculate the slice indices using PosToSliceIndices with the undisplaced position rather than using _LD_SliceIndex. This gave me the correct result.

I am not sure if it could also be solved by adding the displacement correction to the clip surface input like other inputs. And then we could used the _LD_SliceIndex again maybe. I tried but was unsuccessful in getting it correct.

huwb:
Oh, the pixel is getting displaced off the edge of the cascade. Didn't think of that.

PosToSliceIndices sites quite a bit of work to compute lod index from scratch. A lighter weight approach may be to detect if the UVs are going out of [0,1] and if so incrementing slice index to use the next larger cascade size and computing a new UV. 

That feels more efficient but you never know I suppose, if in doubt it might be worth profiling.

daleeidd:
Thanks. Will do. Putting this on hold as getting weird *Shader Compiler IPC Exception* on Windows. Submitted bug report to Unity. If they take too long I will use your approach without trying to compare them.

daleeidd:
@huwb Surprisingly, performance was about the same. The small difference was in PosToSliceIndices favour. I profiled three times for each:

No fix: 3.56, 3.53, 3.55ms
PosToSliceIndices: 3.64, 3.61, 3.62ms
Overflow fix: 3.66, 3.66, 3.69ms

The draft overflow fix function:
```hlsl
float3 WorldToUV2(in float2 i_samplePos, in CascadeParams i_cascadeParams, in float i_sliceIndex)
{
	float2 uv = (i_samplePos - i_cascadeParams._posSnapped) / (i_cascadeParams._texelWidth * i_cascadeParams._textureRes) + 0.5;
	float slice = i_sliceIndex;
	if (uv.x > 1.0 || uv.y > 1.0)
	{
		i_cascadeParams = _CrestCascadeData[slice + 1];
		uv = (i_samplePos - i_cascadeParams._posSnapped) / (i_cascadeParams._texelWidth * i_cascadeParams._textureRes) + 0.5;
		slice += 1.0;
	}
	return float3(uv, slice);
}
```

huwb (approved):
Nice one. There are some data changes which I'll trust you'll revert if not needed

daleeidd:
Thanks for the heads up.

PR #915, opened 2021-09-24
Fix FFTs not updating with spectrum changes when paused

daleeidd:
This is just a quick fix. Downside is that it runs at full rate instead of editor update intervals. Alternatives could be to add a CurrentUnscaledTime similar to CurrentTime. Or to hash the spectrum.

Let me know what you think.


huwb:
There's an updateSpectrum argument that is intended to support this. Maybe it broke somewhere along the way. The intention is that you can set that argument to true and it will force a generate.

daleeidd:
I totally missed that. That does work. Thanks.

huwb (approved):
Thanks great fix

daleeidd:
Thanks.

PR #916, opened 2021-09-26
Fix default spectrum breaking in editor when enabled in play mode for ShapeFFT

daleeidd:
This was an edge case that I missed. If ShapeFFT is enabled in play mode in the editor with a default spectrum, then it will not get the version upgrade it needs.

The fix was to also call the upgrade code in Awake in the editor.


huwb (approved):
Looks good thanks

Unrelated but we should maybe add a minimal example of a data upgrade, like an empty component with a single data field, just to provide a template. If I have time i might look at that today.

daleeidd:
Thanks.

I have been considering doing the same as it can be tricky. I added "Data Upgrade" to the private wiki and added the ISerializationCallbackReceiver example for components. Feel free to add the spectrum example. I'll add some more notes soon.

PR #917, opened 2021-09-29
Refactor ScatterColour

daleeidd:
In preparation for adding underwater shader API, I realised some refactoring was needed/desired for the ScatterColour function. There is some housekeeping in there too. Review by commit.

Most changes are to move out the underwater specific code out of the function and into the underwater shader.


huwb (approved):
Great stuff

daleeidd:
Thanks

PR #919, opened 2021-10-01
Make spline ribbon offset configurable and rework geo generation

huwb:
Ribbon can now be generated centered or on other side of spline.

Removed smoothing iterations which was arbitrary and only changed one side of the ribbon.

Explicitly remove overlaps/self intersections of spline ribbon geometry, and then do a bit of a blur at the end to soften.

Flow/waves look good even with quite extreme spline curvature.


huwb:
Will do docs if the code looks fine to you

crest/Assets/Crest/Crest/Scripts/Spline/Spline.cs
```diff
@@ -29,6 +29,16 @@ public partial class Spline : MonoBehaviour, ISplinePointCustomDataSetup
...
+            Left,
+            Center,
+            Right
+        }
+        [Tooltip("Where generated ribbon should lie relative to spline. If set to Center, ribbon is centered around spline.")]
+        // 01-Oct-21: Default to right for now to support legacy data. In the future we should update this default to Center.
+        public Offset _offset = Offset.Right;
+
         [Tooltip("Connect start and end point to close spline into a loop. Requires at least 3 spline points.")]
         public bool _closed = false;
```
daleeidd, line 44: Missed a smoothing iterations property below

crest/Assets/Crest/Crest/Scripts/Spline/Spline.cs
```diff
@@ -29,6 +29,16 @@ public partial class Spline : MonoBehaviour, ISplinePointCustomDataSetup
...
 #pragma warning restore 414
 
+        public enum Offset
+        {
+            Left,
+            Center,
+            Right
+        }
+        [Tooltip("Where generated ribbon should lie relative to spline. If set to Center, ribbon is centered around spline.")]
+        // 01-Oct-21: Default to right for now to support legacy data. In the future we should update this default to Center.
```
daleeidd, line 39: Could use the ISerializationCallbackReceiver and set version to one similar to clip surface primitives? Details: https://github.com/cowhunt/crest-wiki-private/wiki/Data-Upgrade. Being an MB means it should work fine.

crest/Assets/Crest/Crest/Scripts/Spline/Spline.cs
```diff
@@ -29,6 +29,16 @@ public partial class Spline : MonoBehaviour, ISplinePointCustomDataSetup
...
+        public Offset _offset = Offset.Right;
```
daleeidd, line 40: We could make this a 0-1 range? I don't know if the added flexibility it will give will be useful.

crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstnerSplineHandling.cs
```diff
@@ -142,71 +161,86 @@ public static bool GenerateMeshFromSpline<SplinePointCustomData>(Spline.Spline s
...
+                }
+                else
+                {
+                    pointsNew[i].x = lastGoodPoint.x;
+                    pointsNew[i].y = points[i].y;
+                    pointsNew[i].z = lastGoodPoint.z;
+                }
+            }
+            points = pointsNew;
         }
```
daleeidd, line 233: Some comments here will be helpful. Even small stuff like "Copying spline points" can make things easier.

daleeidd (changes requested):
Great work. Marked "requested changes" with eyeballs.

crest/Assets/Crest/Crest/Scripts/Spline/Spline.cs
huwb: yeah that works, and i guess is more explicit (and avoids future todo) so ive gone with that

crest/Assets/Crest/Crest/Scripts/Spline/Spline.cs
huwb: ah thanks done

crest/Assets/Crest/Crest/Scripts/Spline/Spline.cs
huwb: i spent some time considering if it should be a (-1, 1) value. i thought maybe it would confuse some people, and also wasnt sure it was needed. im still not 100% sure but this is where i landed. what do you think?

huwb: perhaps givign less flexibility and then potentially migrating to more flexibility in the future is preferable than giving too much and then deciding to take it away later

crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstnerSplineHandling.cs
huwb: yeah sorry, documented and tidied a couple of bits

huwb:
All done i think and added docs, re-requesting review

huwb:
Decided to commit test data that i keep juggling/recreating, hopefully this makes sense

crest/Assets/Crest/Crest/Scripts/Spline/Spline.cs
daleeidd: Make sense in regards to it is later better to give than to take. Let's go with what you have.

crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstnerSplineHandling.cs
daleeidd: Thanks.

crest/Assets/Crest/Crest/Scripts/Spline/Spline.cs
daleeidd: Thanks.

daleeidd (approved):
Works great. Nice work.

daleeidd:
> Decided to commit test data that i keep juggling/recreating, hopefully this makes sense

Makes sense. Normally it should have its own prefab etc but I don't see this scene being changed downstream so the setup you have should be fine.

huwb:
Ah yeah I took a few short cuts. Is probably fine for now. Thanks!

PR #922, opened 2021-10-05
Fix visible specks underwater

### LINKED ISSUE #739: Underwater specks/dots
underwater rendering shows up a lot of fireflies like shown below: small white pixel which do not receive the underwater effects.
crestFireflies

i tracked it down in renderdoc and it looks as if the ocean mask contains little 1px gaps here and there which i assume are caused by the gpu rejecting very small triangles (little black dots in the eyes)
oceanMaskFireFlies

so i started "filtering" the ocean mask in the "UnderwaterPostProcessHDRP" shader. initially i tried to use GATHER_RED_TEXTURE2D which did not work due to the fact that the texture is declared as TEXTURE2D_X. setting it to TEXTURE2D did not either so right now i just do 4 additional samples offsetting uvScreenSpace accordingly and accumulate the result in "maskSum". then i use:
mask = (maskSum > 2.0) ? 2.0 : mask; 
from my understanding this should set the mask to 2.0 if any of the 4 surrounding pixels has a vlaue larger than 2 or several have a value of 1...
but it does not fix the fireflies i see!
anything i am missing here or doing totally wrong?

daleeidd:
Fixes #739

Uses a compute shader to fill specks visible when underwater due to ocean surface precision issues and other imperfections producing incorrect mask values. Incorrect values can be incorrectly flipped VFACE and/or gaps in the ocean surface.

Samples both orthogonal and diagonal pixels separately. If the sampled pixels are all equal, then it writes that value to the current pixel. Sampling both orthogonal and diagonal is necessary for when a chain of incorrect values is in the mask which is possible.

grid-orthogonal grid-diagonal
S = current pixel. 1-4 is checked pixels.

On a 1660Ti at 4K, GPU performance cost was 0.31ms. This cost appears to be consistent.

This shader won't work on old macs as it doesn't like read and write to RWTexture2D for some reason (which is also why I think combine compute shader didn't work). Will need to test on newer mac soon.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Mask.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Mask.cs
@@ -13,6 +13,8 @@ namespace Crest
         const string k_ShaderPathOceanMask = "Hidden/Crest/Underwater/Ocean Mask";
         internal const int k_ShaderPassOceanSurfaceMask = 0;
         internal const int k_ShaderPassOceanHorizonMask = 1;
+        internal const string k_ComputeShaderFillMaskArtefacts = "CrestFillMaskArtefacts";
+        internal const string k_ComputeShaderKernelFillMaskArtefacts = "FillMaskArtefacts";
 
         public static readonly int sp_CrestOceanMaskTexture = Shader.PropertyToID("_CrestOceanMaskTexture");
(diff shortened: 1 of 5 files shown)


huwb:
Thanks for this. Its frustratingthat i was not able to solve the specks. I havent seen them in any other context so they should be solvable. I will see if i can take another identifying what is causing them, but i dont have a time line on that so its probably worth getting this in.

huwb (approved):
Oh might want release note

daleeidd:
Done. Thanks.

PR #923, opened 2021-10-09
Refactor: Move ScaleDifferencePow2 to LodTransform

magenta404:
Saves the calculation from per lod to only once.

The only difference in functionality is that if the LOD was disabled mid-game, and then becomes enable, the value will be different in this edge case. It can be solved by ignoring ScaleDifferencePow2 if LOD's first render. But I might be overthinking. Thoughts?


huwb:
Yeah seems like a good change.

Could we discuss that edge case more.. I wonder if it's really a problem since the lod data is removed when disabled I think.

PR #926, opened 2021-10-12
Refactor camera depth helpers

magenta404:
When I added the multisample depth helpers to fix MSAA outlines, it ended up being a bit of a mess which gave me some grief. This PR does the following:

- remove global variable dependency in includes file
- add texture macros from SRP
- use load instead of sample in underwater which simplifies a few things
- add multisample macros
- simplified shared RP code

I have kept most of the macros disabled as we should enable them as we use them.


daleeidd:
Thanks.

PR #927, opened 2021-10-13
LOD Data Buffering

magenta404:
Adds buffering to LOD data. I think I have this in a good place now.

It is buffering:
- Lod targets
- LodTransform RenderData
- OceanRenderer CascadeParams
- OceanRenderer CascadeInstanceData

Persistent data now uses the buffering system. For subsequent sub-steps it swaps between a single RT instead of buffering Lod sources.

It is still using the target/source setup for shaders as not sure if we need multiple previous frames available in a shader.

## Testing

Any persistent data can be used for testing.

I have added [test/mvs](https://github.com/wave-harmonic/crest/tree/test/mvs) for testing animated waves by hacking in MVs. Open the Test.unity scene to see MVs rendered to screen. It has a weirdness where I cannot use the _PreviousM matrix without getting glitches with horizontal movement. If you want to see the MV setup, it is another Ocean.shader pass. I had to set it to opaque for the pass to be triggered.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/LodData/LodTransform.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/LodTransform.cs
@@ -172,6 +172,7 @@ namespace Crest
             {
                 cascadeParams.Current[lodIdx]._posSnapped[0] = _renderData[lodIdx].Current._posSnapped[0];
                 cascadeParams.Current[lodIdx]._posSnapped[1] = _renderData[lodIdx].Current._posSnapped[2];
+                // NOTE: Current scale was assigned to current and previous frame, but not sure why. 2021.10.17
                 cascadeParams.Current[lodIdx]._scale = OceanRenderer.Instance.CalcLodScale(lodIdx);
                 cascadeParams.Current[lodIdx]._textureRes = _renderData[lodIdx].Current._textureRes;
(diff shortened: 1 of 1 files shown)


huwb:
Nice approach, good stuff.

Would it be a good idea to not double buffer the waves by default (or alternatively - to buffer by female but give option to disable), or is the assumption that buffering is always desirable enough to justify additional memory? The data is not small so just checking.

huwb:
> Persistent data now uses the buffering system

This is legit awesome - lovely stuff

daleeidd:
Thanks!
 
> Would it be a good idea to not double buffer the waves by default (or alternatively - to buffer by female but give option to disable), or is the assumption that buffering is always desirable enough to justify additional memory? The data is not small so just checking.

We should. I've disabled it and added the buffer count to the settings hash so it is ready to use. We can just change the BufferCount depending on whether MVs or VFX is used.

daleeidd:
I've also updated the BIRP test/mv branch. MVs are working correctly by the looks of it. Only some minor glitches at the extents when scale changes.

huwb:
> We can just change the BufferCount depending on whether MVs or VFX is used.

Oh so you mean change it on a branch?

I was wondering if it made sense to put an option in the anim wave settings, default to enable. But give with either.

But if it's hardcoded then probably does not need to be in settings hash?

Issue #710, opened 2020-12-18
Caustics showing up on inside of objects ie overhands, top and walls of underwater structures.  Probably should ignore downward facing normals etc.


holdingjason:
Here is a shot though of something interesting with caustics.  Not sure if there is anything to be done here and not a big deal but would look nice if we could avoid down facing normals for caustics perhaps.  See how we are getting them on all surfaces including the ceilings in this shot (I am not on latest btw)

> @Critter Cove We can use the depth and normals texture to accomplish this so it shouldn't be to difficult to add.

Dale 12/05/2020

![badcaustics](https://user-images.githubusercontent.com/42001313/102652158-7f938580-4122-11eb-9a1f-7d97327c501f.png)

holdingjason:
Also here is a video showing off the issue as well.

https://twitter.com/i/status/1325873079461965828

daleeidd:
Thanks. Another option that we should consider is sampling the shadow maps.

holdingjason:
> Thanks. Another option that we should consider is sampling the shadow maps.

Yeah I could see that work as well.  Though it does not need to be perfect so I would go with the most performant option available unless its a big big trade off.  Thanks for taking a look.  Just looks bizarre right now inside of stuff underwater but not a show stopper by any means.

huwb:
Was hoping this would be straightforward but we don't have a normal for the refracted scene.

We could reconstruct one by sampling the depth buffer multiple times don't this reliably across depth continuities etc is a topic in itself, and will cost perf.

Using the shadow maps seems like the best option. I'm guessing this would then be a URP/hdrp only fix.

holdingjason:
Any way to just use a fixed world vector ie down up to a certain angle?

Arugin:
Just fyi, you have acces to the normal buffer at least in HDRP.

daleeidd:
URP 10 has a normal buffer.

Shadows will be a more comprehensive though.

daleeidd:
#929 adds this for built-in. Will also be attempting other RPs for 4.14.

daleeidd:
This feature has landed with 4.14.

PR #929, opened 2021-10-20
Use shadow maps instead of shadow data for caustics

magenta404:
Solves #710

Rather than using the shadow data, it now uses shadow maps.

The shadow data has the following issues:
- doesn't handle shadows casted underwater
- has same (or similar) bug as #914
- caustics didn't match shadows 100% due to moving with waves

Using shadow maps solves all of these cases.

I checked performance but didn't see noticeable difference. It was hard to measure as it fluctuates - even in a frozen scene. But lowest and highest was around 3.05ms and 3.95ms respectively for the for both for the ocean shader.

I took the code from Unity's screen-space shadow shader. ~~Sampling a single cascade seemed good enough~~ (Nope) and using split spheres was necessary or shadows would be incorrect at edges.

### DIFF
--- a/docs/about/history.rst
+++ b/docs/about/history.rst
@@ -28,6 +28,7 @@ Changed
    -  All water features updated to support varying water level.
    -  Add buttons to *Spline* inspector to quickly enable water features.
    -  Exposed control over *Spline* ribbon alignment - spline points now define the center of the ribbon by default.
+   -  Caustics no longer render in shadows casted from objects underwater.
 
 Fixed
 ^^^^^


huwb:
Here are notes about my previous struggles. They can be ignored if this is working though, but just adding for record https://github.com/wave-harmonic/crest/issues/25#issuecomment-415680976

daleeidd:
Thanks for the info. It lead me to changing the sampling to the screen-space shadow texture instead.

> I think back in the day I struggled because I could not get access to shadow maps during transparent rendering

A [forum post](https://forum.unity.com/threads/no-shadows-visible-on-transparency-shaders.9909/page-3#post-4778408) from a thread linked in your linked comment pointed me to a better solution. Unity no longer clears the data so maybe that happened after your initial attempts. So I instead now sample from the screen-space shadow texture which is cheaper.

The SS shadow texture is only available on desktops and consoles, but a [post](https://forum.unity.com/threads/what-is-unity_no_screenspace_shadows-good-for.298117/) which quotes the Unity release notes and might only be only an issue for OpenGL only. Not sure. We could run with this and wait for support requests and see.

> I assume it knows to use the sun shadow maps by default rather than other light shadow maps

I checked and it is only for the sun.

daleeidd:
Thanks.

PR #931, opened 2021-10-21
Add new Example scene

daleeidd:
I think this is ready for review.

Adds a new example scene which contains mini scenes contained to a sphere. The idea is that it will be easy to showcase various features and their usage. Already, this scene could replace the whirlpool scene and one of the boats scene (the other could be cleaned from certain examples).

Examples

The theme is minimalism. Not completely happy with the river example but probably good enough for now. It is missing the underwater part of the sphere but that can come when it is ready.

Also noteworthy is that I have added AutoMoveAndRotate from Unity Standard Assets. It is quite handy. We could make our own though. I placed it under Crest.ThirdParty namespace to avoid conflicts.

The content is pretty robust in fitting in frame except with really small window sizes.


daleeidd:
Pinging @huwb. Whenever you're ready. Can be pushed back for 5.0 or can wait until you have more time.

huwb:
Thanks for bumping, i will look at it soon

huwb:
It looks great! Lots of small tweaks we can make to each example but i can propose those later.

For now what are your thoughts on the presentation/scene? Is it intended to use the underwater volume stuff #960 ?

I was thinking we could take inspiration from this old ref: https://twitter.com/qornflex/status/1195701316418490368?s=20 . a lot of the details are not immediately do-able for this PR, but it seems like a good aspiration and i feel with vfx graph and proper volume lighting and soem other tweaks we should be able to approach this look at some point.

daleeidd:
Thanks.

> For now what are your thoughts on the presentation/scene?

Pretty much complete. I was thinking of changing the background to a solid colour but I don't mind the earthy brown.

> Is it intended to use the underwater volume stuff #960 ?

It was. I would need to make some additions to it. Looks strange at the moment because the volume uses the back faces for the fog, but for this scene full depth looks better. Also the angle of these scene doesn't really work with the volumes as you only see a sliver of the volume. I think I will skip the volume for this scene and look at adding it if there is an example that needs it. It can be used for the fancy sphere scene though.
 
> I was thinking we could take inspiration from this old ref: https://twitter.com/qornflex/status/1195701316418490368?s=20 . a lot of the details are not immediately do-able for this PR, but it seems like a good aspiration and i feel with vfx graph and proper volume lighting and soem other tweaks we should be able to approach this look at some point.

Looks like a good target. I might be able to use prefab variants so we can have a separate scene for more fancy ones like that one.

huwb:
>  I was thinking of changing the background to a solid colour but I don't mind the earthy brown.

I think I'd vote for trying this as well. I find the sky part of the skybox distracting. I'd vote for something like in that ref that pushes it back and makes the water pop. However maybe making it solid colour messes with the water colour after lighting is baked? Perhaps something to look out for.

Actually I have an idea to propose - to make the background a gradient. Will try to do a quick test to propose something later tonight.

> I would need to make some additions to it.

Just to understand, what is required do you think? I naively thought "portal" mode would do it but I'm not sure. Would be great if it can be done as I think seeing the water volume is compelling having only the surface breaks the illusion of the scene I feel. And if it can be made to work then it can be used directly for the "fancy" version of the scene.

huwb:
proposal to for background, that we may use as a starting point for the orb scene. it is a shader applied to sphere geometry that does not impact the skybox nor does it affect the generated lighting. let me know what you think.

before:
image

after:
image

daleeidd:
Looks much better.

huwb:
Ok cool i've added it. It's just dropped into the scene. If you're happy with it could you do the honors and add it to the prefab?

Also i put in a proposed change to lower the opacity of the UI to make it less 100% white. i think it helps a bit to make the ui softer. but let me know if you think its worth keeping. its left as prefab overrides so easy to revert or apply.

daleeidd:
I've added a new scene for underwater lighting (out-scattering) using the naval mine. In the process I created a mask for geometry which reuses the underwater volumes. It also fills the underside of the geometry to create sphere underneath. I've added it to the river example.

River

After another PR merges I will add a particle example with the ocean inputs and that will be it I think.

> Ok cool i've added it. It's just dropped into the scene. If you're happy with it could you do the honors and add it to the prefab?

Done. Thanks.

> Also i put in a proposed change to lower the opacity of the UI to make it less 100% white. i think it helps a bit to make the ui softer. but let me know if you think its worth keeping. its left as prefab overrides so easy to revert or apply.

Looks good. Thanks.

huwb:
Cool will try it out soon

huwb:
I ran the scene and clicked through to the river example and seem to get a null ref thrown each frame in the mask code:

ArgumentNullException: Value cannot be null.
Parameter name: mesh
UnityEngine.Rendering.CommandBuffer.DrawMesh (UnityEngine.Mesh mesh, UnityEngine.Matrix4x4 matrix, UnityEngine.Material material, System.Int32 submeshIndex, System.Int32 shaderPass, UnityEngine.MaterialPropertyBlock properties) (at <be28439f7b014fc38435d198c0640ff7>:0)
Crest.Examples.MaskFill.ExecuteFillMask (UnityEngine.Rendering.CommandBuffer buffer, UnityEngine.Camera camera, UnityEngine.RenderTextureDescriptor descriptor) (at Assets/Crest/Crest-Examples/Examples/Scripts/MaskFill.cs:68)
UnityEngine.Events.InvokableCall`3[T1,T2,T3].Invoke (T1 args0, T2 args1, T3 args2) (at <be28439f7b014fc38435d198c0640ff7>:0)
UnityEngine.Events.UnityEvent`3[T0,T1,T2].Invoke (T0 arg0, T1 arg1, T2 arg2) (at <be28439f7b014fc38435d198c0640ff7>:0)
Crest.Examples.CustomPassForCamera.Execute (UnityEngine.Rendering.CommandBuffer buffer, UnityEngine.Camera camera, UnityEngine.RenderTextureDescriptor descriptor) (at Assets/Crest/Crest-Examples/Examples/Scripts/CustomPassForCamera.cs:104)
Crest.Examples.CustomPassForCameraBase.OnBeforeRender (UnityEngine.Camera camera) (at Assets/Crest/Crest-Examples/Examples/Scripts/CustomPassForCamera.cs:86)
UnityEngine.Camera.FireOnPreRender (UnityEngine.Camera cam) (at <be28439f7b014fc38435d198c0640ff7>:0)

huwb:
I think its looking really good at a broad level so ill list some micro questions / feedback. I can help with these if you want to send any my way:

* Decals - i'd vote for disabling this in BIRP - render alpha on surface is broken and i dont want to put it forward as people may not realise its broken at first
* Whirlpool - looks cool. I might do an iter on it to make it a bit larger and slow it down a bit to give it a feeling of weight/momentum
* Wake - already looks cool but will benefit from updating SWI. Happy to help tweak once this is up to date. Foam would probably help too.
* WaveParticle - mixed feelings about including this. It's one of those things that is cool to me but i dont think its ever been used ever. If you feel we should keep it, perhaps making a few of them moving in random directions or animating in some way rather than a big static bump is better. Also if we keep it we should probably push it backwards in the order if that is easy to change and promote ones we know people use.
* Spinner - should probably rename to something that represents the general feature, perhaps "Complex Object Water Interaction" if thats not too much of a mouthful
* Push Water Below Sphere - typo in title, and also should represent general feature - Push Water Below Geometry perhaps
* Waves Patch - This could probably be exaggerated more to make it clear what is going on. The normal maps make it look like there are waves everywhere which i'm not enamoured with these days - note about that below. Perhaps the waves in the patch should be dialled up a bit? And they exhibit obvious patterns from gerstners, FFT might be a better choice here unless we wanted gersties.
* Bumps - I'd vote for cutting sadly, these are a cool coder feature but i've never seen a use case for them. They also suffer from quality issues at the silhouettes
* Flow & Depth - wonder if this should just be "Flow" as im not sure "Depth" will be clear. But maybe it does no harm and this can be ignored.
* Scales Waves - another case which normal maps are confusing. Disabling normals makes it look much clearer, I'd probably vote for that.
* Dyn Waves Refls - This is cool! It occurs to me that we could have made a circular tank for the entire orb which contained the water. But that would have less square reflections and would be a lot larger and might negatively effect it so happy to go with that.
* "Foam Override" - wonder if this should be "Custom Foam Input". Override to me implies replacing the data.
* "Shadow Override" - "Custom Shadow Input" . A realistic use case of this is to cover an area with shadow, for when CSMs fail. I'm not sure if theres a more representative demo we could do. But if not this works.
* Trail Renderer - This is cool. It's doing a somewhat similar job to the "Foam Override" demo in its current form. Not sure if theres something to do to differentiate it. Maybe animating a figure-8 make it look more trail-y, or making it grow wider along the trail or something. But if not this works.

Normals note - I generally prefer to lean heavily on FFT to give the normals and either not use normal maps, or use them very slightly if needed.  This way it reflects the true spectrum. And its slightly cheaper to render as a bonus - although resolution usually needs to be cranked up to 384 or more to get enough detail. On my next pass of example scenes i'll likely propose to turn them off in the other water materials.

An orbit cam would be ideal but i dont know if we should bother, just registering the thought even if just for the future.

Tempted to forego the animated progress bar for examples which don't have a progression. It creates the illusion that something is changing and i spent a bit of time watching it and the scene and trying to figure out what it signified. On the other hand it works great for the underwater lighting. Not sure what you think?

huwb (commented):
Commented above

daleeidd:
Thanks for great feedback. I've added a gerstner vs FFT comparison. Also added a particle input example using albedo. Not the best example but gets the idea across I suppose.

> ArgumentNullException: Value cannot be null.

Lucky you caught that. I was using a blend file but requires Blender to be installed. Converted to FBX and threw blend file into sources.

> * Decals - i'd vote for disabling this in BIRP - render alpha on surface is broken and i dont want to put it forward as people may not realise its broken at first

Removed.

> * Wake - already looks cool but will benefit from updating SWI. Happy to help tweak once this is up to date. Foam would probably help too.

Brought branch up-to-date so latest wakes are ready to do.

> * WaveParticle - mixed feelings about including this. It's one of those things that is cool to me but i dont think its ever been used ever. If you feel we should keep it, perhaps making a few of them moving in random directions or animating in some way rather than a big static bump is better. Also if we keep it we should probably push it backwards in the order if that is easy to change and promote ones we know people use.

Added the animated motion. Looks much more interesting now. Moved further down the list.

> * Spinner - should probably rename to something that represents the general feature, perhaps "Complex Object Water Interaction" if thats not too much of a mouthful

Done.

> * Push Water Below Sphere - typo in title, and also should represent general feature - Push Water Below Geometry perhaps

Done.

> * Waves Patch - This could probably be exaggerated more to make it clear what is going on. The normal maps make it look like there are waves everywhere which i'm not enamoured with these days - note about that below. Perhaps the waves in the patch should be dialled up a bit? And they exhibit obvious patterns from gerstners, FFT might be a better choice here unless we wanted gersties.

Removed normals. Switched to FFT and spline. There is some weirdness at one of the corners but going to leave it for now.

> * Bumps - I'd vote for cutting sadly, these are a cool coder feature but i've never seen a use case for them. They also suffer from quality issues at the silhouettes

Removed.

> * Flow & Depth - wonder if this should just be "Flow" as im not sure "Depth" will be clear. But maybe it does no harm and this can be ignored.

Done.

> * Scales Waves - another case which normal maps are confusing. Disabling normals makes it look much clearer, I'd probably vote for that.

Done.

> * Dyn Waves Refls - This is cool! It occurs to me that we could have made a circular tank for the entire orb which contained the water. But that would have less square reflections and would be a lot larger and might negatively effect it so happy to go with that.

Now that I have masking tools I made the shell of the pool cover the entire sphere and added some other things:



> * "Foam Override" - wonder if this should be "Custom Foam Input". Override to me implies replacing the data.

I updated the example to exhibit the override behaviour. Almost looks like a Yin and yang.

> * "Shadow Override" - "Custom Shadow Input" . A realistic use case of this is to cover an area with shadow, for when CSMs fail. I'm not sure if theres a more representative demo we could do. But if not this works.

Ditto. I don't think a proper use case would be immediately clear to someone viewing it. Makes more sense to show raw potential I think.

> * Trail Renderer - This is cool. It's doing a somewhat similar job to the "Foam Override" demo in its current form. Not sure if theres something to do to differentiate it. Maybe animating a figure-8 make it look more trail-y, or making it grow wider along the trail or something. But if not this works.

I was thinking the same thing, but I don't think I will get around to it this time round. Feel free to take a stab if you want.

> Normals note - I generally prefer to lean heavily on FFT to give the normals and either not use normal maps, or use them very slightly if needed. This way it reflects the true spectrum. And its slightly cheaper to render as a bonus - although resolution usually needs to be cranked up to 384 or more to get enough detail. On my next pass of example scenes i'll likely propose to turn them off in the other water materials.

Yeah, not a fan of using normal maps myself. Removed normal maps from all examples. I also added a viewpoint to the centre of the sphere to improve detail.

> An orbit cam would be ideal but i dont know if we should bother, just registering the thought even if just for the future.

Sounds good. Will skip for this round. Taken note.

> Tempted to forego the animated progress bar for examples which don't have a progression. It creates the illusion that something is changing and i spent a bit of time watching it and the scene and trying to figure out what it signified. On the other hand it works great for the underwater lighting. Not sure what you think?

That's fair. To me it is just recommended viewing time. Most examples is set to 1 second. But I can see why it could be distracting. Done. Removed for anything that has no progression or no warmup period.

daleeidd:
Also I set the gravity for dynamic waves to 1 and tried to tweak them but I didn't do a great job of it. Feel free to tweak those.

> Whirlpool - looks cool. I might do an iter on it to make it a bit larger and slow it down a bit to give it a feeling of weight/momentum

Sounds good if you want to tackle it.

huwb:
Nice!

Will respond fully later but couple of things i noticed while retesting:

* I think at some point when flicking through the examples the ocean material changes, but flicking back the ocean material doesnt change back
* At some point camera controls are locked, and then seem to be locked forever
* Re gravity==1 i'm tempted to go back to 'cheating' - using a higher value. At the scale of the example scene, reducing the gravity multiplier has taken the look back a fair bit i feel. Not sure what you think.

The yin yang does indeed look cool! I'm enabling the "follow horiz motion" option on the half-foam input to stop it jittering.

I'm going to put forward a tweaked ocean material that is not dominated by SSS. Slightly less spicy perhaps but more realistic i feel.

huwb:
Theres something wrong with the whirlpool. I reduced the "max speed" param and it stops the rotation but still moves really fast towards the center? This feature needs some proper TLC. I'll aim to look at it soon but would need to test in PC and etc. Probably not worth holding this PR up for it.

huwb:
I attempted to tweak wakes. I re-enabled waves as it looks much juicier with them, hope thats ok

I added a sphere as it helps ground it a lot i find, and the appearance of the water within the sphere is a little odd so it hides that. The sphere is quite large, maybe can come down but i was struggling to get a good result.

The main struggle was making it strong enough to get foam without self intersecting water which creates very obvious underwater through the sphere. I could reduce horiz displacement from the sim but that will kill any remaining foam. Is there a mechanism to increase foam coverage per-example?

huwb:
Re animated progress bar, yeah i think thats better, nice one. One last though on that is that the underwater lighting sequence is quite long - at the beginning its not clear anything is happening and i feel like users will think it refers to the underwater colour and click past it. Not sure if briskening this is an option. If not it works as is.

huwb:
Everything else is looking awesome. Close now!

daleeidd:
> I think at some point when flicking through the examples the ocean material changes, but flicking back the ocean material doesnt change back

Fixed.

> At some point camera controls are locked, and then seem to be locked forever

Fixed.

> Re gravity==1 i'm tempted to go back to 'cheating' - using a higher value. At the scale of the example scene, reducing the gravity multiplier has taken the look back a fair bit i feel. Not sure what you think. EDIT: except the wake! im happy with the gravity multiplier for the wake example... not sure if it can be varied per scene. similar question to modifying the foam coverage per example i guess.

> The main struggle was making it strong enough to get foam without self intersecting water which creates very obvious underwater through the sphere. I could reduce horiz displacement from the sim but that will kill any remaining foam. Is there a mechanism to increase foam coverage per-example? (having said that its possible that increasing foam coverage across the board may also work)

For something to be compatible with UnityEvents, they need to be something invokable (method or property) and public. So we would need to add a public setter to the sim property and then you could assign a sim per example.

Using UnityEvents has given Crest good exposure to them but if I had to do this again I would probably use prefabs. But then again we would have 20 or so extra prefabs.

> The yin yang does indeed look cool! I'm enabling the "follow horiz motion" option on the half-foam input to stop it jittering.

Thanks.

> I'm going to put forward a tweaked ocean material that is not dominated by SSS. Slightly less spicy perhaps but more realistic i feel.

I vote for spicy :^) Looked interesting and something different from other scenes. I think the sea green suited the volume more and better contrast to the black background. Up to you though.

> A fun issue is that the wake example completely exited the background sphere after a short time. I've increased the sphere radius substantially so it at least won't happen quickly.

I moved the background to under the camera transform and decreased the size again which should solve the problem.

> (That makes me wonder if there is a compelling floating origin example we could have in the future.. will think about it. i guess having a clearly marked "world origin" that is outside the floating origin component, so that you can see how it stays close to the origin)

That could be good!

> One last though on that is that the underwater lighting sequence is quite long - at the beginning its not clear anything is happening and i feel like users will think it refers to the underwater colour and click past it. Not sure if briskening this is an option. If not it works as is.

Yeah I was thinking that too but couldn't find a way to apply an animation curve to the time to make it smooth.

huwb:
Sorry for the long delay. Coming back to it with fresh eyes i was struck by how cool it is!

Regarding the water colour, sure we can make it more green, but could it be factored into the Diffuse water colour - make this more green. The sun scattering was intended to be a change of colour from the sun light coming directly through water, usually only in peaks etc. Unfortunately historically i've massively over used it and it impacts the entire water colour everywhere which is not what we should put forward. To get to the point, do you think you can get the spicyness back by tinting the diffuse colour green? I can also take a stab but i think its worth you trying.

A minor new bit of feedback for new scene "Transparent Object With Underwater Fog". Firstly could maybe be "Transparent Object Underwater". But potentially more importantly its not clear to me that this is a transparent object, so my brain does not really grok it. Could the alpha be reduced to make this clear? If the sphere becomes invisible underwater, perhaps moving it towards the camera helps if having it off-center is not annoying.

For Albedo, I think putting "(Preview)" in the name would be useful, for the relevant 'RPs

Regarding gravity multiplier, cool i believe i've done now that. I havent applied to prefabs in case i'm doing it wrong. it would be slightly more professional if it was `SetGravityMultiplierOverride`/`ClearGravityMultiplierOverride` which sets a transient override for the authored value rather than changing the authored value. however OnEnable/OnDisable events seem robust and i think it does the job.

The final thing i notice is the camera is quite elevated and looking down steeply by deafult. And only a slither of the underwater is visible. I think we could tweak the layout here. Also fresnel does not show up, so for example the ripple/reflections scenes dont really work. Could the camera be dropped a bit do you think?

Ok i'm semi sure that will be my last feedback! Again its overall looking great! Let me know if i can jump on any of the above.

huwb (commented):
Added last round of comments

daleeidd:
Thanks! Everything makes sense and will make those changes.

> Regarding gravity multiplier, cool i believe i've done now that. I havent applied to prefabs in case i'm doing it wrong. it would be slightly more professional if it was `SetGravityMultiplierOverride`/`ClearGravityMultiplierOverride` which sets a transient override for the authored value rather than changing the authored value. however OnEnable/OnDisable events seem robust and i think it does the job.

I was thinking of having a public property on the ocean renderer for the dynamic waves settings file and just allow us to replace the entire file. I can do this but let me know what you think.

```csharp
public SimSettingsWave SimSettingsDynamicWaves { get => _simSettingsDynamicWaves; set => _simSettingsDynamicWaves = value; }
```

_simSettingsDynamicWaves is already public.

huwb:
Oh yeah I guess that is better! Yes if you don't mind doing that please do

daleeidd:
That should be all done. Thanks for the feedback.

Was somewhat torn on the camera placement as the original placement was better in some cases as it had a better vantage point for the surface where most examples happen. But certainly needed to be changed to address your points.

I have improved the UI and added the preview thing. I added material icons which has an Apache license which I think should be okay. I added the license in the subfolder.

I went through and polished a few examples.

Also redid the "varying water height" example for the new camera placement. It is kind of a dam. I tried making a nice damn but I would need to improve the sphere mask thingy for that design.

huwb:
Nice!

Going through it.

Re the license, will that need to be passed on to users? I assume not but just checking. If it is we might look for something less license-y.

daleeidd:
Thanks. And thank you for improvements/feedback :^)

PR #933, opened 2021-10-22
Replace UpdateShadow.compute with UpdateShadow.shader

magenta404:
Replaces shadow compute with shader.

Summary:
- Improve code sharing with downstream
- Reduce code diff with downstream (C#)
- No more deprecated code
- Fix shadows for MacOS
- Fix "Shadow Projection > Close Fit" mode
- Better performance (old: 0.19ms vs new: 0.10ms on 1660Ti)

This means built-in is more aligned with downstream. Code sharing has greatly improved and code diff reduced. And no longer relying on deprecated code. Also, the compute shader was broken on MacOS (old and new M1); it may have worked with a dGPU or eGPU (cannot remember) but certainly not on iGPUs.

Broken shadows on MacOS.

TODO:
- [x] Test on Windows
- [x] Compare performance
- [x] Add release notes

Not sure if you wanted to keep using the compute shader and make SRP a compute shader. Let me know what you think.

### DIFF
--- a/docs/about/history.rst
+++ b/docs/about/history.rst
@@ -42,6 +42,11 @@ Fixed
    -  Fix *ShapeFFT* component producing inverted looking waves when enabled in editor play mode.
    -  Fix SSS colour missing or popping in the distance.
 
+   .. only:: birp
+
+      -  Fix shadows for MacOS. `[BIRP]`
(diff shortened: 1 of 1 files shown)


daleeidd:
Thanks.

Updated the description. Found out this also fixed closed fit shadows. And almost halved performance cost.

PR #938, opened 2021-11-12
Fix depth cache incorrectly importing 4.14 baked depth caches

daleeidd:
Negative values were being clamped due to compression which I have disabled. I also disabled NPOT Scale which slightly changed values on import.


huwb (approved):
Nice one

daleeidd:
Thanks.

PR #941, opened 2021-11-18
Underwater fog multiplier

daleeidd:
It very often requested to reduce the depth fog only for the underwater effect. This adds a 0 (0.01) to 1 factor for the underwater fog. I don't think we need to go above 1.


daleeidd:
Converting to draft as got to fix a few things.

daleeidd:
@huwb ready for review.

huwb (approved):
Yeah looks good nice one

daleeidd:
Thanks.

PR #942, opened 2021-11-19
Fix underwater specks for Intel iGPUs

daleeidd:
Report on:

_Dell Inspiron 3537_
processor: CPU: Intel (R) Core (TM) i7-4500U CPU @ 1.80GHz [4 cores]
graphics_memory: VRAM: 2216MB. Max texture size: 16384px. Shader level: 50
Version: Direct3D 11.0 [level 11.0]
Renderer: Intel (R) HD Graphics Family VRAM: 2216 MB

_MacBook Pro_ (mine)
Intel iGPU

It looks like UAV Read/Write is patchy which I believe is also the problem with the combine compute shader. There isn't a great way to detect this as I thought `SystemInfo.IsFormatSupported(GraphicsFormat, FormatUsage.LoadStore)` is the correct check but it passes in this case. Going by the _Graphics_ repository there are specific checks for Intel so this is in line with Unity. I have reported it to Unity a while ago and see what they say.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Helpers/Helpers.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Helpers/Helpers.cs
@@ -23,6 +23,13 @@ namespace Crest
             return false;
         }
 
+        public static bool IsIntelGPU()
+        {
+            // Works for Windows and MacOS. Grabbed from Unity Graphics repository:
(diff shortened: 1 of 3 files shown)


huwb (approved):
nice one

daleeidd:
Thanks.

PR #943, opened 2021-11-21
Underwater Shader API

daleeidd:
_**Status:** Needs some work but ready for early feedback_

Adds an "includes" file for developers to add the underwater fog to their transparent shaders.

When the "Enable Shader API" option is enabled on the _Underwater Renderer_, it will:
- render the underwater effect before the transparent pass
- enable underwater rendering on ocean shader (like how it was before the underwater PP)
- populate global shader data (uses three global keywords)

Second point is unfortunate as it was nice not having underwater rendering in the ocean shader, but this allows users to keep their transparent objects rendered once in the transparent pass.

**Todo:**

- [ ] Documentation
- [ ] Remove test data
- [ ] Add _Crest_ prefix to more shader things

The last one is a bit of a hassle now that most of the code is exposed. We can probably get away with not adding a Crest prefix to a few things like _LD_TexArray_Shadow. Not sure about CascadeParams though. Might be worth doing as a separate PR.

**Testing**

_Test.unity_ under _Test_ directory has it setup. I have included three shaders as examples of implementation for BIRP. ShaderGraph nodes will need to be added downstream.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Effect.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Effect.cs
@@ -15,15 +15,30 @@ namespace Crest
         internal const string k_KeywordDebugViewOceanMask = "_DEBUG_VIEW_OCEAN_MASK";
         internal const string k_KeywordDebugViewStencil = "_DEBUG_VIEW_STENCIL";
 
-        internal static readonly int sp_CrestCameraColorTexture = Shader.PropertyToID("_CrestCameraColorTexture");
-        internal static readonly int sp_CrestWaterVolumeStencil = Shader.PropertyToID("_CrestWaterVolumeStencil");
-        static readonly int sp_InvViewProjection = Shader.PropertyToID("_InvViewProjection");
-        static readonly int sp_InvViewProjectionRight = Shader.PropertyToID("_InvViewProjectionRight");
-        static readonly int sp_AmbientLighting = Shader.PropertyToID("_AmbientLighting");
-        static readonly int sp_HorizonNormal = Shader.PropertyToID("_HorizonNormal");
-        static readonly int sp_DataSliceOffset = Shader.PropertyToID("_DataSliceOffset");
-        static readonly int sp_LightColor0 = Shader.PropertyToID("_LightColor0");
-        static readonly int sp_WorldSpaceLightPos0 = Shader.PropertyToID("_WorldSpaceLightPos0");
+        public static partial class ShaderIDs
+        {
+            // Local
+            public static readonly int s_HorizonNormal = Shader.PropertyToID("_HorizonNormal");
+
+            // Global
+            public static readonly int s_CrestCameraColorTexture = Shader.PropertyToID("_CrestCameraColorTexture");
+            public static readonly int s_CrestWaterVolumeStencil = Shader.PropertyToID("_CrestWaterVolumeStencil");
+            public static readonly int s_CrestAmbientLighting = Shader.PropertyToID("_CrestAmbientLighting");
+            public static readonly int s_CrestDataSliceOffset = Shader.PropertyToID("_CrestDataSliceOffset");
+            public static readonly int s_CrestDepthFogDensity = Shader.PropertyToID("_CrestDepthFogDensity");
+            public static readonly int s_CrestDiffuse = Shader.PropertyToID("_CrestDiffuse");
+            public static readonly int s_CrestDiffuseGrazing = Shader.PropertyToID("_CrestDiffuseGrazing");
+            public static readonly int s_CrestDiffuseShadow = Shader.PropertyToID("_CrestDiffuseShadow");
+            public static readonly int s_CrestSubSurfaceColour = Shader.PropertyToID("_CrestSubSurfaceColour");
+            public static readonly int s_CrestSubSurfaceSun = Shader.PropertyToID("_CrestSubSurfaceSun");
+            public static readonly int s_CrestSubSurfaceBase = Shader.PropertyToID("_CrestSubSurfaceBase");
+            public static readonly int s_CrestSubSurfaceSunFallOff = Shader.PropertyToID("_CrestSubSurfaceSunFallOff");
+
+            // Built-ins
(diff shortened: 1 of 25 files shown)


huwb:
> enable underwater rendering on ocean shader (like how it was before the underwater PP)

oh is this so transparents can blend over the underside of the water surface, because the underside will have fog applied?

ok i guess this is all good as long as its not adding significant complexity/expense

daleeidd:
> oh is this so transparents can blend over the underside of the water surface, because the underside will have fog applied? 

Yes.

> will transparents blend correctly over the background (skybox)? and over opaque geometry?

Yes. The fog is applied to everything before the transparent pass... But also no as it doesn't take the transparent object thickness into account. So a large transparent cube (air volume) will have more fog behind it than it should. But solving that is beyond scope.
 
> ok i guess this is all good as long as its not adding significant complexity/expense

I have to bring it up-to-date and check complexity again. And will make sure there is no extra perf cost when not enabled.

huwb:
Makes sense

daleeidd:
Alrighty... I believe this is finished. There is still test data present under _Test_.

As for adding complexity to the ocean shader, it doesn't really add much. We just need to render the fog and provide the underwater lighting (ambient and shadows). I guess it is a matter of what features come in the future and whether or not this approach can support it. It currently supports all of the underwater features including culling.

I got rendering before transparent pass working downstream. I haven't looked at adding rendering the fog in shader graph yet but it should be workable. I'll do that next before we consider merging.

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Effect.cs
```diff
@@ -245,6 +243,29 @@ internal void ExecuteEffect(CommandBuffer buffer, Material material, MaterialPro
             }
         }
 
+        static void UpdateGlobals(Material oceanMaterial)
+        {
+            {
+                // Unity is not setting the sun correctly both in scene view and before transparent pass. Use most
+                // likely sun candidate.
+                Shader.SetGlobalVector("_CrestWorldSpaceLightPos0", -RenderSettings.sun.transform.forward);
```
huwb, line 251: If this is run each frame it should probably use param IDs rather than strings

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Effect.cs
```diff
@@ -291,6 +313,17 @@ internal void ExecuteEffect(CommandBuffer buffer, Material material, MaterialPro
...
+
+                    if (setGlobalShaderData)
+                    {
+                        UpdateGlobals(material);
+                    }
+                }
+
+                if (setGlobalShaderData)
+                {
+                    Shader.SetGlobalVector("_CrestDepthFogDensity", dominantWaterBody == null
```
huwb, line 325: Ditto re Param ID

huwb (approved):
Reviewed as well as I could, only a few minor things.

Mostly interested now in documentation/examples to make sure this stuff is discoverable and usable for people. Will that come in this pr or later?

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Effect.cs
daleeidd: Done. I've added them using an approach I borrowed from Unity by placing them in a static class called ShaderIDs. LMKWYT.

daleeidd:
> Mostly interested now in documentation/examples to make sure this stuff is discoverable and usable for people. Will that come in this pr or later?

Partly in this PR and partly later as will be easier documentation the SG once I have done it. I have a few examples in the Test folder. One we should keep is the surface shader example as this is likely what people will use for built-in. For SRP it will be shader graph.

I'm thinking of adding an example to the new Examples scene. So I should look at merging that one first.

daleeidd:
I looked at SG and just realised this approach won't work. SG doesn't have a "finalcolor" option or anything similar so the underwater effect will be overwritten by volumetric fog for the ocean back faces.

But this is the correct approach. It is just a matter of Unity bringing SG up to parity with surface shaders (SS). So I am thinking of keeping this for built-in and adding an alternative for SG or not support SG for the time being.

The alternative approach would be rendering the transparent object twice using the "Draw Renderers" feature. All I would have to do in theory is add a node for people to use and they would have to make a separate unlit SG and add the node. They'll render again with DR for their normal material, and render again with DR for unlit SG to apply fog. Still wouldn't be perfect like SS approach but best I can think of.

huwb:
I'm not sure if finalcolor can be a thing in the SG world.

The best I can think of would be to have a custom master node that would add the additional render pass automatically. Not sure if/when custom master nodes will land though, nor whether they'd allow this.

The way if probably approach this is give up trying to use unitys lighting, rather just make an unlit graph that potentially does it's own simple lighting, then does fog calculationb before returning the colour. Given this is only for underwater transparents, if hope that we could get away with not supporting the full lighting model. And since it's SG, advanced users will be able to modify it as required. Would this work?

daleeidd:
> I'm not sure if finalcolor can be a thing in the SG world.
> 
> The best I can think of would be to have a custom master node that would add the additional render pass automatically. Not sure if/when custom master nodes will land though, nor whether they'd allow this.

After reading a [thread discussing new SSs](https://forum.unity.com/threads/focusing-on-shaders.1171444/), it looks like SG won't get it and instead a new SS system might have it once released instead.

> The way I'd probably approach this is give up trying to use unity's lighting, rather just make an unlit graph that potentially does it's own simple lighting, then does fog calculation before returning the colour. Given this is only for underwater transparents, i'd hope that we could get away with not supporting the full lighting model. And since it's SG, advanced users will be able to modify it as required. Would this work?

That would work for anything completely submerged. We might be able to go with a lit graph if we disable the fog for the material (transparency has a fog toggle). Having the HDRP lighting over the top might be okay.

I guess I can cross the bridge once I am there as compatibility with SRP can be done downstream.

huwb:
Makes sense about fully submerged. In any case I'd propose keeping things simple and evaluating requests to extend/improve as they come. But yeah see how you get on. Sounds good.

daleeidd:
Thanks. I have dropped test data and documentation for now as will be easier to write once I know how SRP will work with it exactly.

PR #945, opened 2021-11-23
Use displaced position when sampling shadow maps

daleeidd:
Uses the displaced position in UpdateShadow to sample the shadow maps to improve shoreline (and elsewhere) shadowing. The improvement is less noticeable in HDRP as we don't use the shadow data for specular shadows. But improves SSS shadowing for every RP.


huwb (approved):
Oh cool id noticed this but didn't realise this was the fix. Nice

daleeidd:
Thanks.

Issue #726, opened 2021-01-15
Making Depth Cache Attenuation Developer Friendly


GXMark:
Currently to make waves attenuate correctly a terrain mesh needs to be approximately 500m - 1000m before the sea level as per advised in the video https://www.youtube.com/watch?v=jcmqUlboTUk

This is not a very practical approach for terrain construction and remains a chore for the developer to have to design or take an existing terrain and adjust it for these depth heights.

What is required is a workflow as follows to accomplish something much more dev friendly 👍 

1.  Create Depth Cache Game Object 
2.  Adjust it to average height of the sea level ( e.g. 50m )
3.  Set Maximum Attenuation Depth Property on the Ocean Depth Cache component   e.g. ( 50m )
4.  Note that this is 50m lower than the sea level ( e.g. 50m ) which means the calculated world coordinate depth is 0m
     ( This will ensure that the maximum amplitude for a wave will occur at the maximum Attenuation Depth )

This workflow will ensure that existing terrains are easy compatible with the crest ocean system.

huwb:
a couple of ideas:

* depth cache input could have a option to fade down to max depth at the sides. so you could specify eg 25m, and over that distance the depths lerp down to 'deep'
* or maybe the issue is that more control is needed over how the attenuation happens. maybe there should be a scale value so that you can still get big wavelengths in shallow water

thoughts?

GXMark:
I have been thinking about this process in more depth especially since you mentioned lerping down to deep.

I don't think it is necessarily a good idea to base the depth cache off the terrain as it is inherently tedious to shape it to suit the need of the waves.  What if we use a transform position with a component to specify what the depth should be.   Each transform effectively creates a position and depth which would require  some interpolation between the positions to dynamically build the depth cache.   This is very powerful approach because it disassociates the terrain shape from the depth cache thus removing terrain design difficulties from the solution.

Is it possible to interpolate between a set of adjacent transform points with assigned depths within a 2D grid space to create a depth cache ?

your thoughts?

huwb:
its straightforward to provide your own shape - for example you could deform a plane in blender as desired, then it can be brought in and assigned to a particular layer like "SeabedProxy" or such and the depth cache can be configured to render this later.

more crudely one could also kitbash together a bunch of unity primitives to create the shape, or similar quick and dirty approaches.

a third avenue would be to use ProBuilder so you can work entirely in Unity. i found it to be slightly ropey but its probably good enough for this

[...]

PR #946, opened 2021-11-24
Optional maximum wave attenuation

daleeidd:
One solution for #726.

Adds a maximum depth value for wave attenuation. Anything below this value will have no attenuation. Drawbacks is that wave attenuation is overall less effective. Moderate spectrums seem to work well enough. Completely optional as if the value is set to the maximum, the interpolation is skipped so this is a safe change.

TODO: documentation and release notes

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgrSeaFloorDepth.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgrSeaFloorDepth.cs
@@ -16,6 +16,7 @@ namespace Crest
     /// </summary>
     public class LodDataMgrSeaFloorDepth : LodDataMgr
     {
+        // NOTE: Must match CREST_OCEAN_DEPTH_BASELINE in OceanConstants.hlsl.
         internal const float k_DepthBaseline = 1_000f;
 
         public override string SimName => "SeaFloorDepth";
--- a/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgrSeaFloorDepth.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgrSeaFloorDepth.cs
@@ -16,11 +16,13 @@ namespace Crest
     /// </summary>
     public class LodDataMgrSeaFloorDepth : LodDataMgr
     {
+        internal const float k_DepthBaseline = 1_000f;
+
         public override string SimName => "SeaFloorDepth";
         protected override GraphicsFormat RequestedTextureFormat => Settings._allowVaryingWaterLevel ? GraphicsFormat.R32G32_SFloat : GraphicsFormat.R16_SFloat;
         protected override bool NeedToReadWriteTextureData => false;
         // We want the clear colour to be the min terrain height (-1000m) in X, and sea level offset 0m in Y.
-        readonly static Color s_nullColor = Color.red * -1000f;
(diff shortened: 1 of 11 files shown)


huwb:
I tried it and i think it works really well. i stress tested it and i had to go fairly extreme to get waves intersecting with the underlying terrain, is this what you refer to by "overall less effective"? i think it might be good to specify what hte drawback can be in the tooltip as its not necessarily clear what less effective means in practice.

i've been thinking about how to expose the parameter. its under the "Attenuation in Shallows". i wonder if maybe calling it "Shallow Max Depth" or something, and perhaps tweaking the tooltip to say "Any water deeper than this will receive full wave strength". i'm not sure these suggestions are perfect but just throwing them in the mix.

huwb:
I'm a thumbs up for merging once you're happy but i am happy to re-review later if it helps as well

crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgrSeaFloorDepth.cs
```diff
@@ -16,11 +16,13 @@ namespace Crest
     /// </summary>
     public class LodDataMgrSeaFloorDepth : LodDataMgr
     {
+        internal const float k_DepthBaseline = 1_000f;
```
huwb, line 19: may be worth also referencing the constant in the shader code from here (you have already referenced in the other direction)

crest/Assets/Crest/Crest/Scripts/LodData/Settings/SimSettingsAnimatedWaves.cs
```diff
@@ -27,6 +27,11 @@ public partial class SimSettingsAnimatedWaves : SimSettingsBase
         float _attenuationInShallows = 0.95f;
         public float AttenuationInShallows => _attenuationInShallows;
 
+        [Tooltip("Any deeper than this value will have no attenuation (ie full wave spectrum). Set to maximum value (1,000) to disable. If it is used, it will reduce the overall attenuation effectiveness.")]
+        [SerializeField, Range(0f, LodDataMgrSeaFloorDepth.k_DepthBaseline)]
```
huwb, line 31: it may be we should not allow this to be 0 to avoid div by 0, (or clamp it before giving it to the shader)

crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgrSeaFloorDepth.cs
daleeidd: Done. Thanks.

crest/Assets/Crest/Crest/Scripts/LodData/Settings/SimSettingsAnimatedWaves.cs
daleeidd: I made minimum 1. I doubt anyone wants anything below that and nice and round.

daleeidd:
> I tried it and i think it works really well. i stress tested it and i had to go fairly extreme to get waves intersecting with the underlying terrain, is this what you refer to by "overall less effective"? i think it might be good to specify what hte drawback can be in the tooltip as its not necessarily clear what less effective means in practice.

Thanks. Pretty much. I've updated it. LMKWYT.

Also addressed other feedback.

huwb (approved):
Looks good

huwb:
Nice one

daleeidd:
Thanks.

PR #947, opened 2021-11-25
Fix shadow bleeding at shorelines

daleeidd:
Jittering shadow sampling causes shadows to bleed as a sample can land inside of geometry where it is shadowed. Easier to explain with screenshots:

You may notice there is still leaking, but that is only because the data being rendered is from the vertex shader which doesn't have enough density at that distance. Below is if the shadows are moved to the pixel shader (not part of this PR):

## Performance

Adds about 0.04-0.05ms to the entire shadow sampling pass.

## Testing

I have left the shadows rendered on the ocean surface on (from the vertex shader). Toggle "Create Sea Floor Depth Data" to see before and after.

### DIFF
--- a/crest/Assets/Crest/Crest/Shaders/OceanHelpersNew.hlsl
+++ b/crest/Assets/Crest/Crest/Shaders/OceanHelpersNew.hlsl
@@ -21,6 +21,19 @@ float3 WorldToUV(in float2 i_samplePos, in CascadeParams i_cascadeParams, in flo
 	return float3(uv, i_sliceIndex);
 }
 
+// Moves to the next LOD slice if out of bounds of current LOD slice.
+float3 WorldToSafeUV(in float2 i_samplePos, in CascadeParams i_cascadeParams, in float i_sliceIndex)
+{
+	float2 uv = WorldToUV(i_samplePos, i_cascadeParams);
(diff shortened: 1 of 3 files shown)


crest/Assets/Crest/Crest/Shaders/Ocean.shader
```diff
@@ -438,6 +438,8 @@ Shader "Crest/Ocean"
 				// We need this when sampling a screenspace texture.
 				UNITY_SETUP_STEREO_EYE_INDEX_POST_VERTEX(input);
 
+				return half4(input.flow_shadow.zw, 0.0, 1.0);
```
huwb, line 441: Assume this is debug :)

huwb (approved):
Yeah I think this works. Trying to think of it will fail in some situation. With checking if it works inside the cave in pirate cove.

Approved anyway, nice work

crest/Assets/Crest/Crest/Shaders/Ocean.shader
daleeidd: Indeedy. Removed :^)

daleeidd:
> Yeah I think this works. Trying to think of it will fail in some situation. With checking if it works inside the cave in pirate cove.
> 
> Approved anyway, nice work

Thanks. I checked Pirate Cove and it has no negative affect. If I add the rocks to the depth cache it improves it. I checked different sun orientations too. Overall an improvement.

PR #948, opened 2021-11-26
Fix server builds and LodData* toggling

daleeidd:
The LodData buffering breaks server builds currently as I forgot to test it. This is just a quick fix. It sets the OceanRenderer.BufferSize to 2 as a minimum. This value sets the shared buffered size like for cascade params and doesn't affect the buffered size for individual LodDataMgrs.

Also, if no sims requiring buffer count of two were enabled, exceptions would also be thrown as it assumed two as minimum.

Will make a proper fix. I have also added checking server/headless options to the release process to catch this in the future.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
+++ b/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
@@ -502,8 +502,9 @@ namespace Crest
             CreateDestroySubSystems();
 
             // TODO: Have a BufferCount which will be the run-time buffer size or prune data.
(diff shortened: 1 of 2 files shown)


huwb (approved):
Sorry for delay

I haven't quite understood, after a brief reading of your notes. I think I'm not familiarity enough with what this count means at different levels.

However if you are preparing a better fix later and want this to go in for now, I'm fine with that

huwb:
Reread.. yeah I think it would be good to later change the logic so that cascade params doesn't depend on this same value? Or maybe BufferSize should be renamed to make things clearer. Anyway thumbs up for this

daleeidd:
Thanks. Sounds good.

PR #949, opened 2021-11-27
Add Attenuation in Shallows setting to Dynamic Waves

daleeidd:
Since we have one for animated waves, it makes sense to have it for dynamic waves too. Had feedback that the depth cache was dampening wakes which was undesirable.


huwb (approved):
Looks good

daleeidd:
Thanks.

PR #950, opened 2021-12-04
Fix floating origin for ShapeGerstner and ShapeFFT

daleeidd:
Adds an offset as a global shader vector which the Gerster/FFT shader uses. It's not perfect but same tolerance as ShapeGerstnerBatched. The benefit of the global shader vector is no need to search for Shape* components.


huwb (approved):
Wow lovely

daleeidd:
Thanks

PR #951, opened 2021-12-05
Fix Floating Origin popping

daleeidd:
_**Status:** Improvement. But not complete._

Fixes Floating Origin popping on teleports for:
- mesh
- normal maps
- foam
- bubbles
- caustics
- persistent simulations

But the following is still a problem:
- [x] wave jitter (very noticeable in whirlpool scene and might have something to do with combine shader)
- [x]  collision queries?
- [x] conditional compilation
- [x] wave patch


daleeidd:
@huwb any thoughts on this so far?

huwb:
Yeah looks good so far.

Yeah i suppose the wave queries need the offset too

Ideally the offset would be mod'd down. The whole point of floating origin is to take away a large amount of the position, i.e. if player is 10km from origin, an offset of 10km is subtracted from everything. You store 10km in a variable and then add it everywhere to compensate and i think we unfortunately expose ourselves to the original precision issues.

For a normal map that tiles every say 12m, it should be possible to instead of storing the full world space offset, store `offset % 12f` which will be much smaller. that might be a good option for things that obviously repeat like textures, assuming we can read their periodicity from the material params or etc.

for waves its a bit trickier. maybe modding with the size of the biggest wave slice does it. maybe modding with 2048 is big enough for 99% of situations.

does this make sense?

daleeidd:
> Ideally the offset would be mod'd down. The whole point of floating origin is to take away a large amount of the position, i.e. if player is 10km from origin, an offset of 10km is subtracted from everything. You store 10km in a variable and then add it everywhere to compensate and i think we unfortunately expose ourselves to the original precision issues.

Thanks. I didn't see any anomalies without it, but make sense.

> For a normal map that tiles every say 12m, it should be possible to instead of storing the full world space offset, store `offset % 12f` which will be much smaller. that might be a good option for things that obviously repeat like textures, assuming we can read their periodicity from the material params or etc.

`% _NormalsScale * 2.0 * lodDataGridSize` appeared to work in all cases. We could either use a keyword or calculate in C# and pass through to each chunk. Will look at doing the other textures.

> for waves its a bit trickier. maybe modding with the size of the biggest wave slice does it. maybe modding with 2048 is big enough for 99% of situations. so basically LodDataMgrAnimWaves could store `offset % 2048` as a test. Maybe this works for other loddata too.

In regards to the jitter? It appears to be something else as it occurs even not far from zero. It looks like it has something to do with scale, LOD count, and FO threshold. Eyeballing the data there is a slight difference between non FO and FO:



That being said, I was able to reduce (possibly eliminate) the jitter by playing with LOD scale, LOD count, and LOD resolution (and in some case FO threshold), but I haven't been able to determine how these values correlate mathematically yet. Perplexed why this is even an issue at like 100 m from zero.

huwb:
Hmm weird. 100m from zero presumably has a very small offset - but adding this offset makes it jitter?

Test sounds good

daleeidd:
Done all of the textures and sims using mod.

> Hmm weird. 100m from zero presumably has a very small offset - but adding this offset makes it jitter?

Yeah. It shifts it a little bit when the origin is moved. I thought I managed to fix it but it's still there.

> Yeah i suppose the wave queries need the offset too

Using the query visualiser it is possible to add the offset as the queries return and keep the result in the expected place. But not sure whether this is the correct approach yet. Anything running in FixedUpdate probably won't be able to use this approach as FO is LateUpdate.

Will put this down for a little bit and push other stuff forward.

daleeidd:
I think this is ready for early review...

I moved FO to FixedUpdate which solves queries I think. If queries return everything in local space then I don't think we need to add FO offset to it. Is that correct?

There are two dev scenes. One is for LOD data and the other for Ocean.shader textures. Once #931 I will merge them together using some of the scripts from there and prefab it.

For the first scene, just hit play and it will cross the threshold and should pause automatically.

For the second scene, before entering play mode, hit pause and then enter play mode so you are in paused state, then you can step through the frames and cross the threshold. The camera moves automatically and threshold is less than a unity away.

I added some general debug options and developer only debug options:

Anything under "Developer Only" only appears if CREST_DEBUG symbol is present. Those debug options are for positioning the camera at an offset from the threshold.

As for the waves have a slight pop, I think I have fixed it. There is still a pixel level pop but I doubt it is noticeable. I am viewing things with the ocean in a paused state but in movement probably won't be noticeable.

I'll need to add release notes still and clean up.

daleeidd:
I found that the combination of restricting the threshold to a minimum of 4096 and power of 2 to avoid wave popping. Also I only shift the origin by threshold sized increments which also was needed. Aforementioned is done in FloatingOrigin.cs. With this subtracting the origin in the wave shader has no positive effect and probably can be removed. I will have to investigate further but wouldn't mind hearing your thoughts.

crest/Assets/Crest/Crest/Scripts/Helpers/FloatingOrigin.cs
```diff
@@ -18,7 +18,9 @@
 // NOTE 2: Of particular note - any world space texturing should have a period that tiles perfectly across the teleport distance, otherwise visible
 // pops will occur. Example - set teleport radius to 16384m, and set normal map scale to 16m which divides evenly into the teleport radius.
 
+using System.Collections;
 using UnityEngine;
+using UnityEngine.Experimental.Rendering;
```
huwb, line 23: does this work in our current supported versions? just making sure there isnt something that got moved out of Experimental

crest/Assets/Crest/Crest/Scripts/Helpers/FloatingOrigin.cs
```diff
@@ -65,35 +69,149 @@ public class FloatingOrigin : MonoBehaviour
         [Tooltip("Optionally provide a list of Gerstner components to avoid doing a FindObjectsOfType() call."), SerializeField]
         ShapeGerstnerBatched[] _overrideGerstnerList = null;
 
+        [Space(10)]
```
huwb, line 72: this is good. i should probably look at adding spaces to try make OceanRenderer ui a bit clearer

crest/Assets/Crest/Crest/Scripts/Helpers/FloatingOrigin.cs
```diff
@@ -65,35 +69,149 @@ public class FloatingOrigin : MonoBehaviour
...
+                // Cannot clear at start of FixedUpdate (without extra logic) as it can run multiple times.
+                TeleportOriginThisFrame = Vector3.zero;
 
+                _isFirstFixedUpdate = true;
+            }
+        }
+
+        void FixedUpdate()
+        {
+            if (!_isFirstFixedUpdate)
```
huwb, line 133: the way i've typically handled this is something like

int _lastUpdateFrame;

void FixedUpdate()
{
    if( _lastUpdateFrame == Time.frameCount )
    {
        return;
    }
    _lastUpdateFrame = Time.frameCount;

    ....
}

this would be a bit simpler for me. also if this were to mean we could make the coroutine above debug-only (so not in final builds) i'd be in favour - but only because i'm hopelessly bad at coroutines and view them with suspicion :). but if you prefer to keep it the way it is and the coroutine is not generating per frame garbage then i wouldnt block the review on it.

crest/Assets/Crest/Crest/Scripts/Helpers/FloatingOrigin.cs
```diff
@@ -65,35 +69,149 @@ public class FloatingOrigin : MonoBehaviour
...
-            TeleportOriginThisFrame = Vector3.zero;
+            while (true)
+            {
+                // NOTE: Will not work in batch mode:
+                // https://docs.unity3d.com/Manual/CLIBatchmodeCoroutines.html
+                yield return Helpers.WaitForEndOfFrame;
+
+                if (_debug._isCapturing)
+                {
+                    ExternalGPUProfiler.EndGPUCapture();
```
huwb, line 113: interesting, nice

crest/Assets/Crest/Crest/Scripts/Helpers/FloatingOrigin.cs
```diff
@@ -65,35 +69,149 @@ public class FloatingOrigin : MonoBehaviour
...
             if (newOrigin != Vector3.zero)
             {
                 MoveOrigin(newOrigin);
             }
         }
 
+        void OnValidate()
+        {
+            // Must be power of two to avoid popping.
+            _threshold = Mathf.Pow(2f, Mathf.Round(Mathf.Log(_threshold, 2f)));
```
huwb, line 188: does > 2048 need to be validated here too? I know you have a min attribute but its public and could be set from code or users could hack the min i guess

crest/Assets/Crest/Crest/Shaders/OceanConstants.hlsl
```diff
@@ -13,6 +13,9 @@
 // C# code to check if any parameters are within the MAX_LOD_COUNT limits
 #define MAX_LOD_COUNT 15
 
+// The size of the largest LOD.
```
huwb, line 16: is this the size of the largest lod? i think it might be good to expand this comment and rename the symbol to make it really explicit and clear. i cant quite remember the background for this, i guess its in our thread but it would be good to make the code really clear

crest/Assets/Crest/Crest/Shaders/OceanEmission.hlsl
```diff
@@ -70,7 +70,8 @@ void ApplyCaustics
 {
 	// could sample from the screen space shadow texture to attenuate this..
 	// underwater caustics - dedicated to P
-	const float3 scenePosUV = WorldToUV(i_scenePos.xz, cascadeData1, _LD_SliceIndex + 1);
+	// TODO: Investigate if necessary anymore.
```
huwb, line 73: soft vote for looking into this if its fast rather than adding a todo, but also fine with it being deferred.

crest/ProjectSettings/ProjectSettings.asset
```diff
@@ -620,7 +620,7 @@ PlayerSettings:
   webGLThreadsSupport: 0
   webGLDecompressionFallback: 0
   scriptingDefineSymbols:
-    1: CROSS_PLATFORM_INPUT;UNITY_POST_PROCESSING_STACK_V2;CREST_OCEAN
+    1: CROSS_PLATFORM_INPUT;UNITY_POST_PROCESSING_STACK_V2;CREST_OCEAN;CREST_DEBUG
```
huwb, line 623: just double checking this is intentional? not objecting just checking

crest/Assets/Crest/Crest/Shaders/OceanEmission.hlsl
```diff
@@ -70,7 +70,8 @@ void ApplyCaustics
...
+	const float3 scenePosUV = WorldToUV(CREST_WITH_FLOATING_ORIGIN_LOD_OFFSET_XZ(i_scenePos.xz), cascadeData1, _LD_SliceIndex + 1);
```
huwb, line 74: this is the only case where this FO macros is used going into WorldToUv, it might be good to document why its needed here. or maybe thats the TODO?

crest/Assets/Crest/Crest/Shaders/OceanHelpersNew.hlsl
```diff
@@ -21,6 +21,18 @@ float3 WorldToUV(in float2 i_samplePos, in CascadeParams i_cascadeParams, in flo
 	return float3(uv, i_sliceIndex);
 }
 
+float3 WorldToUVSource(in float2 i_samplePos, in CascadeParams i_cascadeParams, in float i_sliceIndex)
```
huwb, line 24: i didnt automatically understand the reason for this function. is there an iteration on the naming that would make it clearer (even if its quite explicit), or could we add a comment above it. just to make it clear which function to use from this library for different situations.

huwb (commented):
Thats it for this pass, awesome work. Thanks for getting all the debug functionality in!

crest/Assets/Crest/Crest/Scripts/Helpers/FloatingOrigin.cs
daleeidd: It is experimental for all versions so far. I can surround it and dependent code with CREST_DEBUG for now? It is for the RenderDoc capture which I doubt anyone will need.

daleeidd: After reading your other comment I will wrap with CREST_DEBUG as going to wrap that stuff too.

crest/Assets/Crest/Crest/Scripts/Helpers/FloatingOrigin.cs
daleeidd: Thanks. I have changed to your approach as it does reduce our dependency on the coroutine. But we still depend on it because TeleportOriginThisFrame and HasTeleportedThisFrame has to be cleared but FixedUpdate does not run each frame.

crest/Assets/Crest/Crest/Scripts/Helpers/FloatingOrigin.cs
daleeidd: Good find. Done. Thanks.

crest/Assets/Crest/Crest/Scripts/Helpers/FloatingOrigin.cs
daleeidd: Done. I have wrapped it with CREST_DEBUG and wrapped RenderDoc too. Thanks.

crest/Assets/Crest/Crest/Scripts/Helpers/FloatingOrigin.cs
daleeidd: Indeed. It was very useful/necessary as LOD doesn't run when paused.

crest/Assets/Crest/Crest/Shaders/OceanConstants.hlsl
daleeidd: I have since removed it as the issue was actually a bug in C#.

crest/Assets/Crest/Crest/Shaders/OceanEmission.hlsl
daleeidd: I've removed it. Wasn't necessary anymore after testing.

crest/ProjectSettings/ProjectSettings.asset
daleeidd: It is. For convenience since FO testing is wrapped with this.

crest/Assets/Crest/Crest/Shaders/OceanEmission.hlsl
daleeidd: No longer needed. Fixed in C#. Thanks :^)

crest/Assets/Crest/Crest/Shaders/OceanHelpersNew.hlsl
daleeidd: Removed as fixed in C#.

crest/Assets/Crest/Crest/Scripts/Helpers/FloatingOrigin.cs
daleeidd: As for garbage, I did not see any from this. Found some other garbage :\ which I will fix separately.

daleeidd:
Feedback addressed. Thanks.

I was able to remove most of the shader code as I was able to fix it in C# instead.

After review I will fix up the dev stuff and do release notes / docs. But I think this fixes everything FO wise.

crest/Assets/Crest/Crest/Shaders/Resources/UpdateFoam.compute
```diff
@@ -41,30 +41,33 @@ void UpdateFoam(uint3 id : SV_DispatchThreadID)
 	const CascadeParams cascadeData = _CrestCascadeData[sliceIndex];
 	const float2 worldPosXZ = UVToWorld(input_uv, sliceIndex, cascadeData);
 	const float3 uv_slice = float3(input_uv, id.z);
+	half foam = 0.0;
```
huwb, line 44: the diff of this file is a little hard to review here which is fine but just checking this is safe to do? the logic below can be a bit delicate but if it looks good ignore this comment

huwb (approved):
Yeah just one paranoia question but ill approve, nice one

crest/Assets/Crest/Crest/Shaders/Resources/UpdateFoam.compute
daleeidd: It is only whitespace changes sorry. If hide whitespace is enabled it makes it easier:

daleeidd: In case you're on mobile and it isn't available:



I might drop the whitespace changes and do separately.

huwb: Oh forgot about that thanks. Looks good to me, feel free to keep

daleeidd:
I've added release notes and removed paragraphs about FO limitations since I believe there aren't any anymore 🤞

huwb:
Great!

nitroman123:
Everything work well I've tested , no popping not nothing well done !!!

PR #952, opened 2021-12-06
Fix scale smoothing on height change on teleports

daleeidd:
For the varying water level feature, smoothing between the different levels was added. But it didn't take teleports into account.

So if I had the scene camera up very high above sea level and the game camera at sea level, and then entered play mode, the smoothing would be applied from the first frame. Or if I were to teleport from a beach to a lake at a higher elevation inland then, then it would also smooth which is not necessary.

This PR handles generic teleports with Floating Origin support. Since I don't know the best threshold, I exposed a property (_Teleport Threshold_) and set the default value to 10.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
+++ b/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
@@ -107,6 +107,11 @@ namespace Crest
             }
         }
 
+        [Tooltip("The height where detail is focused is smoothed to avoid popping which is undesireable after a teleport. Threshold is in Unity units."), SerializeField]
+        float _teleportThreshold = 10f;
+        float _teleportTimer = 0f;
+        Vector3 _oldViewerPosition = Vector3.zero;
+
         public Transform Root { get; private set; }
 
         // does not respond to _timeProvider changing in inspector
@@ -459,6 +464,9 @@ namespace Crest
                 return;
             }
 #endif
+
+            // Force teleport timer after being enabled to always disable scale smoothing for first second.
+            _teleportTimer = 1f;
 
             // Setup a default time provider, and add the override one (from the inspector)
             _timeProviderStack.Clear();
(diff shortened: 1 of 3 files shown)


huwb:
Thanks for putting this forward

I didn't understand the distinction in the code between normal teleports and FO teleports. I guess an FO teleport does not require any special action as everything is moved together - is that right? Maybe a bit more comment detail on each step is useful.

daleeidd:
Thanks.

> I didn't understand the distinction in the code between normal teleports and FO teleports.

After more thorough testing, I discovered that logic was incorrect. I have replaced it with a better solution.

> I guess an FO teleport does not require any special action as everything is moved together - is that right?

Pretty much. We need to exclude FO shifts, because it is effectively the same position, but still be able to recognise normal teleports. New solution does this correctly.

daleeidd:
@huwb ping. In case you forgot about this one. Whenever you're ready :^)

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
```diff
@@ -1081,9 +1086,33 @@ void LateUpdateViewerHeight()
...
+            // Calculate teleport distance and create window for height queries to return a height change.
+            {
+                if (_teleportTimer > 0f)
+                {
+                    _teleportTimer -= Time.deltaTime;
+                }
+
+                // Find the distance. Adding the FO offset will exclude FO shifts so we can determine a normal teleport.
+                // FO shifts are visually the same position and it is incorrect to treat it as a normal teleport.
+                var teleportDistance = (_oldViewerPosition - camera.transform.position - FloatingOrigin.TeleportOriginThisFrame).sqrMagnitude;
```
huwb, line 1098: perhaps call this `teleportDistanceSqr`

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
```diff
@@ -1081,9 +1086,33 @@ void LateUpdateViewerHeight()
...
+                // Threshold as sqrMagnitude.
+                var threshold = Mathf.Pow(_teleportThreshold, 2f);
```
huwb, line 1100: i'd prefer multiplying by itself rather than calling into a .net function

huwb (approved):
Thanks for bumping. minor feedback but thumbs up from me

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
daleeidd: Done

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
daleeidd: Sure. Done

daleeidd:
Addressed all review items. Also, I am now setting the teleport timer to 1s in OnEnabled so that we always disable smoothing for the first load.

PR #954, opened 2021-12-09
Fix shadow simulation error recovery

daleeidd:
When the shadow simulation would hit an error it would disable itself and become unrecoverable - even in edit mode. This PR mainly addresses that but lead to a few other fixes and improvements:

- Failed state is now recoverable in edit mode without error spam
- Add "allow no shadows" option as disabling shadows on a light dynamically can be valid
- Fix shadow data not always clearing
- Fix "allow null light" from not working correctly

I've added an Error enum to prevent error spam which works well since errors are reported in order and one at a time.


huwb (approved):
Looks great nice work

daleeidd:
Thanks.

PR #955, opened 2021-12-10
Add option for Ocean Renderer to ignore Water Body culling

daleeidd:
With this option, the _Water Body Override Material_ option can be used whilst still having an ocean. I made it _public_ as I believe people may want to script this.


huwb (approved):
Looks good.

Using quotes around the component name in the tooltip seems fine (its a shame italics dont work). If we like this maybe we should commit to this pattern everywhere. But thats a future thought..

daleeidd:
Thanks. I switched to single quotes as that is what we used elsewhere.

PR #956, opened 2021-12-10
Improve Water Body overlap when using Override Material

daleeidd:
Calculates the overlapping area between water body and water tile to determine the best suited water body when using the material override feature. Makes this feature more predictable.

Before these changes, the water body which is first enabled would be chosen regardless of how significant the overlap was. This leaves the feature to the mercy of Unity's script execution order. And is quite confusing when toggling these components in the editor.


daleeidd:
Thanks.

PR #957, opened 2021-12-10
Underwater Renderer support for Water Body material

daleeidd:
Adds support for the water body material override. It grabs the material from one of the inner tiles. Seems to work well enough if we consider how it is used (WB should be far enough apart).

**EDIT:** Change the solution to something more accurate to the WB itself. Checking camera position is within the WB bounds in the UR. Other benefit is that it is code is contained within the UR which is nicer. Check the first commit to see original solution.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Effect.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Effect.cs
@@ -37,6 +37,7 @@ namespace Crest
 
         CommandBuffer _underwaterEffectCommandBuffer;
         PropertyWrapperMaterial _underwaterEffectMaterial;
+        Material _currentOceanMaterial;
         internal readonly UnderwaterSphericalHarmonicsData _sphericalHarmonicsData = new UnderwaterSphericalHarmonicsData();
 
         RenderTargetIdentifier _colorTarget = new RenderTargetIdentifier
@@ -243,7 +244,7 @@ namespace Crest
             }
         }
 
-        internal static void UpdatePostProcessMaterial(
+        internal void UpdatePostProcessMaterial(
             Mode mode,
             Camera camera,
             PropertyWrapperMaterial underwaterPostProcessMaterialWrapper,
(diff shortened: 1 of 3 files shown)


daleeidd:
Change the solution to something more accurate to the WB itself. Checking camera position is within the WB bounds in the UR. Other benefit is that it is code is contained within the UR which is nicer. Check the first commit to see original solution.

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Effect.cs
```diff
@@ -115,8 +115,30 @@ int dataSliceOffset
             Material underwaterPostProcessMaterial = underwaterPostProcessMaterialWrapper.material;
             if (copyParamsFromOceanMaterial)
```
huwb, line 116: just to check, will things still work ok if `copyParamsFromOceanMaterial` is false - i guess it will take params from the global water material as usual?

huwb (approved):
Looks good

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Effect.cs
daleeidd: It will take global OR material or a WB material if it finds one which means it could get stuck with WB material. I've added validation to recommend users enable it if they use a WB material. LMKWYT

daleeidd: @huwb ping

huwb: Maybe it should always copy params each frame if WBs with material overrides are present. Or even better, cache which material it last copied params from and only copy when necessary (if "always copy" is not enabled, if it is just always copy). What do you think?

In any case I'm happy for this to merge.

daleeidd: The latter sounds great. I've gone with that. Thanks!

daleeidd:
Thanks!

PR #958, opened 2021-12-12
Prewarm foam sim on first frame and teleports

daleeidd:
This prewarms the foam sim if it is the first step since being enabled or the first step for a teleport.

Currently foam can take a second or so to populate and the first moment has no foam. This PR mitigates that by prewarming the sim.

I worked out that swapping the `_SimDeltaTime` with `1.0 / _FoamFadeRate` will match the foam generated when the ocean is in a paused state in the editor (add a custom time provider and override time as zero or pause). So when entering play mode, it is seamless.

It's not the same as foam produced from running all the sims a few frames which would be the perfect solution, but that is probably costly for teleports.

1_1_Target
Editor (ocean is at time zero)

1_2_Current
First frame of play mode (current)

1_3_Fixed
First frame of play mode (this PR)

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgrFoam.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgrFoam.cs
@@ -33,6 +33,7 @@ namespace Crest
         readonly int sp_WaveFoamCoverage = Shader.PropertyToID("_WaveFoamCoverage");
         readonly int sp_ShorelineFoamMaxDepth = Shader.PropertyToID("_ShorelineFoamMaxDepth");
         readonly int sp_ShorelineFoamStrength = Shader.PropertyToID("_ShorelineFoamStrength");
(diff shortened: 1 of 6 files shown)


huwb (approved):
Cool so this doesnt "overshoot" - make intense foam in patches which then fades? If it looks good then big thumbs up from me, good idea

daleeidd:
> Cool so this doesnt "overshoot" - make intense foam in patches which then fades?

After further testing it can overshoot slightly depending on the foam strength and coverage. If they are low it will overshoot slightly (perhaps negligibly). If they are high it more undershoots compared to running the sim multiple times.

huwb:
Ok yeah, doesn't sound too bad, but an option would be to add 80% or 90% of the current amount to compensate slightly. But I'll let you decide.

daleeidd:
I changed the formula which works a lot better overall. Noticeable overshooting was mainly at the lower end of _WaveFoamStrength which I have now factored in. I've also made it optional just in case.

huwb:
Nice one

daleeidd:
Thanks.

PR #959, opened 2021-12-15
Fix underwater high memory

daleeidd:
Reverts #928

I used Unity's memory profiler and saw a ballooned amount of RT memory usage. There were multiple instances of the same temporary RTs. I have gone back to using creating RTs manually :^(

Unity has not gotten back to me yet (been over a month). Unity's temporary RT pool has another strange bug, where temporary RTs would be bound to the wrong variable, which they have confirmed. This was preventing me from finishing underwater portals so will be able to finally finish it.

If you're wondering why I am using a static reference, it is because URP shares these APIs. The UR will hold the RTs which URP will use. This will reduce friction when merging and when reverting back to temporary RTS once bug is fixed.


huwb (approved):
Fun times!

daleeidd:
Thanks :^)

PR #960, opened 2021-12-20
Underwater from geometry

daleeidd:
Adds a feature to render the underwater effect from a provided mesh. New options highlighted in yellow:

Portals

There are four modes:
- Full Screen: Same as before
- Geometry 2D: Only front face. Fog depth is to the far plane
- Geometry 3D: Front face and back face to create a volume of water
- Geometry Volume: As above but can fly through

For the _Geometry Volume_ option, the stencil buffer is used. It is a separate stencil buffer context so it will not interfere with the global stencil buffer.

Two global keywords have been used. But these are unlimited in 2022.

## Performance

- Higher memory usage due to extra variants for ocean shader and extra passes for underwater. They are not stripped from builds so it is a new baseline cost **(1MB)**
- Every frame SetUpBoundary is called which is new cost but probably negligible
- Full screen should have same performance as before **(Yes)**
- 2D has 1 extra SS texture for front-faces and extra pass to populate. Might be more expensive due to mesh vs FST
- 3D in addition has another extra SS texture for back faces and extra pass to populate
- Volume in addition uses the stencil buffer which requires copy/clear passes to populate and make separate from global stencil buffer. It has two extra passes to fog things but no passes retreads old ground

### Memory Cost

I didn't see anything unexpected. 1MB new baseline cost due to shader passes/variants. And just the cost for each texture.

### Shader Cost

1660 Ti. Following times are for the entire transparent pass.

- Full-Screen Before: ~3.12ms
- Full-Screen After: ~3.08ms
- Portal (Quad): ~3.54ms
- Volume (Sphere): ~3.12ms
- Volume FT (Sphere): ~3.36ms

Portal, Volume and Volume FT are averages of three runs. Fluctuated quite a bit from 2.64ms to 4.11ms (fluctuation was mostly in the mask). But it appears that we are looking at a ~0.5ms extra cost when using one of these modes.

~~Will check shader and memory costs soon.~~

## TODO

- [x] Caustics fade away at a distance due to scene depth applied to depth of field.
- [ ] Remove test data
- [x] Profile

## Testing

Open the _Demo_ scene. Anything under Demo will be deleted before merging.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Effect.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Effect.cs
@@ -13,17 +13,46 @@ namespace Crest
         const string k_ShaderPathUnderwaterEffect = "Hidden/Crest/Underwater/Underwater Effect";
         internal const string k_KeywordFullScreenEffect = "_FULL_SCREEN_EFFECT";
         internal const string k_KeywordDebugViewOceanMask = "_DEBUG_VIEW_OCEAN_MASK";
+        internal const string k_KeywordDebugViewStencil = "_DEBUG_VIEW_STENCIL";
 
         internal static readonly int sp_CrestCameraColorTexture = Shader.PropertyToID("_CrestCameraColorTexture");
+        static readonly int sp_CrestWaterVolumeStencil = Shader.PropertyToID("_CrestWaterVolumeStencil");
         static readonly int sp_InvViewProjection = Shader.PropertyToID("_InvViewProjection");
         static readonly int sp_InvViewProjectionRight = Shader.PropertyToID("_InvViewProjectionRight");
         static readonly int sp_AmbientLighting = Shader.PropertyToID("_AmbientLighting");
         static readonly int sp_HorizonNormal = Shader.PropertyToID("_HorizonNormal");
         static readonly int sp_DataSliceOffset = Shader.PropertyToID("_DataSliceOffset");
 
+        // If changed then see how mode is used to select the front-face pass and whether a mapping is required.
+        // :UnderwaterRenderer.Mode
+        enum EffectPass
+        {
+            FullScreen,
+            VolumeFrontFace2D,
+            VolumeFrontFace3D,
+            VolumeFrontFaceVolume,
+            VolumeBackFace,
+            VolumeScene,
+        }
+
         CommandBuffer _underwaterEffectCommandBuffer;
         PropertyWrapperMaterial _underwaterEffectMaterial;
         internal readonly UnderwaterSphericalHarmonicsData _sphericalHarmonicsData = new UnderwaterSphericalHarmonicsData();
+
+        RenderTargetIdentifier _colorTarget = new RenderTargetIdentifier
+        (
+            BuiltinRenderTextureType.CameraTarget,
+            0,
+            CubemapFace.Unknown,
+            -1
+        );
+        RenderTargetIdentifier _depthStencilTarget = new RenderTargetIdentifier
+        (
+            sp_CrestWaterVolumeStencil,
+            0,
+            CubemapFace.Unknown,
+            -1
+        );
 
         internal class UnderwaterSphericalHarmonicsData
         {
@@ -57,6 +86,7 @@ namespace Crest
 
             RenderTextureDescriptor descriptor = XRHelpers.GetRenderTextureDescriptor(_camera);
             descriptor.useDynamicScale = _camera.allowDynamicResolution;
+
             // Format must be correct for CopyTexture to work. Hopefully this is good enough.
             if (_camera.allowHDR)
             {
@@ -64,14 +94,17 @@ namespace Crest
             }
 
             var temporaryColorBuffer = RenderTexture.GetTemporary(descriptor);
+            temporaryColorBuffer.name = "_CrestCameraColorTexture";
 
             UpdatePostProcessMaterial(
+                _mode,
                 _camera,
                 _underwaterEffectMaterial,
                 _sphericalHarmonicsData,
                 _meniscus,
                 _firstRender || _copyOceanMaterialParamsEachFrame,
                 _debug._viewOceanMask,
+                _debug._viewStencil,
                 _filterOceanData
             );
 
@@ -80,10 +113,49 @@ namespace Crest
 
             _underwaterEffectCommandBuffer.Clear();
 
-            if (_camera.allowMSAA)
+            // Create a separate stencil buffer context by copying the depth texture.
+            if (UseStencilBufferOnEffect)
+            {
+                descriptor.colorFormat = RenderTextureFormat.Depth;
+                descriptor.depthBufferBits = 24;
+                // bindMS is necessary in this case for depth.
+                descriptor.SetMSAASamples(_camera);
+                descriptor.bindMS = descriptor.msaaSamples > 1;
+
+                _underwaterEffectCommandBuffer.GetTemporaryRT(sp_CrestWaterVolumeStencil, descriptor);
+
+                // Use blit for MSAA. We should be able to use CopyTexture. Might be the following bug:
+                // https://issuetracker.unity3d.com/product/unity/issues/guid/1308132
+                if (Helpers.IsMSAAEnabled(_camera))
+                {
+                    // Blit with a depth write shader to populate the depth buffer.
+                    _underwaterEffectCommandBuffer.Blit
+                    (
+                        BuiltinRenderTextureType.None,
+                        _depthStencilTarget,
+                        Helpers.UtilityMaterial,
+                        (int)Helpers.UtilityPass.CopyDepth
+                    );
+                }
+                else
+                {
+                    // Copy depth then clear stencil.
+                    _underwaterEffectCommandBuffer.CopyTexture(BuiltinRenderTextureType.Depth, _depthStencilTarget);
+                    _underwaterEffectCommandBuffer.Blit
+                    (
+                        BuiltinRenderTextureType.None,
+                        _depthStencilTarget,
+                        Helpers.UtilityMaterial,
+                        (int)Helpers.UtilityPass.ClearStencil
+                    );
+                }
+            }
+
+            // Copy the color buffer into a texture.
+            if (Helpers.IsMSAAEnabled(_camera))
             {
                 // Use blit if MSAA is active because transparents were not included with CopyTexture.
-                // Not sure if we need an MSAA resolve? Not sure how to do that...
+                // This appears to be a bug. CopyTexture + MSAA works fine when the stencil is required.
                 _underwaterEffectCommandBuffer.Blit(BuiltinRenderTextureType.CameraTarget, temporaryColorBuffer);
             }
             else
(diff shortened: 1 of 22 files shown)


daleeidd:
As for the caustics, the scene depth is applied to the depth of field which causes them to fade away which I don't think is as desirable for this feature. I don't understand what applying the scene depth is suppose to represent. Is it blurring from the fog density or does caustics have a blurring even through air at a distance?

huwb:
Wow looks great!

If waves are all of a similar wavelength, they'll sort of have a focal point - a depth below the surface where the rays tend to be converged and generate the brightest sharpest caustics. So this feature makes them blurrier away from that depth.

It's a gross approximation and happy to try with it torn out. Maybe just fading the intensity based on depth looks better.

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Effect.cs
```diff
@@ -57,21 +86,28 @@ void OnPreRenderUnderwaterEffect()
 
             RenderTextureDescriptor descriptor = XRHelpers.GetRenderTextureDescriptor(_camera);
             descriptor.useDynamicScale = _camera.allowDynamicResolution;
+            // PERF: Auto resolves which triggers more resolves than necessary.
```
huwb, line 89: could you expand this comment. is this a todo or is it more a note that its not ideal? can this cost be avoided in some cases or does it always need to be paid?

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Effect.cs
```diff
@@ -57,21 +86,28 @@ void OnPreRenderUnderwaterEffect()
...
+            descriptor.bindMS = false;
+
             // Format must be correct for CopyTexture to work. Hopefully this is good enough.
             if (_camera.allowHDR)
             {
                 descriptor.colorFormat = RenderTextureFormat.DefaultHDR;
             }
 
             var temporaryColorBuffer = RenderTexture.GetTemporary(descriptor);
+            temporaryColorBuffer.name = "Crest Temporary Color";
```
huwb, line 100: not sure if it makes sense to add "Underwater" to this name?

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.cs
```diff
@@ -66,13 +91,16 @@ public class DebugFields
...
             public bool _disableOceanMask = false;
+            public bool _viewStencil = false;
             public bool _disableHeightAboveWaterOptimization = false;
             public bool _disableArtifactCorrection = false;
         }
 
         Camera _camera;
         bool _firstRender = true;
 
+        internal bool IsStencilBufferRequired => _mode == Mode.GeometryVolume;
```
huwb, line 102: could/should a stencil buffer be used in general? i guess if cam is far above water the effect is disabled completely. if cam is close to water or partially submerged then a stencil buffer could save a bunch of processing. maybe we dont care about this case enough to warrant the complexity/memory though.

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Mask.cs
```diff
@@ -185,6 +265,48 @@ void OnPreRenderOceanMask()
             FixMaskArtefacts(_oceanMaskCommandBuffer, descriptor, _maskTarget);
         }
 
+        void PopulateBoundary(CommandBuffer buffer, RenderTargetIdentifier frontTarget, RenderTargetIdentifier backTarget)
```
huwb, line 268: could you add a comment to explain what "Boundary" means

huwb: ah its the geometry enclosing the volume. i think it would be a bit more idiomatic to call shapes rasterised for this kind of thing as just "volumes". however if this is going to conflict with other things i'm ok with boundary but i think a comment would be useful.

huwb (commented):
looks great, just some nitpicks

would be awesome to have a submarine demo at some point. i guess when the camera enters the submarine the mode would change from fullscreen to mesh?

could you summarise what the perf impact is - i think theres just a one or two new textures created? are they always created or just when mesh mode is being used?

daleeidd:
> Wow looks great!

Thanks!
 
> If waves are all of a similar wavelength, they'll sort of have a focal point - a depth below the surface where the rays tend to be converged and generate the brightest sharpest caustics. So this feature makes them blurrier away from that depth.
> 
> It's a gross approximation and happy to try with it torn out. Maybe just fading the intensity based on depth looks better.

Oops... I used the wrong term. I meant scene Z which I don't understand.

huwb:
> Oops... I used the wrong term. I meant scene Z which I don't understand.

Ah ok, im not sure i'm afraid, guess i have forgotten :/. Perhaps a blame/history check digs up some info.

daleeidd:
> would be awesome to have a submarine demo at some point. i guess when the camera enters the submarine the mode would change from fullscreen to mesh?

Indeed. Yeah that makes the most sense for an FPS game.
 
> could you summarise what the perf impact is - i think theres just a one or two new textures created? are they always created or just when mesh mode is being used?

Done. I've added a performance section to the post.

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.cs
daleeidd: I think so. I am looking at replacing the mask with a stencil buffer completely but will wait for 2021.2. 2021.2 has some new APIs which make it a little easier to use.

huwb: sure sounds good

huwb:
> Indeed. Yeah that makes the most sense for an FPS game.

probably going outside the scope now but if the camera was outside the sub, does this tech also allow a volume of air to be carved out of water? just asking out of interest, probably a dumb question. not a request for extending this pr.

> I've added a performance section to the post.

could you add the bpp for each of those texture targets? otherwise that looks good.

daleeidd:
> > Oops... I used the wrong term. I meant scene Z which I don't understand.
> 
> Ah ok, im not sure i'm afraid, guess i have forgotten :/. Perhaps a blame/history check digs up some info.

It was in the comment. It's just a vanilla mip selection formula to reduce stretching/dilating of samples. I guess devs can always adjust the DoF property to suit their needs.

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Effect.cs
daleeidd: Not really as it is before the underwater is applied. It serves the same purpose as the color texture Unity provides in HDRP/URP. I've renamed it as per other textures: _CrestCameraColorTexture.

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Effect.cs
daleeidd: I have solved the issues in the comment and removed. Some code there was leftovers from trying to get CopyTexture and MSAA working which I think is a Unity bug. Going to look again at getting CopyTexture and MSAA working in 2021.2.

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Mask.cs
daleeidd: Happy to do a rename but will do it throughout to make it consistent. Prefer to do it now rather than later. I'm thinking about renaming "Water Volume Boundary Geometry" to just "Volume Geometry". Might keep the Mode enum the same but could be renamed to something like:

```csharp
public enum Mode
{
    FullScreen,
    Volume2D,
    Volume3D,
    VolumeFilled,
}
```

or 

```csharp
public enum Mode
{
    FullScreen,
    Portal,
    Volume,
    VolumeFilled,
}
```

LMKWYT

huwb: ok sounds good. i like the latter enum! only i forgot what VolumeFilled is? (prob the name is fine but a comment above each one might be useful)

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Effect.cs
huwb: ok sure

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Mask.cs
daleeidd: VolumeFilled is a fly-through volume. I guess I should name it VolumeFlyThrough :^)

huwb: im really sorry but i dont know what a fly through volume is either? im sure it will click in a second :)

daleeidd: It's when you can fly through it and still see the effect. Borrowed from fly through clouds. The "Volume" one (third option) will disappear once camera enters it as it doesn't execute the extra passes for the interior or require a stencil buffer.

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Mask.cs
```diff
@@ -124,54 +179,79 @@ internal static void SetUpMaskTextures(CommandBuffer buffer, RenderTextureDescri
...
+            SetUpBoundary(_oceanMaskMaterial.material);
+            SetUpMaskTextures(descriptor);
 
-            SetInverseViewProjectionMatrix(_oceanMaskMaterial.material);
+            // Populate water boundary before mask so we can use the stencil.
+            if (_mode != Mode.FullScreen)
+            {
+                SetUpBoundaryTextures(descriptor);
+                PopulateBoundary(_oceanMaskCommandBuffer, _boundaryFrontFaceTarget, _boundaryBackFaceTarget);
+                if (IsStencilBufferRequired)
```
daleeidd, line 245: Just realised I should be using the stencil buffer here regardless as it rejects a lot of the mask processing.

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.cs
daleeidd: Actually yes... I should be using the stencil buffer for the mask for all volume modes except full screen (which I am not). I was thinking of using the stencil for the full screen effect itself in my reply. Will expand its usage.

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Mask.cs
daleeidd: Done.

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Mask.cs
daleeidd: Done. Let me know if you can think of a better name for volume fly through if you think it is confusing.

daleeidd:
@huwb updated post with performance info. Addressed review items. Just need to remove test data which I can do before merging.

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Mask.cs
huwb: i wont lie - its still bending my mind each time i revisit, but i cant think of better names either. i think a comment explaining what teach one does would be useful if possible? then i htink its good with the current names.

huwb (approved):
LGTM

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Mask.cs
daleeidd: Heh... Sure. Done :^)

daleeidd:
Thanks! Also added an invert culling option last minute for the Portal option. Very useful.

PR #961, opened 2021-12-21
Allow any renderer as Ocean Input

daleeidd:
Validation only allowed MeshRenderer to be used as an Ocean Input. Now anything deriving from Renderer (include particles) can be used. Also changes the behaviour of "disabled renderer" for trails and lines to only disable the rendering so the trails continue to update.

Foam and TrailRenderer go really well together.

The downside to this PR is that it could be a breaking change if someone has multiple renderers on the same GO. I've added a help box to the top of the component to inform the user of what renderer it is using.

Also, they aren't straightforward to use as the are quads so they need to be aligned the right way. But we have that issue with mesh renderer too.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
@@ -58,7 +58,7 @@ namespace Crest
     {
 #if UNITY_EDITOR
         [SerializeField, Tooltip("Check that the shader applied to this object matches the input type (so e.g. an Animated Waves input object has an Animated Waves input shader.")]
-        [Predicated(typeof(MeshRenderer)), DecoratedField]
+        [Predicated(typeof(Renderer)), DecoratedField]
         bool _checkShaderName = true;
 #endif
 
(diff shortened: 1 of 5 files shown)


huwb:
wow i was literally thinking about this last night. using particles with water inputs can lead to all sorts of things

huwb (approved):
Looks great to me

Btw for meshes, i think we were hoping to at some point change to drawing MeshFilter's rather than MeshRenderers, as theres probably a fair bit of time getting spent in MeshRenderer setup. do you think theres still a way to draw meshes directly in the future? obv DrawMesh would not support particle systems so wondering if there can be multiple paths.

daleeidd:
> Looks great to me

Thanks

> Btw for meshes, i think we were hoping to at some point change to drawing MeshFilter's rather than MeshRenderers, as theres probably a fair bit of time getting spent in MeshRenderer setup. do you think theres still a way to draw meshes directly in the future? obv DrawMesh would not support particle systems so wondering if there can be multiple paths.

Indeed. Worst case scenario we could add a "mode" setting which could default to whatever it finds that makes sense (spline, mesh filter, mesh renderer)

huwb:
Yeah it I guess maybe it could detect at some level if it's a MeshRenderer and if so draw its mesh directly. So could probably be automatic, just need to see for the code comes out and if it's cheaper. Good to know anyway

PR #964, opened 2021-12-26
Add scrollbar to Ocean Debug GUI

daleeidd:
Adds a scrollbar to the _Ocean Debug GUI_ when the _Draw LOD Datas Actual Size_ is used. Scrollbar is appears on the right and scrolls all LODs simultaneously.


huwb (approved):
Good idea

daleeidd:
Thanks!

PR #965, opened 2021-12-26
Add more debug visualisers

daleeidd:
Adds more debug visualisers for visualising the LOD data on the ocean surface. It also merges all of the visualiser options into one dropdown (highlighted in yellow).


huwb (approved):
Great stuff

daleeidd:
Thanks!

PR #966, opened 2021-12-27
Add TiledTexture shader struct

daleeidd:
Refactors our tiled (foam, caustics etc) texture usage.

Key points:
- creates a TiledTexture struct which holds all of the relevant data
- migrates from tex2D and friends to Texture2D
- will be further expanded in #951 which prompted this refactor
- uses namespaces (prompted by #943) and statics

I will be adding functions to the struct for handling floating origin. Not using a struct has become clunky.

As for namespaces, #943 means that our code is included in other people's shaders which could mean naming conflicts. We could just prefix everything but this is much nicer. It looks verbose right because I haven't added the namespace everywhere which would remove the need WaveHarmonic::Crest everywhere. Will do that in another PR.

A final note is that there is proxy methods for Sample and SampleLevel. From what I read, all functions get inlined so this should have zero cost.

LMKWYT!

### DIFF
--- a/crest/Assets/Crest/Crest/Shaders/Ocean.shader
+++ b/crest/Assets/Crest/Crest/Shaders/Ocean.shader
@@ -266,6 +266,8 @@ Shader "Crest/Ocean"
 			#include "Helpers/BIRP/Core.hlsl"
 			#include "Helpers/BIRP/InputsDriven.hlsl"
 
+			#include "ShaderLibrary/Common.hlsl"
+
 			#include "OceanGlobals.hlsl"
 			#include "OceanInputsDriven.hlsl"
 			#include "OceanShaderData.hlsl"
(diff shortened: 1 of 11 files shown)


huwb (approved):
nice. makes sense to me. i didnt even know namespaces were a thing in hlsl. assume this will work on all platforms?

daleeidd:
Not 100% sure. It works on Metal, DX11 and Vulkan from testing. If I look at the compiled code, all of the abstractions are compiled out. That leaves PSSL (PS5 etc) as I cannot confirm myself. I assume the same will happen and it will work for PSSL but there is a risk it won't because it is a lesser known feature. Not much information out there.

I will remove it for now and try adding it in 2022. That way we are not caught out and have to patch it.

huwb:
I hear that namespace's are fine in PSSL so lets leave it in!

daleeidd:
Thanks!

PR #969, opened 2022-01-02
Specialise spectrum editor for different wave components

huwb:
I started implementing the gravity multiplier as a time scale but its not the same thing - wave speed is proportional to the square root of gravity, so while it does depend on gravity, its not a simple time scale.

There are other things in the spectrum editor that arent relevant / dont work on fft, like wave direction variance and all the advanced controls and it sucks that they are exposed.

So i decided to look at supporting a generic argument that the Embedded attribute can pass to the editor, so that the editor could change fields for Gerstner vs FFT. Not sure how this scores on beauty/simplicity, but it is a general-ish feature and i think the UX is much better. Let me know what you think?

It does not disturb existing embedded editors - they dont require change and if the functionality is not needed, the usage of the Embedded attribute remains the same as previously.


crest/Assets/Crest/Crest/Scripts/Helpers/Attributes/EmbeddedFieldAttribute.cs
```diff
@@ -15,20 +15,29 @@ public class EmbeddedAttribute : DecoratedPropertyAttribute
     {
 #if UNITY_EDITOR
         internal EmbeddedAssetEditor editor;
+
+        // Generic argument that can be passed to embedded editor
+        int _argument;
```
daleeidd, line 20: Would this be better as a Type instead of int?

daleeidd:
Yeah, this works for now with the least hassles. Nice.

Is the plan to move more things into the spectrum (like Turbulence) using this approach, or perhaps the reverse?

What about the case where the spectrum is edited in a standalone inspector? Right now it shows only FFT. Should it show everything? 

Maybe down the line we could make things more generic. Turbulence and Variance could be the same property if variance is normalised to 0-1. That would be a fair bit of work migration wise so it would have to be a breaking change.

crest/Assets/Crest/Crest/Scripts/Helpers/Editor/EmbeddedAssetHelpers.cs
```diff
@@ -200,6 +211,10 @@ public void UpdateEditor(SerializedProperty property)
             if (m_Editor == null && target != null)
             {
                 m_Editor = Editor.CreateEditor(target);
+
+                // Pass through argument for editors that receive it
+                (m_Editor as IEditorWithArg)?.SetArgument(argument);
```
daleeidd, line 216: I think we could change this to accepting a type. And then getting the type like so: `property.serializedObject.targetObject.GetType()`. And maybe rename IEditorWithArg to IEmbeddable. And SetArgument to SetEmbedderType? That way it is more clear what the pattern is for. If this whole thing is temporary then we can leave as is but would be nice if it was more clear if it does stick around. LMKWYT.

huwb: ah yes much better! one

huwb:
oh yeah good shout! done now, much better

> Is the plan to move more things into the spectrum (like Turbulence) 

yeah would be great but i believe thats a breaking change. ill add to the list for 2022?

> What about the case where the spectrum is edited in a standalone inspector

yeah in general i guess it should show everything. one could add a "_isEmbedded" bool to help with that kind of stuff. however here i'd vote for sticking to FFT as it is the primary wave system. let me know if this sounds ok.

daleeidd:
> > Is the plan to move more things into the spectrum (like Turbulence)
> 
> yeah would be great but i believe thats a breaking change. ill add to the list for 2022? migration for this feels like a can of worms.

Agreed.

> > What about the case where the spectrum is edited in a standalone inspector

> yeah in general i guess it should show everything. one could add a "_isEmbedded" bool to help with that kind of stuff. however here i'd vote for sticking to FFT as it is the primary wave system. let me know if this sounds ok.

True... Could add a help box when not embedded letting the user know it is only showing FFT options and should be edited in a ShapeGerstner if wanting Gestner options? Or even a "mode" enum when not embedded to show FFT or Gerstner options. Or only show common options and show help box to edit in a Shape component to see specific options. Mainly so people aren't confused.

huwb:
Done now, let me know if it looks ok

crest/Assets/Crest/Crest/Scripts/Shapes/OceanWaveSpectrum.cs
```diff
@@ -444,6 +468,16 @@ public override void OnInspectorGUI()
...
+            // Display a notice if its being edited as a standalone asset (not embedded in a component) because
+            // it displays the FFT-interface.
+            if (_hostComponentType == null)
+            {
+                EditorGUILayout.Space();
+                EditorGUILayout.HelpBox("This editor is displaying the FFT spectrum settings. " +
+                    "To edit settings specific to the ShapeGerstner component, assign this asset to a ShapeGerstner component " +
+                    "and edit it there by expanding the Spectrum field.", MessageType.Info);
+            }
+
```
daleeidd, line 480: This would make more sense at the top of the inspector which would be consistent with "This is a shared asset." message when embedded.

daleeidd (approved):
Nice. Just one minor thing.

crest/Assets/Crest/Crest/Scripts/Shapes/OceanWaveSpectrum.cs
huwb: done

huwb:
Thanks, merging

crest/Assets/Crest/Crest/Scripts/Shapes/OceanWaveSpectrum.cs
daleeidd: Thanks

PR #970, opened 2022-01-04
Underwater rendering in Edit Mode

daleeidd:
Adds optional (enabled by default) underwater rendering to scene and preview cameras.

I've implemented it in the simplest way without impacting the standalone player. I believe it will be even simpler for SRPs.

It has some drawbacks:
- RT reference is shared but the RTs aren't due to different RT requirements. So RTs are recreated every frame each view is open (not an issue for HDRP now or URP in 2022)
- Disabled underwater being disabled when high enough above surface in edit mode
- Disable fog extinction culling in edit mode

It seems reliable which is surprising as adding edit mode support normally introduces many issues and complexities.

### DIFF
--- a/docs/about/history.rst
+++ b/docs/about/history.rst
@@ -25,6 +25,8 @@ Changed
 .. bullet_list::
 
    -  Added new CPU-based collision provider - *Baked FFT Data*.
+   -  Add support for rendering in edit mode (camera preview and scene view) to *Underwater Renderer*.
+      It can be enabled/disabled with the fog scene view toggle.
    -  Add *CREST_OCEAN* scripting defines symbol.
    -  Add *Depth Fog Density Factor* to *Underwater Renderer* which can be used to decrease underwater fog intensity when underwater.
       Greatly improves shadows at shorelines.
--- a/crest/Assets/Crest/Crest/Scripts/Helpers/Editor/EditorHelpers.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Helpers/Editor/EditorHelpers.cs
@@ -53,6 +53,21 @@ namespace Crest.EditorHelpers
                 UnityEditorInternal.InternalEditorUtility.layers);
             return UnityEditorInternal.InternalEditorUtility.ConcatenatedLayersMaskToLayerMask(temporary);
         }
+
+        /// <summary>Attempts to get the scene view this camera is rendering.</summary>
+        /// <returns>The scene view or null if not found.</returns>
+        public static SceneView GetSceneViewFromSceneCamera(Camera camera)
+        {
+            foreach (SceneView sceneView in SceneView.sceneViews)
+            {
+                if (sceneView.camera == camera)
+                {
+                    return sceneView;
+                }
+            }
+
+            return null;
+        }
     }
 }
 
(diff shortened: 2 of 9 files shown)


huwb:
> So RTs are recreated every frame each view is open

doesnt sound like a blocker esp if resolved in 2022 (is it resolved in birp 2022 btw? also not a blocker but just checking). just to check - it doesnt leak memory right? and is this edit mode only?

i think the other tradeoffs sound fine too

huwb (commented):
looks great generally - nice work! this is a huge upgrade

just to check, is there a way to also add an enable flag to the scene view toggles (where Animated Materials sits)? not a blocker, but just thinking that an artist would tend to look there i would think.

having said that, there is still the Layers dropdown with Water - does it turn off when the Water layer is disabled?

daleeidd:
> looks great generally - nice work! this is a huge upgrade

Thanks.

> just to check, is there a way to also add an enable flag to the scene view toggles (where Animated Materials sits)? not a blocker, but just thinking that an artist would tend to look there i would think.

Nope... but we could reuse the toggles already there? There is Fog and Post-Processing... Fog makes the most sense to me. I'd keep the clipping enabled though.

> having said that, there is still the Layers dropdown with Water - does it turn off when the Water layer is disabled?

It doesn't. I can do that as a separate PR as it is a general issue.

huwb:
> Nope... but we could reuse the toggles already there?

Sounds good to me

daleeidd:
All done including documentation. I have removed that toggle and used the fog scene view toggle instead. The clipping will still work with fog disabled. All rendering is skipped if the ocean proxy plane is used.

daleeidd:
> doesnt sound like a blocker esp if resolved in 2022 (is it resolved in birp 2022 btw? also not a blocker but just checking).

Missed this one. URP 2022 was Unity 2022.1 specifically. But I think that is moot as hopefully Unity will have fixed the problem with the temporary RTs by then which will solve it for both BIRP and URP. I don't know when that will be.

> it doesnt leak memory right?

I'm using Release so it shouldn't. I'll test it to be sure. It wasn't an extra path I had to code. It is already developed to recreate RTs on certain RT descriptor changes like resizing.

> and is this edit mode only?

Nope. It doesn't happen in builds though.

daleeidd:
> it doesnt leak memory right?

It doesn't look like it. The texture count keeps going up and the references aren't cleared which looks pretty scary in the profiler, but they are all zero bytes and memory usage looks okay for RTs. It hangs around 400mb. Total memory looks like it goes up a little bit over but not totally sure and not sure if related. I guess I will have to do a deep dive in memory profiling...

docs/about/history.rst
```diff
@@ -25,6 +25,8 @@ Changed
 .. bullet_list::
 
    -  Added new CPU-based collision provider - *Baked FFT Data*.
+   -  Add support for rendering in scene view to *Underwater Renderer*.
```
huwb, line 28: Not sure if also worth saying "edit mode". Up to you

huwb (approved):
Great stuff!

docs/about/history.rst
daleeidd: Sure. Done.

daleeidd:
Fix those reference leaks by destroying the texture after releasing them.

huwb (approved):
Lgtm!

PR #972, opened 2022-01-05
Fix Underwater not respecting Ocean layer

daleeidd:
Disables underwater if ocean layer is not being rendered.


huwb (approved):
Nice

daleeidd:
Thanks!

PR #973, opened 2022-01-05
Underwater Renderer validate Ocean Renderer Material

daleeidd:
Adds validation for underwater keyword and cull mode.


huwb (approved):
Cool this will be super helpful

Just checking these are earnings, not errors?

Happy either way

daleeidd:
Thanks. They're warnings.

PR #974, opened 2022-01-11
Foam sim reads dynamic wave sim as foam source

huwb:


huwb:
Putting on ice as we got pretty good results with latest wake tech

PR #975, opened 2022-01-11
Reuse RT reference in Underwater Renderer

daleeidd:
Before we were destroying/recreating which was wasteful. This PR avoids that by just calling Release.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Mask.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Mask.cs
@@ -34,40 +34,16 @@ namespace Crest
             BackFace,
         }
 
-        internal RenderTargetIdentifier _maskTarget = new RenderTargetIdentifier
-        (
-            sp_CrestOceanMaskTexture,
-            mipLevel: 0,
-            CubemapFace.Unknown,
-            depthSlice: -1 // Bind all XR slices.
-        );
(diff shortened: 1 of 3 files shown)


crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Mask.cs
```diff
@@ -162,61 +147,38 @@ internal void SetUpMaskTextures(RenderTextureDescriptor descriptor)
...
+            }
+
             descriptor.msaaSamples = 1;
             descriptor.colorFormat = RenderTextureFormat.Depth;
             descriptor.depthBufferBits = 24;
 
-            Helpers.CreateRenderTargetTexture(ref _volumeFrontFaceRT, ref _volumeFrontFaceTarget, descriptor);
-            _volumeFrontFaceRT.name = "_CrestVolumeFrontFaceTexture";
+            _volumeFrontFaceRT.Release();
+            _volumeFrontFaceRT.descriptor = descriptor;
```
huwb, line 173: Do you not have to call create, or does that happen elsewhere?

huwb (approved):
Just that one question but I'll approve to not hold it up

crest/Assets/Crest/Crest/Scripts/Underwater/UnderwaterRenderer.Mask.cs
daleeidd: We don't need to. It is done automatically when we use it with SetRenderTarget.

PR #976, opened 2022-01-14
For scale changes, pull in 0.0 values at cascade start/end

huwb:
WIthout this fix, when scale is e.g. reduced, lod0 was getting populated with the previous lod0 data. This essentially dupes lod0 (lod0 data from previous frame ends up in both lod0 and lod1 in current frame), creating a pop and large displacement.


daleeidd (approved):
Looks good.

PR #977, opened 2022-01-15
Fix few missing things for UnderwaterRenderer and WaterBody material

daleeidd:
Forgot to implement:
- Underwater tile culling
- Underwater depth fog density multiplier

No release notes as this completes incomplete feature that has yet to be released.


huwb (approved):
Good catch

daleeidd:
Thanks.

PR #979, opened 2022-01-15
SphereWaterInteraction improved wakes

huwb:
Current version pushes water up ahead of sphere, and pulls down behind sphere, and does nothing inside the sphere. This results in water raising up too high within the sphere, and then being aggressively pulled back down behind the sphere, leaving a big trailing indentation in the water. Together these things caused multiple problems such as badly impacting floating physics, and large amplitude bumps/indentations play badly with LODing.

To combat this behaviour, an "inner sphere" is modelled slightly inside the primary sphere which does the inverse. Ahead of this sphere the water is pulled back down before it raises too high, and the combination of the action of the two spheres at the leading edge creates a sharp wave front which is good for generating foam and will be good for keying FX off. At the trailing edge, pushing the water up before pulling down at the outer sphere seems to make a good wake shape, with a bump following the object, and a general shape and foam pattern which matches reference pretty well.

Along the way i found another artifact that was generating rings - there is a feedback loop between the injected dynamic wave displacement and the "displacement correction" feature of our ocean inputs, which made the sphere interaction tend to apply in groups making distinct circles. This is more pronounced for smaller radii spheres (like 0.5m). I dont think theres a perfect solution to this, but I could combat it by adding a param to control how much displacement correctoin applies. I found that a little bit of feedback actually adds desirable noise/distortion to an otherwise regular wake so this turned out to be a useful control.

I tweaked the existing boats to work, but didnt touch OceanLiner. I suppose this needs to be removed in favour of the new boat FBX. Will look at that later.

I added a new wakerig.unity scene (without prefabs atm) but I think i should probably remove this and the wake tests into boat.unity or threeboats.unity. If you have thoughts let me know @daleeidd . As you've mentioned before we should plan/consolidate scenes for the next version of crest as boat and threeboats are redundant and could be unified I thiink.

Finally while checking existing examples/boats I retweaked a few bits in the existing scenes to try to get a more realistic result.


daleeidd:
Awesome work. Heaps better.

It looks like the sim is a lot slower than before. Very noticeable with the ripple generator in threeboats.unity. Is this expected? Otherwise so far it has all been really good - even spheres dropping in water look better from memory.

> I added a new wakerig.unity scene (without prefabs atm) but I think i should probably remove this and the wake tests into boat.unity or threeboats.unity. If you have thoughts let me know @daleeidd . As you've mentioned before we should plan/consolidate scenes for the next version of crest as boat and threeboats are redundant and could be unified I thiink.

We could add the wake scene and then reconsolidate next version. I'm thinking those other scenes will either be updated or deleted. The new scene does a good job demonstrating the wakes but up to you how you want to approach it.

Would you be open to capitalised filenames? Maybe just BoatWakes.unity or Wakes.unity for the new one if you run with it?

> I tweaked the existing boats to work, but didnt touch OceanLiner. I suppose this needs to be removed in favour of the new boat FBX. Will look at that later.

Are we allowed to edit the FBX? I think it needs some specular/metallic and smoothness. It has numerous sub meshes with their own material so I don't really want to do that in Unity. It should be done in Blender or something.

huwb:
> It looks like the sim is a lot slower than before

Yeah i've set gravity multiplier to 1 - so realistic. The ripple does indeed suck. I have retweaked it so its not completely broken. It still looks very wrong compared to what it was. But it is accurate im fairly sure, its just the scene/camera/scale are weird. Also the bump does not add in a multi-res compatible way like the SWI does. I should probably do a pass on it at some point to try to improve it.

> Would you be open to capitalised filenames

Ok done now. It's not prefab'd, not sure if it's worth it / required. LMK what you think.

Re the fbx, lets ask on discord

huwb:
Besides the FBX question i think this is ready for review @daleeidd

crest/Assets/Crest/Crest-Examples/Shared/Scripts/SimpleMotion.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Crest Ocean System
```
daleeidd, line 1: Could it be added to Crest.Examples namespace?

crest/Assets/Crest/Crest-Examples/BoatDev/Scenes/threeboats.unity
```diff
@@ -38,7 +38,7 @@ RenderSettings:
   m_ReflectionIntensity: 1
   m_CustomReflection: {fileID: 0}
   m_Sun: {fileID: 0}
-  m_IndirectSpecularColor: {r: 0.40657905, g: 0.5909717, b: 0.9444369, a: 1}
+  m_IndirectSpecularColor: {r: 0.40652782, g: 0.59053, b: 0.9443165, a: 1}
```
daleeidd, line 41: Should be reverted.

crest/Assets/Crest/Crest/Scripts/Interaction/SphereWaterInteraction.cs
```diff
@@ -236,7 +260,7 @@ public void Draw(LodDataMgr lodData, CommandBuffer buf, float weight, int isTran
             if (_debugSubsteps)
             {
                 var col = 0.7f * (Time.frameCount % 2 == 1 ? Color.green : Color.red);
-                var pos = transform.position + /*(fixup ? 1f : 0f) **/ -_velocity * timeBeforeCurrentTime;
+                var pos = transform.position + /*(fixup ? 1f : 0f) **/ -_velocity * (timeBeforeCurrentTime - _velocityOffset);
```
daleeidd, line 263: Does something need to happen with this?

daleeidd:
> Ok done now. It's not prefab'd, not sure if it's worth it / required. LMK what you think.

It is. It's pretty simple to do. Here's the steps:

1. Place everything under single parent GO
2. Drag parent GO into file view to create prefab
3. Rename prefab with core suffix
3. Right click prefab and create prefab variant
4. Rename variant and remove core suffix
5. Delete parent GO in scene
6. Drag in prefab _**variant**_
7. Done... prefabs go in _Internal_ folder (see others)

daleeidd (changes requested):
Looks great. Ocean material changes are nice too. Just a few things and BoatWakes need prefabbing.

huwb:
Prefab'd

> But LMKWYT about it being breaking change

Yeah that sounds fine to me

crest/Assets/Crest/Crest-Examples/BoatDev/Scenes/boat.unity
```diff
@@ -131,6 +131,106 @@ PrefabInstance:
   m_Modification:
     m_TransformParent: {fileID: 0}
     m_Modifications:
+    - target: {fileID: 2876771014875864530, guid: eeab37efd0ddc46a5ba05d96be38bea4,
```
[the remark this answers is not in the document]
huwb: done

crest/Assets/Crest/Crest-Examples/Shared/Scripts/SimpleMotion.cs
huwb: done

crest/Assets/Crest/Crest-Examples/BoatDev/Scenes/threeboats.unity
huwb: done

crest/Assets/Crest/Crest/Scripts/Interaction/SphereWaterInteraction.cs
huwb: cleaned up now

huwb:
Ok i added two release notes. The first one could potentially be moved to abreaking changes section. Or a new note could be added as a breaking change. Maybe the latter but I'm happy for you to decide what makes sense, if you're still up for adding that.

Happy for this to be merged when you are.

Thanks for the feedback and help!

crest/Assets/Crest/Crest-Examples/BoatDev/Data/WavesWakeRig.asset
```diff
@@ -0,0 +1,69 @@
+%YAML 1.1
```
daleeidd, line 1: Missed this one. Probably should be renamed to WavesBoatWakes.

daleeidd (approved):
Nice work. One minor thing. Can be merged.

crest/Assets/Crest/Crest-Examples/BoatDev/Scenes/boat.unity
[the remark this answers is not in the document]
daleeidd: Thanks

crest/Assets/Crest/Crest-Examples/BoatDev/Scenes/threeboats.unity
daleeidd: Thanks

daleeidd:
> Ok i added two release notes. The first one could potentially be moved to abreaking changes section. Or a new note could be added as a breaking change. Maybe the latter but I'm happy for you to decide what makes sense, if you're still up for adding that.

Actually... won't need a breaking change... I was thinking of how the ripple was affect but that was the gravity change which was just a setting for an example scene. My bad.

crest/Assets/Crest/Crest-Examples/BoatDev/Data/WavesWakeRig.asset
huwb: ah thanks done

huwb:
Ok cool merging, thanks again

PR #980, opened 2022-01-18
Fix non Mesh Renderer inputs and add Albedo shader

daleeidd:
I could not get particles to work with Albedo so I had to fix something and fixed a few other things:
- Albedo shader as most other shaders are multi-pass and we render all passes with DrawRenderer
- Render all sub-meshes so we can support particle trails
- Add a weird fix (Camera.SetupCurrent) to stop some things from flickering. I was given this by someone trying to get particles working for HDRP so looks like we need it anyway. Not sure what it does exactly but I think it sets up Camera.current.

Particles on surface in Demo scene.

This should work for SRPs too. But can always come back and modify or create another shader there.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/BuildCommandBuffer.cs
+++ b/crest/Assets/Crest/Crest/Scripts/BuildCommandBuffer.cs
@@ -111,6 +111,12 @@ namespace Crest
 
             BuildLodData(OceanRenderer.Instance, _buf);
 
+            if (OceanRenderer.Instance.ViewCamera != null)
+            {
+                // Fixes flickering non mesh renderer renderers (like particles). Method is undocumented.
+                Camera.SetupCurrent(OceanRenderer.Instance.ViewCamera);
+            }
+
             // This will execute at the beginning of the frame before the graphics queue
             Graphics.ExecuteCommandBuffer(_buf);
 
--- a/crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
@@ -148,7 +148,15 @@ namespace Crest
                     buf.SetGlobalVector(sp_DisplacementAtInputPosition, Vector3.zero);
                 }
 
-                buf.DrawRenderer(_renderer, _material);
+                for (var i = 0; i < _renderer.sharedMaterials.Length; i++)
+                {
+                    // Empty material slots is a user error. Unity complains about it so we should too.
+                    Debug.AssertFormat(_renderer.sharedMaterials[i] != null, _renderer,
+                        "Crest: {0} has empty material slots. Remove these slots or fill them with a material.", _renderer);
+
+                    // Submesh count generally must equal number of materials.
+                    buf.DrawRenderer(_renderer, _renderer.sharedMaterials[i], submeshIndex: i);
+                }
             }
         }
 
--- a/crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
@@ -154,8 +154,10 @@ namespace Crest
                     Debug.AssertFormat(_renderer.sharedMaterials[i] != null, _renderer,
                         "Crest: {0} has empty material slots. Remove these slots or fill them with a material.", _renderer);
 
+                    // By default, shaderPass is -1 which is all passes. Shader Graph will produce multi-pass shaders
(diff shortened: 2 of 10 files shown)


crest/Assets/Crest/Crest/Scripts/BuildCommandBuffer.cs
```diff
@@ -111,6 +111,12 @@ public void BuildAndExecute()
 
             BuildLodData(OceanRenderer.Instance, _buf);
 
+            if (OceanRenderer.Instance.ViewCamera != null)
+            {
+                // Fixes glitching trail renderers. Method is undocumented.
+                Camera.SetupCurrent(OceanRenderer.Instance.ViewCamera);
```
huwb, line 117: There isn't a big CPU hot from this is there? Just checking

huwb (approved):
Nice work. Never would have known the flickering fix

crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
```diff
@@ -148,7 +148,17 @@ public virtual void Draw(LodDataMgr lodData, CommandBuffer buf, float weight, in
                     buf.SetGlobalVector(sp_DisplacementAtInputPosition, Vector3.zero);
                 }
 
-                buf.DrawRenderer(_renderer, _material);
+                for (var i = 0; i < _renderer.sharedMaterials.Length; i++)
+                {
```
huwb, line 152: Out of interest, why doesn't DrawRenderer do all this automatically?

daleeidd: I have no idea. No one has questioned it from what I can see. I did a test to see if it is the same for mesh renderers and it is.

daleeidd: I'm thinking of changing null/continue to an assertion as empty slots is a problem. How does that sound?

crest/Assets/Crest/Crest/Scripts/BuildCommandBuffer.cs
daleeidd: About 0.01 to 0.03ms on 2014 MacBook Pro (pretty weak).

huwb: Ok

crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
huwb: Is it a problem? If so then sounds fine to me

daleeidd: Yeah. Done. Thanks

daleeidd:
Alrighty... I made a somewhat breaking change. Before we were rendering all passes in sequence for ocean inputs but now I have changed it to only rendering the first pass. This is compatible with all of our shaders as they are only a single pass and should be fine with most of everyone else's. The reason for the change is to make compatible with URP SG unlit which is nice to have to be able to use SG with our inputs. Unfortunately, HDRP is a big dumb and has the shadow caster pass first even for unlit. I've only checked this for 2020 and 2021. 2021 BIRP SG has forward base as first pass. LMKWYT.

huwb:
Erm yeah i doubt multi-pass is needed. Actually doesnt the "remove convex hull" shader use multiple passes? Might be remembering that wrong.

Not sure if its worth a validation warning for shaders that do have multiple passes. Maybe not.

I didnt understand the outlook for HDRP. Is that a blocker?

Happy for you to go with what makes sense

daleeidd:
> Erm yeah i doubt multi-pass is needed. Actually doesnt the "remove convex hull" shader use multiple passes? Might be remembering that wrong.

It does but the Draw method is overridden for the sim so it is unaffected by this change.

> Not sure if its worth a validation warning for shaders that do have multiple passes. Maybe not. Actually maybe this is required, because it seems theres a lot of fail cases - URP lit SG, HDRP SGs, normal shaders with multiple passes.

Good idea. I've added validation.

I looked into adding a shader pass selector (stores the shader pass index) which I thought would be a good idea, but it gets complicated with sub-shaders as we'd have to have a pass for each sub-shader to do it correctly.

> I didnt understand the outlook for HDRP. Is that a blocker?

Not a blocker as not a new issue. Just means HDRP SG cannot be used as an input as the shader pass at index zero is for shadows which is useless.

daleeidd:
@huwb pinging just in case you missed it.

huwb:
> Just means HDRP SG cannot be used as an input...

Ok makes sense

daleeidd:
Thanks.

PR #981, opened 2022-01-19
Fix Underwater Renderer not using sun when other light shining on transparent object

daleeidd:
Unity doesn't set up lighting for us when using command buffers to draw things. When rendering the underwater effect, we were getting the last values which often will be the sun. But if there is another light which shines on a transparent object, and said transparent object is in view, then the last values will be from the ForwardAdd passes which is not the sun.


huwb (approved):
Well found

holdingjason:
> Well found

Thanks and appreciate you all jumping on this so fast.

daleeidd:
Thanks both.

PR #982, opened 2022-01-27
Fix caustics no rendering if shadow data is disabled

daleeidd:
When shadow data is disabled the SS shadow texture was not set. This PR binds that texture to a white texture if disabled.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/LodData/Shadows/LodDataMgrShadow.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/Shadows/LodDataMgrShadow.cs
@@ -120,6 +120,12 @@ namespace Crest
         {
             base.OnDisable();
 
+            // Built-in RP only.
+            {
+                // Black for shadows. White for unshadowed.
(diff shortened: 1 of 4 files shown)


crest/Assets/Crest/Crest/Scripts/LodData/Shadows/LodDataMgrShadow.cs
```diff
@@ -120,6 +120,11 @@ internal override void OnDisable()
         {
             base.OnDisable();
 
+            {
```
huwb, line 123: Not certain the additional scope is necessary but that's minor

huwb (approved):
Looks good

crest/Assets/Crest/Crest/Scripts/LodData/Shadows/LodDataMgrShadow.cs
daleeidd: It's for adding a built-in renderer conditional downstream. Setting up the scope here reduces conflicts. I'll add a comment.

daleeidd:
Thanks.

PR #984, opened 2022-01-27
Fix "Underwater Renderer > Filter Ocean Data" not being applied to caustics

daleeidd:
Fixes "Underwater Renderer > Filter Ocean Data" option not being applied to caustics. It was using slice index + 1, but the cascade data had the filter applied so by default it would be much higher.


huwb (approved):
Nice one

daleeidd:
Thanks.

PR #987, opened 2022-02-12
Control over force underwater

daleeidd:
Screenshot (392)

My previous fix #986 was a bit shortsighted and so was a subsequent fix 59fb1b623e70b0cc7e677dff16f1c74c75ef83ab.

This PR gives a reasonable level of control of the __ForceUnderwater_ variable.

This control is only really needed when using the portals/volumes feature as it can easily expose the underside of the surface and is even worse in combination with varying water levels.

_Automatic_ just enables/disables __ForceUnderwater_ if portals/volumes feature is used which is a sane default.

I've added a not-so-nice method for UnityEvent support as they do not support Enums :^(

LMKWYT.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/LodData/RegisterHeightInput.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/RegisterHeightInput.cs
@@ -40,21 +40,6 @@ namespace Crest
         [SerializeField, Tooltip("Inform ocean how much this input will displace the ocean surface vertically. This is used to set bounding box heights for the ocean tiles.")]
         float _maxDisplacementVertical = 0f;
 
-        static int s_Count;
-        internal static int Count => s_Count;
-
-        protected override void OnEnable()
-        {
-            base.OnEnable();
-            s_Count++;
-        }
-
-        protected override void OnDisable()
-        {
-            base.OnDisable();
-            s_Count--;
-        }
-
         protected override void Update()
         {
             base.Update();
--- a/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
+++ b/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
@@ -185,6 +185,17 @@ namespace Crest
         [Tooltip("Whether 'Water Body' components will cull the ocean tiles. Disable if you want to use the 'Water Body' 'Material Override' feature and still have an ocean.")]
         public bool _waterBodyCulling = true;
 
+        public enum SurfaceSelfIntersectionFixMode
+        {
+            Off,
+            On,
+            Automatic,
+        }
+
+        [SerializeField]
+        [Tooltip("How Crest should handle self-intersections of the ocean surface caused by choppy waves which can cause a flipped underwater effect. Automatic will enable/disable the fix if portals/volumes are used.")]
+        SurfaceSelfIntersectionFixMode _surfaceSelfIntersectionFixMode = SurfaceSelfIntersectionFixMode.Automatic;
+
 
         [Header("Detail Params")]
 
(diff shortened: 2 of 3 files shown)


huwb:
I'm not sure how to use this new option myself, so am unsure about exposing it to users.

Would it be possible to get a summary of the problem which i'm understanding has multiple facets/cases? Then we can iterate on how this is exposed to the user. I think they need to be told exactly what to do with this setting.

huwb (commented):
Commented above

daleeidd:
That's fair. It is tricky.

Firstly, if the ocean surface is contiguous, then ForceUnderwater can be used safely. It only becomes and issue when it is not like when using a portal or volume and possibly the clip surface.

1_1_SelfIntersecting
**Self-Intersecting:** If it is **Off**, the issue you encountered with self-intersecting waves is present.

1_2_VaryingHeightAW
**Varying Height:** If it is **On**, the issue is that behind the sphere the mask is not clipped (and mustn't be) and so we rely on the enclosing nature of the surface to tell us this is indeed underwater. But here it is forced to above water which flips that. It can be mitigated by expanding the height input so the camera's height above water is the same as the highest point, which would disable it because it would be within 2m from the surface. But hard to cover all cases with that trick. And also hard to explain to users. With https://github.com/wave-harmonic/crest/commit/59fb1b623e70b0cc7e677dff16f1c74c75ef83ab we could handle it automatically, but there could be cases where it is not wanted.

1_3_VaryingHeightBW
**Varying Height:**: Same as above but the inverse.

1_4_BumpAW
**Bump:** If **On**, the the underside becomes incorrectly inverted.

1_5_BumpBW
**Bump:** Same as above but the inverse.

1_6_Waves
**Waves:** If **On**, intense waves can expose the underside and be incorrectly inverted.

An automatic way to handle all these cases would be quite difficult I think. Solving self-intersections in the shader would be ideal if that is possible or acceptable performance wise. It looks like using a Jacobian sort of works, but all that would have to be done in the mask too.

huwb:
I haven't quite understood what are on/off/automatic? Is it that on/off enables/disables the fix that forces underwater on/off based on viewer height. Which breaks for above cases.

Or is on/off "force above"/"force below"? I think on/off should probably be renamed to say what verbosely they do.

So Automatic works for the most part but if there's a volume/clip surface situation, and there are inversions, then it doesn't work? 

If so I think I would vote for defaulting it to automatic and taking this setting off the OR and moving to UR or a settings asset (or putting inside an advanced group on OR). I feel like the combination of having inversions that are a problem and using UR are edge case-y enough to not crowd the OR further.

Thinking more I do feel it should be possible to detect flipped triangles. Each triangle has undisplaced world positions on the verts. I think doing a ddx(worldPos.x) should signal if a triangle is flipped. I'll see if I can try it tonight.

huwb:
> Thinking more I do feel it should be possible to detect flipped triangles

Initial test looks promising.. maybe.. currently only if camera is aligned with x/z world axes, it seems to detect inversions. It will need more work to see if it works for all camera orientations.

Pushed test so far to `experiment/flip-inverted-faces`

The examples scene in that branch looks like the following. There's still the weird discontinuity highlighted. thats actually the beginning of the inversion (but not any inverted faces), which i dont think is fixable, but is hopefully not too obviously broken for normal cases

image

huwb:
Update on detecting inversions - I put in code to make it work for any camera view. Unfortunately doesnt work in some cases still. Need to figure out if math is incomplete or if its something fundamental..

daleeidd:
> I haven't quite understood what are on/off/automatic? Is it that on/off enables/disables the fix that forces underwater on/off based on viewer height. Which breaks for above cases.

Correct. It completely enables/disables the fix (for both above and below water). As if it breaks for one it will break for the other too.

> Or is on/off "force above"/"force below"? I think on/off should probably be renamed to say what verbosely they do.

See above.

> So Automatic works for the most part but if there's a volume/clip surface situation, and there are inversions, then it doesn't work?

Correct. I haven't disabled it for clip surface as I wasn't sure about that one.

> If so I think I would vote for defaulting it to automatic and taking this setting off the OR and moving to UR or a settings asset (or putting inside an advanced group on OR). I feel like the combination of having inversions that are a problem and using UR are edge case-y enough to not crowd the OR further.

Sounds good.

> Pushed test so far to experiment/flip-inverted-faces

Awesome. That works much better so far. Hopefully it pans out.

huwb:
Ok i think i get it all now, thanks for walking me through it

I thought of a simple approach for flipped tris that i thought was going to be awesome but doesnt seem to be - project the undisplaced positions to clip space, and then simply check if the clip space coordinates are flipped. Almost works with the complex math, but still has a fail case:

image

i think its a problem of the triangle not being flipped but curled around rotationally enough so that the coords start going negative.

i tried combining the two approaches but can still see holes sometimes..

daleeidd:
That's unfortunate :^(

I have moved under an _Advanced_ header. I also moved the _Underwater Cull Limit_ under there too.



I also documented both options.

daleeidd:
Thanks.

> I'm not sure if this is already fixed but related to the force option - if i move the scene view camera below the water and press save, it serialises a new Force Underwater value into the mat file and shows up as a diff. I'm testing on master so this may not be relevant once this PR is merged.

Still relevant and now even more likely to happen since we handle above water too. I think the only options of avoid that behaviour is setting it through the MPB or using a global float.

I think global property makes sense since we also using it in the mask. Hard to say which one is more expensive between setting it on every MPB or a global.

huwb:
Yeah i have no idea the cost of globals. I've seen others using them pretty liberally. I think a global is fine if its easiest.

daleeidd:
Thanks! All done and used it to fix self-intersections in the examples.

PR #989, opened 2022-02-22
Document Underwater Renderer shader API

daleeidd:
Now that the feature is done for all RPs I have added documentation. It can be found here: https://crest.readthedocs.io/en/docs-ur-shader-api/user/underwater.html#underwater-shader-api

Let me know if you think it makes sense. I am leaning a bit on the example shaders for the shader setup but that should be fine I think.

### DIFF
--- a/docs/user/underwater.rst
+++ b/docs/user/underwater.rst
@@ -128,6 +128,76 @@ This feature also clips the ocean surface to match.
 A common use case would be a window on a boat.
 
 
+.. _underwater-shader-api:
+
+Underwater Shader API
+^^^^^^^^^^^^^^^^^^^^^
+
+.. admonition:: Preview
+
+   This feature is in preview and may change in the future.
+
+The underwater effect uses opaque depth and thus will not render correctly for transparent objects.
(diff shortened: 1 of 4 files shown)


docs/user/underwater.rst
```diff
@@ -128,6 +128,80 @@ This feature also clips the ocean surface to match.
...
+
+.. only:: urp
+
+   .. tab:: `URP`
+
+      .. include:: includes/_underwater-shader-graph.rst
+
+      For best results using the `URP_Lit_Shader` graph:
+
+      -  Be mindful of what features on the *Shader Graph* you enable as it might affect the underwater fog
```
huwb, line 202: Maybe which features on the shader graph master node

huwb (approved):
Looks good just a minor clarification

docs/user/underwater.rst
daleeidd: I will remove this as on second look there doesn't appear to be any.

daleeidd: Removed

daleeidd:
Thanks!

PR #990, opened 2022-02-23
Fix ShapeFFT popping when enabled with zero weight

daleeidd:
Weight was not being applied to reporting causing scale pop. Reported on Discord by scrappy.


huwb (approved):
Good find

daleeidd:
Thanks!

PR #991, opened 2022-02-24
Fix ShapeGerstner weight not updating with _spectrumFixedAtRuntime

daleeidd:
Removed from scripting as it would get stuck at zero. And the weight was being applied twice which is incorrect.

I was thinking of mentioning the latter as a breaking change but on second thought a slight behaviour change probably doesn't warrant it.

ShapeGerstnerBatched has the same problem but will leave it as deprecated and has CPU queries which I would have to apply the weight too.


huwb (approved):
Ha crazy right was applied twice. boats.unity may have a shapegerstner with a weight between  0-1 maybe. Nice find

daleeidd:
Thanks! I checked and both _Shape*_ components had weight of one.

PR #992, opened 2022-02-24
Fix FFT speed boosting when two or more are active in scene

daleeidd:
If there is more than one ShapeFFT with different resolutions, the __Size_ will be wrong and cause FFT to speedup/jitter. Can also happen with only one ShapeFFT active at a time but one enabled after the other will overwrite it incorrectly.


huwb (approved):
Nice find

daleeidd:
Thanks

PR #993, opened 2022-02-24
Examples: Add audio height events example

daleeidd:
Add a quick audio demo using the _OceanSampleHeightEvents_ component. This sort of thing gets asked a bit and I also don't think many are taking advantage of it.

The sphere goes up and down and audio will be played depending on proximity to surface and whether underwater or not.


huwb (approved):
Ah good idea, nice one

Playing VFX on hitting water could be an option in future too. 

I didn't actually try as I still don't have internet (!). But I'll approve to not hold things up and I will try later.

huwb (approved):
Ah good idea, nice one

Playing VFX on hitting water could be an option in future too. 

I didn't actually try as I still don't have internet (!). But I'll approve to not hold things up and I will try later.

daleeidd:
Thanks!

> Playing VFX on hitting water could be an option in future too.

Indeed.

> I didn't actually try as I still don't have internet (!)

Oof that's a long time D:

PR #994, opened 2022-02-25
Demo: Add lakes/river examples scene

daleeidd:
LakeRiver

Since we won't have the upcoming lake/river scene ready in time, I have adapted [Unity's recent terrain example scene](https://assetstore.unity.com/packages/3d/environments/unity-terrain-urp-demo-scene-213197). This is can be replaced with something better for 2022.

I stripped it bare of any details so it is just the terrain. I have also reduced the resolution (maybe a bit too zealously) of everything to bring it from 4GB to under 200MB. It doesn't look as good but works as an example for now.

Most of the water is at sea level so I added a river that leads up to a mini lake. I've also used a reflection probe instead of planar reflections as the later doesn't support fog which the scene uses nicely.

I need to do a minor cleanup with some asset file naming.

LMKWYT.


daleeidd:
@huwb This is ready for review. I used the wave data points for the spline and referred to the scene in documentation.

daleeidd:
Merging. Any issues post here :^)

PR #996, opened 2022-02-26
Support scaling FFT waves on spline

huwb:
Allow scaling FFT waves on spline (not supported previously).

*SplinePointDataGerstner* has been renamed to *SplinePointDataWaves* which works for both *ShapeFFT* and *ShapeGerstner*.


huwb:
Thanks merged

PR #997, opened 2022-02-28
Maximum Buoyancy Force

daleeidd:
Adds a _Maximum Buoyancy Force_ to clamp buoyancy.

Useful when buoyant objects are submerged a decent depth which can cause them to gain too much force and launch into the stratosphere.

By default, it is set to Infinity (ie no effect) so users will have to opt in.

### DIFF
--- a/crest/Assets/Crest/Crest-Examples/BoatDev/Scripts/BoatAlignNormal.cs
+++ b/crest/Assets/Crest/Crest-Examples/BoatDev/Scripts/BoatAlignNormal.cs
@@ -32,6 +32,8 @@ public class BoatAlignNormal : FloatingObjectBase
     public float _boyancyTorque = 8f;
     [Tooltip("Approximate hydrodynamics of 'surfing' down waves."), Crest.Range(0, 1)]
     public float _accelerateDownhill = 0f;
+    [Tooltip("Clamps the buoyancy force to this value. Useful for handling fully submerged objects. Use Infinity to disable.")]
+    public float _maximumBuoyancyForce = Mathf.Infinity;
 
     [Header("Engine Power")]
     [Tooltip("Vertical offset for where engine force should be applied.")]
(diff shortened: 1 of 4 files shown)


huwb (approved):
Oh yeah good shout

Does typing Infinity into a float field work? In any case thumbs up

Also I thought floating point comparisons with inf always returned false. Or maybe that's for NaNs. Either way if it works that's fine.

daleeidd:
> Oh yeah good shout

Thanks.
 
> Does typing Infinity into a float field work? In any case thumbs up

It does. Can also just type _Inf_.

> Also I thought floating point comparisons with inf always returned false. Or maybe that's for NaNs. Either way if it works that's fine.

Probably NaNs. This works as expected.

crest/Assets/Crest/Crest-Examples/BoatDev/Scripts/BoatAlignNormal.cs
```diff
@@ -32,6 +32,8 @@ public class BoatAlignNormal : FloatingObjectBase
     public float _boyancyTorque = 8f;
     [Tooltip("Approximate hydrodynamics of 'surfing' down waves."), Crest.Range(0, 1)]
     public float _accelerateDownhill = 0f;
+    [Tooltip("Clamps the buoyancy force to this value. Useful for handling fully submerged objects. Use Infinity to disable.")]
```
huwb, line 35: Does the user need to enter 'Inf' or does Infinity work? If the former probably worth giving this as instruction as I would 
not know how to put infinity into a float field.

daleeidd: Either _Inf_ or _Infinity_. _Inf_ will expand to _Infinity_ when field loses focus. The actually value is _Infinity_.

I've change that sentence to "Enter 'Infinity' to disable". Inf is just a shortcut which users can work out for themselves through Unity.

PR #1000, opened 2022-04-01
Fix UR XR in 2021.2 builds

daleeidd:
XR wasn't working for UR because keywords were being stripped as they have changed or keyword system has changed.

I've also brought file up-to-date with downstream.


huwb (approved):
nice find

daleeidd:
Thanks

Issue #971, opened 2022-01-04
Dynamic waves are present in the editor and are absent in the build due to inaccuracy in the code


aaakkkaaa:
**Describe the bug**
Dynamic waves work well, when running an app in Unity editor mode, but completely absent in standalone build. 
The problem is that the script "SphereWaterInteraction.cs" ("\Project\Assets\Crest\Crest\Scripts\Interaction\SphereWaterInteraction.cs") uses the method Shader.Find("Crest/Inputs/Dynamic Waves/Sphere-Water Interaction").

**Screenshots / video**

![Shader_Find](https://user-images.githubusercontent.com/42075485/148102526-7f822cbe-2822-4461-be33-405cb568caf9.jpg)

This code works seamlessly in editor mode. But, since there are no references to this shader in the project scene, Unity does not include the shader in the player build. As a result, the Shader.Find method returns null, which causes an error and aborting the script.

This is one reason to use "Find" in Unity with care.
There are several ways to solve this problem, they are described in the Unity documentation: https://docs.unity3d.com/2021.2/Documentation/ScriptReference/Shader.Find.html

Fix it please. Some users may do it themselves in their projects, but others may have difficulty to find the cause and will not be able to fully enjoy the wonderful Crest Ocean product.

**Versions**
Unity 2021.2.7f1
Render pipeline URP 12.1.2
Crest Ocean System 4.14 for URP

**To Reproduce**
Make a Windows build with Crest example scene "boat.unity".

**Platform**
Standalone Windows

**Hardware**
Not important.

daleeidd:
Thank you for reporting. We will have this fixed for the next release.

huwb:
Thanks again. Fixed merge and will be out in 4.15.

daleeidd:
For anyone who needs this now, move the following:
_Crest/Crest/Shaders/OceanInputs/DynWavesSphereWaterInteraction.shader_
To the following:
_Crest/Crest/Shaders/Resources_

daleeidd:
> For anyone who needs this now, move the following:
> _Crest/Crest/Shaders/OceanInputs/DynWavesSphereWaterInteraction.shader_
> To the following:
> _Crest/Crest/Shaders/Resources_

With 4.15 we attempted to fix but didn't do it properly... if having issues with dynamic waves, then move the following (could be in either of these locations):
> _Crest/Crest/Shaders/OceanInputs/DynWavesSphereWaterInteraction.shader_
[...]

PR #1001, opened 2022-04-04
Give feedback instead of errors due to bad SWI shader move

daleeidd:
Related #971 

We moved this shader into a Resources folder one level down and corrected the path for an include. But we forgot to add a dummy/stub shader in the original location which has caused a few upgrade issues.

This adds the dummy/stub shader to two locations and provides instructions on how users can fix it themselves.

I've added one to _Crest/Crest/Shaders/OceanInputs/_ and _Crest/Crest/Shaders/Resources_ since I was recommending people move it there before 4.15 was released.


huwb (approved):
Looks good

daleeidd:
Thanks

Issue #999, opened 2022-03-29
Texture, Meshes and Material Memory never cleaned due to hideFlags and no Release call when switching scenes


lorenaweder:
Hi! Running an app with several scenes with different instances of an Ocean in them, loading one, another, then back to an menu/into screen (separate Unity scenes all of them) I observed that the profiler shows an increased amount of RenderTextures, Textures, Meshes and Materials each time I return to this intro scene. There's an unload unused resources being called in the middle, but due to the hideFlags of several things some stuff is never cleaned.

Some places where this is found:
- https://github.com/wave-harmonic/crest/blob/master/crest/Assets/Crest/Crest/Scripts/OceanBuilder.cs#L346
- https://github.com/wave-harmonic/crest/blob/master/crest/Assets/Crest/Crest/Scripts/LodData/RegisterClipSurfaceInput.cs#L145
- In the version I'm using also Crest/Scripts/Helpers/CustomPassHelpers.cs
- No code seems to call Release() for the FFTCompute unless new waves are created, but because of the scene switching this also doesn't clean things and I see with each run an accumulation of _tempFFT1, _tempFFT2, etc textures showing up in the profiler. From this script https://github.com/wave-harmonic/crest/blob/master/crest/Assets/Crest/Crest/Scripts/Shapes/FFT/FFTCompute.cs#L76

Maybe I'm using something wrong, but looking at the code itself it looks like a proper memory leak. The ones I found more and more on each run are:

Materials, from ShapeFFT.cs:
_matGenerateWavesGlobal
_matGenerateWavesGeometry

Textures from FFTCompute.cs:
_texButterfly
_texSpectrumControls

RenderTextures from FFTCompute.cs:
_spectrumInit
_spectrumHeight
_spectrumDisplaceX
_spectrumDisplaceZ
_tempFFT1
_tempFFT2
_tempFFT3

Meshes from OceanBuilder.cs:
Each Mesh created from BuildOceanPatch()

--
Unity Version: 2021.2.3f1
Ocean: 4.14 for HDRP (but some of the code remains the same in master in new versions)

lorenaweder:
[...]

PR #1002, opened 2022-04-06
Fix texture leaks

daleeidd:
Helps solve #999.

These are leaks that happen during a single scene session as Unity will unload them on scene switch. But still nice to have them cleared. I have separated them into commits but can be reviewed as a whole.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Helpers/Helpers.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Helpers/Helpers.cs
@@ -139,6 +139,28 @@ namespace Crest
             RenderTargetIdentifierXR(ref texture, ref target);
         }
 
+        /// <summary>
+        /// Creates an RT with an RTD if it does not exist or assigns RTD to RT (RT should be released first). This
+        /// prevents reference leaks.
+        /// </summary>
+        /// <remarks>
+        /// Afterwards call <a href="https://docs.unity3d.com/ScriptReference/RenderTexture.Create.html">Create</a> if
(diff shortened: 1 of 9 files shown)


crest/Assets/Crest/Crest/Scripts/Helpers/Helpers.cs
```diff
@@ -142,6 +142,22 @@ public static void CreateRenderTargetTextureReference(ref RenderTexture texture,
...
+        public static void SafeCreateRenderTexture(ref RenderTexture texture, RenderTextureDescriptor descriptor)
+        {
+            // Do not overwrite reference or it will create reference leak.
+            if (texture == null)
+            {
+                texture = new RenderTexture(descriptor);
+            }
+            else
+            {
+                texture.descriptor = descriptor;
```
huwb, line 157: Do we need to call create or does this apply the new descriptor settings automatically

huwb (approved):

crest/Assets/Crest/Crest/Scripts/Helpers/Helpers.cs
daleeidd: We'd could call Create when ready or let Unity create the texture when is used

daleeidd: I've updated the summary. The links are useful to understand when Create is needed. I think it is there mainly for reading/writing pixels in C# maybe. But generally can get away without it.

huwb (approved):
Looks good

daleeidd:
Thanks

PR #1003, opened 2022-04-06
Fix leaks due to hide flags usage

daleeidd:
Helps solve #999.

Fixes several leaks that will persist during an entire Unity session due to using [HideFlags.DontSave](https://docs.unity3d.com/ScriptReference/HideFlags.DontSave.html) or [HideFlags.HideAndDontSave](https://docs.unity3d.com/ScriptReference/HideFlags.HideAndDontSave.html).


huwb (approved):
Sorry for delay, looks good

daleeidd:
Thanks

PR #1004, opened 2022-04-10
Fix XR SPI both eyes being the same

daleeidd:
I forgot to add XR things to the utility shader which broke stereo rendering.


huwb (approved):

daleeidd:
Thanks

PR #1007, opened 2022-04-20
Fix FFT leaks

daleeidd:
Fixes the last of #999

FFTCompute resources are not released so I am releasing/destroying resources when the last ShapeFFT is destroyed.


huwb (approved):
Nice one

daleeidd:
Thanks.

PR #1008, opened 2022-04-20
Fix long build times when underwater is not used

daleeidd:
When no UR is present, shader stripping is skipped. The shader being in the Resources folder prevents Unity from stripping it.


huwb (approved):
I'm not super familiar with this area but it seems logical to me

daleeidd:
Thanks

Issue #630, opened 2020-08-25
Memory leak


daleeidd:
**Describe the bug**
Reported on Discord. Occasionally, memory usage will grow until it crashes Unity when the editor is left with an active ocean in scene.

**Screenshots / video**
![iu67iu6i](https://user-images.githubusercontent.com/5249806/91191315-7b41e900-e6a9-11ea-92c4-f0dd01baf1df.jpg)

**Versions**
Crest 4.5, HDRP 7.3.1 and Unity 2019.4.8.f1

**Platform**
 - Editor
 - Windows

**Additional context**
We fixed a few memory issues a little while ago which were easier to pinpoint. This one is "occasionally" so it will be more difficult.

daleeidd:
From Discord:
> so there is still a memory leak in editor mode in the latest hdrp version on 2020.1.  The scene with crest I left open overnight went from 4gb ram usage to 25gb.  This looks like maybe a different issue then one I hit previously that seemed linked to entering/exiting play mode.   Refocusing Unity took a while and that brought the memory down quite a bit.

huwb:
Interesting. Any suspicions as to the culprit? Any potential repro steps, or seemingly random so far?

I feel like its CPU memory, but i'm not 100% sure. If you can trap it happen it would be interesting to see the profiler memory pane while connected to the editor.

daleeidd:
I forgot to add this was reported on Discord by two users (updated post). I haven't experienced this myself. It's hard to know since the timeframe is so long.

huwb:
One issue is that I think the ocean may be updating every editor frame even if the viewport is rendering infrequently. I think this is a regression, or maybe it was always broken. This may exacerbate any memory leak (but doesn't explain it).

I'm still yet to see this happen across multiple editor versions and multiple PCs. Maybe it's an editor setting, or window layout/settings or something..

huwb:
On I've seen it now and holy moly it's bad. It comes from creating async GPU readback requests while in edit mode (not play mode). It seems to be an engine bug.

I think a short term fix could just not do them in edit mode. I don't think we rely on them.

daleeidd:
Perfect. Did you use the Unity memory profile to find it?

huwb:
Sort of - the profiler showed unity memory increasing, but it was not attributed to any of the stats, or call stacks, so I binary searched to find the culprit.

huwb:
Fixed in 2019.4.8, and 2020.1.?

[...]

PR #1012, opened 2022-05-01
Re-enable height queries in edit mode

daleeidd:
Related #630

Two cases of memory leaks have been fixed by Unity related to queries for a while now:
https://issuetracker.unity3d.com/issues/asyncgpureadback-can-leak-memory-under-certain-circumstances
https://issuetracker.unity3d.com/issues/asyncgpureadback-dot-request-never-releases-unused-allocated-memory

To be safe I have added a toggle for edit mode queries (enabled by default):

Re-enabling these is useful for visualising height queries and edit mode underwater rendering when height is modified.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
+++ b/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
@@ -311,6 +311,10 @@ namespace Crest
         bool _followSceneCamera = true;
 #pragma warning restore 414
 
+        [Tooltip("Whether height queries are enabled in edit mode."), SerializeField]
+        bool _heightQueries = true;
+
+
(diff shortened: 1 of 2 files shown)


huwb (approved):
Glad to hear those are fixed.

Only thought is that oceanrenderer UI is already large. Worth putting in an Advanced object/subgroup? Then we could migrate more params to it over time. Lmk what you think.

Will leave it up to you and approve.

daleeidd:
Thanks.

> Only thought is that oceanrenderer UI is already large. Worth putting in an Advanced object/subgroup? Then we could migrate more params to it over time. Lmk what you think.

We can look at overhauling for 2022 since migration won't be possible anyway.

PR #1016, opened 2022-05-08
Fix Ocean Depth Cache baked texture size limit

magenta404:
Texture import settings can limit the baked texture size. If users specify a resolution in ODC UI, then they would expect the bake to be the same I think. Reported on Discord that baked texture was limited to 2K.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
@@ -433,6 +433,8 @@ namespace Crest
                 ti.wrapMode = TextureWrapMode.Clamp;
                 // Values are slightly different with NPOT Scale applied.
                 ti.npotScale = TextureImporterNPOTScale.None;
+                // Round up so it encompasses desired resolution.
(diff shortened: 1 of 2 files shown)


crest/Assets/Crest/Crest/Scripts/LodData/OceanDepthCache.cs
```diff
@@ -433,6 +433,8 @@ public override void OnInspectorGUI()
                 ti.wrapMode = TextureWrapMode.Clamp;
                 // Values are slightly different with NPOT Scale applied.
                 ti.npotScale = TextureImporterNPOTScale.None;
+                // Round up so it encompasses desired resolution.
+                ti.maxTextureSize = Mathf.RoundToInt(Mathf.Pow(2f, Mathf.Ceil(Mathf.Log(dc._resolution, 2f))));
```
huwb, line 437: may be worth using max(dc._resolution, 1) . i'm not sure if it could ever be 0 in any situation but log() will go haywire if it is

huwb (approved):

huwb:
will approve anyway so you can pull the trigger, thanks

daleeidd:
Going to defer this one for now. This will fight against the Unity preset system and developers can always do it themselves. Validation could work, but will have to think more about it.

huwb:
That's fair

PR #1017, opened 2022-05-08
Scale samples-per-wave with FFT res

huwb:
Now if you increase the resolution of the FFT, the waves will receive more samples, so the quality ramps with the FFT res. This is how it should have been all along, and it makes for a more natural implementation.

Increased FFT default res to 128 to help improve default quality. This is still cheaper than doing a single 512x512 FFT which would be standard in many systems.

I tried a few test scenes and it did not seem to break the results but a second pair of eyes might be useful.

Finally it adds validation to ensure the ShapeFFT behaves gracefully for invalid settings.


daleeidd (approved):
Nice one. I looked through the example scenes and they all appear to be good. Just needs release notes.

daleeidd:
Perfect. Thanks!

PR #1018, opened 2022-05-18
Fix memory/reference leaks

daleeidd:
Helps solve #999

Fixes more leaks. I went through to see if I could find more, but found none which were proper leaks where duplicates where made. There are still a few references which are not cleared on scene switch, but they were more involved/risky fixes to make and didn't seem worth it for a patch release since they were minor and non duplicating.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
@@ -238,8 +238,17 @@ namespace Crest
 
         void InitData()
         {
+            if (_waveBuffers == null)
             {
                 _waveBuffers = new RenderTexture(_resolution, _resolution, 0, GraphicsFormat.R16G16B16A16_SFloat);
+            }
+            else
+            {
+                _waveBuffers.Release();
+            }
+
+            {
+                _waveBuffers.width = _waveBuffers.height = _resolution;
                 _waveBuffers.wrapMode = TextureWrapMode.Clamp;
                 _waveBuffers.antiAliasing = 1;
                 _waveBuffers.filterMode = FilterMode.Bilinear;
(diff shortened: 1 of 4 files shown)


crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
```diff
@@ -222,8 +238,17 @@ static void InitStatics()
...
         void InitData()
         {
+            if (_waveBuffers == null)
             {
                 _waveBuffers = new RenderTexture(_resolution, _resolution, 0, GraphicsFormat.R16G16B16A16_SFloat);
+            }
+            else
+            {
+                _waveBuffers.Release();
+                _waveBuffers.width = _waveBuffers.height = _resolution;
```
huwb, line 248: Not a blocker but why does the resolution get set here?

crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
```diff
@@ -729,6 +755,23 @@ void OnDisable()
...
 
+        void Awake()
+        {
+            s_Count++;
+        }
+
+        void OnDestroy()
+        {
+            if (--s_Count <= 0)
+            {
```
huwb, line 766: Minor but white space

crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
```diff
@@ -90,6 +90,22 @@ public partial class ShapeGerstner : MonoBehaviour, LodDataMgrAnimWaves.IShapeUp
 
         float _windSpeedWhenGenerated = -1f;
 
+        static int s_Count = 0;
```
huwb, line 93: Wonder if s_InstanceCount or such is more descriptive

huwb (approved):
Will approve as comments minor, nice work

crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
daleeidd: Because it might have changed. I'll move it down as it wouldn't hurt.

crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
daleeidd: Thanks

crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
daleeidd: Sure. Done.

crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
daleeidd: Done.

daleeidd:
Thanks

PR #1019, opened 2022-05-19
Fix jagged shoreline foam when using baked depth cache

daleeidd:
I made baked depth cache point filtering which was a huge mistake... Switched to bilinear.


huwb (approved):
Oh I must have missed that. Approved

daleeidd:
Thanks

PR #1020, opened 2022-05-23
Main Scene cross RP support

daleeidd:
Adds cross RP support for main scene.

The following is the prefab selector which changes prefabs depending on the RP:

It sets the instantiated prefab to _Not Editable_ as editing it in the scene can cause problems. If you need to edit the prefab in the scene, then edit the prefab in _Prefab Mode in Context_:

And here is the Lit shader graph:

There is also a component called _RenderPipelineEditLock_ which sets hierarchy to not editable if the correct RP is not active. In addition, there is an asset processor which prevents prefabs with naming convention (eg _**Crest**\_MainScene_Lighting\_**BIRP**.prefab_) from not saving if correct RP is not active (there might be a smarter way to do this). This prevents a bunch of problems like HDRP automatically changing lights and such being saved.

It works in builds but I have not tested RP switching in builds. I don't think it is necessary to test that case for our examples scenes though.

Downstream the following will need to be done:
- Materials will require changes downstream by going through SRPs and making sure non exposed properties are good and saving them.
- SRP prefabs need to be made and added to prefab selector component

One problem is that I am get the following on load (because light is in prefab):
> Failed creating SceneObjectIdentifier for a light!

I've also cleaned up a few things by removing mines and logo from scene and some materials under _Main_.

Let me know if you want to try this with multiple RPs running and I will add a repository which has a minimal example.

### DIFF
--- a/crest/Assets/Crest/Crest-Examples/Crest.Examples.asmdef
+++ b/crest/Assets/Crest/Crest-Examples/Crest.Examples.asmdef
@@ -1,8 +1,12 @@
 {
     "name": "Crest.Examples",
+    "rootNamespace": "",
     "references": [
         "Crest",
-        "Unity.InputSystem"
+        "Unity.InputSystem",
+        "Unity.RenderPipelines.Core.Runtime",
+        "Unity.RenderPipelines.Universal.Runtime",
+        "Unity.RenderPipelines.HighDefinition.Runtime"
     ],
     "includePlatforms": [],
     "excludePlatforms": [],
@@ -21,6 +25,21 @@
             "name": "com.unity.modules.xr",
             "expression": "1.0.0",
             "define": "ENABLE_XR_MODULE"
+        },
+        {
+            "name": "com.unity.render-pipelines.core",
+            "expression": "",
+            "define": "CREST_SRP"
+        },
+        {
+            "name": "com.unity.render-pipelines.universal",
+            "expression": "",
+            "define": "CREST_URP"
+        },
+        {
+            "name": "com.unity.render-pipelines.high-definition",
+            "expression": "",
+            "define": "CREST_HDRP"
         }
     ],
     "noEngineReferences": false
--- a/crest/Assets/Crest/Crest-Examples/Shared/Scripts/Editor.meta
+++ b/crest/Assets/Crest/Crest-Examples/Shared/Scripts/Editor.meta
@@ -1,0 +1,8 @@
+fileFormatVersion: 2
+guid: d11d3827648a44374b3bc53cd56cfdab
+folderAsset: yes
+DefaultImporter:
+  externalObjects: {}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 
--- a/crest/Assets/Crest/Crest-Examples/Shared/Scripts/Editor/Crest.Examples.Editor.asmdef
+++ b/crest/Assets/Crest/Crest-Examples/Shared/Scripts/Editor/Crest.Examples.Editor.asmdef
@@ -1,0 +1,18 @@
+{
+    "name": "Crest.Examples.Editor",
+    "rootNamespace": "",
+    "references": [
+        "Crest"
+    ],
+    "includePlatforms": [
+        "Editor"
+    ],
+    "excludePlatforms": [],
+    "allowUnsafeCode": false,
+    "overrideReferences": false,
+    "precompiledReferences": [],
+    "autoReferenced": true,
+    "defineConstraints": [],
+    "versionDefines": [],
+    "noEngineReferences": false
+}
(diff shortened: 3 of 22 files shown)


huwb:
Cool thanks for kicking this off. I should be able to test it today.

Initial thought - matching crest*rp.asset may collide with users data. Would matching part of the path be an option? Preventing saving is a heavy handed approach but I can see the reason and am fine with it as long as it will never ever apply to users data.

It's a shame that these things can still be changed by unity even though you marked them non editable. Did wonder if making them non editable by default would help, if that is an option, and then unlocking them using a component or some other event/hook. This may end up over complicating things though.

huwb:
Or maybe we just revert changed we don't want. Once set up we're not making tons of changed to this stuff. Do you think this would hinder development?

huwb:
I did some testing but not sure if the following is useful. It's a bit complicated and I'm not sure what to test, but I did some dumb things listed below and hit quite a few issues so I'll stop here.

I opened in BIRP and it looked good. Then i installed HDRP and enabled it and HDRP worked i think, but acting a bit oddly. When i change the landscape colour it changes the sky intensity. Restarting didnt seem to fix. I'm guessing this is unity getting itself confused rather than a problem with this PR.

Crest is throwing errors in edit mode I think, and when entering play mode

> Crest: An unknown render pipeline is active

I guess `RenderPipelineHelper.CurrentRenderPipeline` is set to `None` . I don't think `RenderPipelineHelper.OnLoad` gets called, but I'm not 100% certain.

The water is pink but I'm sure that's expected in master.

huwb (commented):
Added comments and test results to the thread

daleeidd:
> Initial thought - matching crest*rp.asset may collide with users data. Would matching part of the path be an option? Preventing saving is a heavy handed approach but I can see the reason and am fine with it as long as it will never ever apply to users data.

Path could be an option. I'll do another pass and see if I can find something more safe.

> It's a shame that these things can still be changed by unity even though you marked them non editable. Did wonder if **making them non editable by default in the ondisk asset** would help, if that is an option, and then unlocking them using a component or some other event/hook. This may end up over complicating things though.

I don't think that will make a difference if I understand correctly. One thing Unity does is add those "Additional x Data" components, but only happens when inspecting the component. It bypasses the not editable flag.

> Or maybe we just revert changed we don't want. Once set up we're not making tons of changed to this stuff. Do you think this would hinder development?

We could do that too. This has always been a problem so but not sure if it will be more of a problem with a single branch.

> I'd feel less anxious if these components that manipulate editor behaviour were outside of the crest package.
> 
> However I'm guessing that users will then be faced with the data change mess of they open the crest examples?

They will but as above always been a problem. Probably not a significant problem for users as rarely they poke around there or change RPs.

daleeidd:
> I did some testing but not sure if the following is useful. It's a bit complicated and I'm not sure what to test/how to use this multi-RP stuff, but I did some dumb things listed below and hit quite a few issues so I'll stop here.
> 
> I opened in BIRP and it looked good. Then i installed HDRP and enabled it and HDRP worked i think, but acting a bit oddly. When i change the landscape colour it changes the sky intensity. Restarting didnt seem to fix. I'm guessing this is unity getting itself confused rather than a problem with this PR.

That could be exposure. By default it is adaptive. I pushed the repository I used to develop this:
https://github.com/wave-harmonic/unity-rp-portability

To switch RPs, go to the _Project Settings > Quality_ and there is a quality level per RP.

> Crest is throwing errors in edit mode I think, and when entering play mode
> 
> > Crest: An unknown render pipeline is active
> 
> I guess `RenderPipelineHelper.CurrentRenderPipeline` is set to `None` . I don't think `RenderPipelineHelper.OnLoad` gets called, but I'm not 100% certain.

That's curious. It should work fine going from edit to play mode. I'll look into that. Thanks.

daleeidd:
> Crest is throwing errors in edit mode I think, and when entering play mode
> 
> > Crest: An unknown render pipeline is active
> 
> I guess `RenderPipelineHelper.CurrentRenderPipeline` is set to `None` . I don't think `RenderPipelineHelper.OnLoad` gets called, but I'm not 100% certain.

Fixed. The _CREST_HDRP/URP_ defines weren't in the assembly.

huwb:
Thanks it sounds like my concerns were unfounded so happy to go with what you think makes sense.

I can give it another test in a bit.

I did think about exposure, but it was acting really crazy. Maybe it's overly sensitive in hdrp by default.

daleeidd:
Made a PR downstream which demonstrates RP switching.

> Assets\Crest\Crest\Scripts\Editor\BuildProcessor.cs(149,33): warning CS0618: 'ShaderKeyword.GetKeywordType(Shader, ShaderKeyword)' is obsolete: 'GetKeywordType is deprecated. Only global keywords can have a type. This method always returns ShaderKeywordType.UserDefined.'

This is unrelated.

daleeidd:
Implemented downstream.

PR #1021, opened 2022-05-23
Wave quality improvement

huwb:
Expose option to double sample rate when generating final waves. This unfortunately requires higher res LOD data (1024 works well) because this effectively halves the range of detail and now at res=384 there is very little resolution left in the waves.

As previously, I'm finding higher LOD data resolution is required to get high quality results. To increase the default resolution, we'd likely need to break out LOD resolutions on a per-LOD basis as 1024x1024x8xTexelBytes x LodDataCount is a lot of video memory. For now i've added a tooltip that warns about this memory usage and labels the setting as preview.

Before [video](https://fr2ygz2xuzx8urqegqxn7uysnqr-my.sharepoint.com/:v:/g/personal/huw_bowles_waveharmonic_com/EScsHKPw375Aj1cClLx7ZJQB4vvtR-HGStAyczBSGThzNQ?e=MjWYkm)
After [video](https://fr2ygz2xuzx8urqegqxn7uysnqr-my.sharepoint.com/:v:/g/personal/huw_bowles_waveharmonic_com/Ebdyzv4msD5PmvdlERb5V7wBFB4ZocE6qz7GNMtQMotLrA?e=vaisyw)

The before/after shows some change in apparent wave amplitude, I believe this is a bad result of undersampling the waves rather than something being scaled wrong. I'll think about if this can be proven, but in any case I believe this can be merged.

PR #1022, opened 2022-05-23
Don't let flow period below 1s to avoid obvious phasing

huwb:
Broken out from #1021 as this is simple and can be applied separately.


daleeidd (approved):
Looks good. Nice one.

huwb:
Thanks

PR #1024, opened 2022-05-25
Remove Shape Gerstner Batched

daleeidd:
Removes deprecated _Shape Gerstner Batched_.


huwb (approved):
Yay! There go another couple of thousand lines of non-trivial code

PR #1026, opened 2022-05-25
AnimWaves: disable spline support

huwb:
The use case for a spline would be to set the average water level. This is functionality is now served by the "RegisterHeight" component.

A shader _AnimWavesSetHeightToGeometry.shader_ is left in the project. This actually overrides the height of the water surface with the geometry (rather than changing the sea level which waves can then add on top). It's perhaps a little obscure and could be removable if we really want to, but it does do a thing and is not deprecated so i left it in.


daleeidd (approved):
Nice one. Just needs RNs but approving.

huwb:
TY

PR #1028, opened 2022-05-26
Refactor LodDataMgr to be self contained

daleeidd:
**Status: Proposal**

I'm putting this out early to get early feedback. Hopefully it is understandable.

> **Note**
> Currently this does not compile as it is incomplete. But I know approach does compile. Also the shadow simulation has received the most attention so best to pay attention there.

The goal of this refactor is to make the LodDataMgr self contained and utilise new C# features to replace workaround patterns (like virtual properties accessing static properties). The end goal is that these LDMs will abstract management away and easily allow for things such as rebuilding individual sims when needed instead of the entire OR on change and other things.

New C# features used:
- [Default Interface Methods](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/default-interface-methods)
- [Covariance](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/covariance-contravariance/creating-variant-generic-interfaces)

With these we can reduce have the LDMs (Simulation or LodDataMgr) in a collection (like in OR) without sacrificing code reuse. It also moves enabled and settings property and management of said properties out of OR and greatly simplifies the class. The LDMs are always available which means we can take advantage of inheritance all the time like with validation.

## Wrapper Approach

Relationship graph (Simulation is serialised):
```mermaid
graph TD;
    LodDataMgr-->|Inherits|LodDataMgrShadow;
    Simulation-->|Inherits|ShadowSimulation;
    Simulation-->|Holds| LodDataMgr;
    LodDataMgrShadow-->|References|ShadowSimulation;
    OceanRenderer-->|Holds|Simulation;
```
## Non Wrapper Approach

An alternative approach which uses no wrapper: [refactor/serialized-loddatamgr-2](https://github.com/wave-harmonic/crest/compare/refactor/serialized-loddatamgr-2). There are two commits there: the first one has no intermediary class to handle the generic to keep us having to add a generic class to const/static calls. The second commit adds that in. Another alternative to the second commit (intermediary class) is to move const/statics into a separate class.

## Separate Ocean Material Validation

This replaces the Ocean Material validation from using statics/consts to using an interface with a default method handling validation. The end result is increase code reuse and reduced line count burden on OR.

Definition:
https://github.com/wave-harmonic/crest/compare/refactor/serialized-loddatamgr?expand=1#diff-98bc0292c27302285bdfc62e9db5145ebbd9f56a760906f6d0d6d9c66be225f8R17-R50
Implementation:
https://github.com/wave-harmonic/crest/compare/refactor/serialized-loddatamgr?expand=1#diff-1099375d850e5fdef0a11b95916bbb8ba7ee35123a193dfc71d32812dcbddafaR14-R30
Usage (bit hard to see this one so pasted code here):
https://github.com/wave-harmonic/crest/compare/refactor/serialized-loddatamgr?expand=1#diff-4a2204eda7d2c86ba91bf28c341605275433597ca0586ff3d9fe2a61bb851198R1682-R1684
```csharp
                foreach (var simulation in _simulations.OfType<ISimulationWithMaterialKeyword>())
                {
                    simulation.Validate(ocean, showMessage);
                }
```

## Early Feedback Thoughts

I'll need feedback on whether the wrapper route or the non wrapper route is preferred. Or if neither is preferred.

An advantage of the current approach is it is a little more optimised as when an LDM is disabled it is removed from play. In both new approaches this is not the case so there will be a little overhead (one method call per disabled sim).

With the new approach the UI will need a custom one to be nicer as right now the settings UI is nested in a dropdown which is too much nesting for me. But I think we were looking at doing something like that at some point anyway.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Helpers/OceanDebugGUI.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Helpers/OceanDebugGUI.cs
@@ -311,7 +311,7 @@ namespace Crest
             DrawSim(OceanRenderer.Instance._lodDataDynWaves, ref _drawDynWaves, ref column, 0.5f, 2f);
             DrawSim(OceanRenderer.Instance._lodDataFoam, ref _drawFoam, ref column);
             DrawSim(OceanRenderer.Instance._lodDataFlow, ref _drawFlow, ref column, 0.5f, 2f);
-            DrawSim(OceanRenderer.Instance._lodDataShadow, ref _drawShadow, ref column);
+            DrawSim(OceanRenderer.Instance._shadowSimulation._data, ref _drawShadow, ref column);
             DrawSim(OceanRenderer.Instance._lodDataSeaDepths, ref _drawSeaFloorDepth, ref column);
             DrawSim(OceanRenderer.Instance._lodDataClipSurface, ref _drawClipSurface, ref column);
         }
--- a/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgr.cs
+++ b/crest/Assets/Crest/Crest/Scripts/LodData/LodDataMgr.cs
@@ -101,28 +101,10 @@ namespace Crest
         // shape texture resolution
         int _shapeRes = -1;
 
+        // TODO: Used internally by shadows on failed states. Also for persistent sims.
         public bool enabled { get; protected set; }
 
         protected OceanRenderer _ocean;
-
-        // Implement in any sub-class which supports having an asset file for settings. This is used for polymorphic
-        // operations. A sub-class will also implement an alternative for the specialised type called Settings.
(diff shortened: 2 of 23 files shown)


daleeidd:
I'm leaning towards the alternative approach with the intermediary class as I don't think the wrapper class adds much benefit.

huwb:
Thanks I didn't absorb every detail but i can see this is a great proposal, can see how it would simplify and tidy things up. Gut tells me also no-wrapper, but i'd like to confirm my understanding with you to finalise that vote. Is it basically we have only LodDataMgrShadow (or only ShadowSimulation)?

This is quite a big fish and we have a time window for Crest 2022 that I'd love to get an initial release done, so lets discuss tonight this proposal and when we want to land it.

PR #1029, opened 2022-05-27
Simplify main scene

daleeidd:
Removes the floating logo and naval mines as both are demonstrated elsewhere.


huwb (approved):
Sounds good

daleeidd:
Thanks

PR #1030, opened 2022-05-27
Painted water input

huwb:
This adds a new modality for providing water system inputs - painting.

To make this work, inputs now have a *input mode* (spline/painting/custom geo/primitive/global/...) which explicitly controls how input is driven. This simplified a bunch of code.

Also:

* A manual migration path is added via a *Run 2022 Migration on Ocean Inputs* button added to OceanRenderer Inspector.
* Register components have an 'Unset' input mode and `Awake` tries to auto configure it, which serves as rudimentary migration
* Adds validation to ShapeFFT which was missing before
* Cleans out custom geometry mode from ShapeFFT which was half implemented in a zombie state
* Expands validation for MeshRenderer
* Creates new documentation section for input authoring

Nice to haves / future work:

* A resize/move feature would be nice. Right now the data will simply transform with the object, which is not great. Could come later though.
* Eraser would be nice (not the same as 'painting down' for height input)
* Painting interaction is not very good, can probably be massaged, or needs change to algorithm
* I don't think this will ever happen but it has become evident that ShapeFFT could have been RegisterInputFFT and would have benefitted from being integrated into the rest of the input code. But it works as is.
* The input shaders (like `FlowPainted`, `FoamPainted`) might be all collapsible into a single shader. Needs investigation to see if we can get all the variation needed.


crest/Assets/Crest/Crest/Scripts/LodData/RegisterClipSurfaceInput.cs
```diff
@@ -78,6 +82,41 @@ public enum Primitive
         // The clip surface samples at the displaced position in the ocean shader, so the displacement correction is not needed.
         protected override bool FollowHorizontalMotion => true;
 
+        #region Painting
+        [Header("Paint Settings")]
+        public CPUTexture2DPaintable_R16_AddBlend _paintData;
```
huwb, line 87: the data used by the input is defined by this concrete derived type from CPUTexture

crest/Assets/Crest/Crest/Scripts/LodData/RegisterClipSurfaceInput.cs
```diff
@@ -78,6 +82,41 @@ public enum Primitive
...
+        public IPaintedData PaintedData => _paintData;
+        public Shader PaintedInputShader => Shader.Find("Hidden/Crest/Inputs/Clip Surface/Painted");
+
+        protected override void PreparePaintInputMaterial(Material mat)
+        {
+            base.PreparePaintInputMaterial(mat);
+
+            _paintData.CenterPosition3 = transform.position;
+            _paintData.PrepareMaterial(mat, CPUTexture2DHelpers.ColorConstructFnOneChannel);
```
huwb, line 96: the data used by the input dictates what lambda is used here

crest/Assets/Crest/Crest/Scripts/LodData/RegisterClipSurfaceInput.cs
```diff
@@ -78,6 +82,41 @@ public enum Primitive
...
+        }
+
+        protected override void UpdatePaintInputMaterial(Material mat)
+        {
+            base.UpdatePaintInputMaterial(mat);
+
+            _paintData.CenterPosition3 = transform.position;
+            _paintData.UpdateMaterial(mat, CPUTexture2DHelpers.ColorConstructFnOneChannel);
```
huwb, line 104: the data used by the input dictates what lambda is used here

crest/Assets/Crest/Crest/Scripts/LodData/RegisterClipSurfaceInput.cs
```diff
@@ -78,6 +82,41 @@ public enum Primitive
...
+        }
+
+        public void ClearData()
+        {
+            _paintData.Clear(this, 0f);
```
huwb, line 109: the data used by the input dictates the data type passed to Clear() (in this case float)

crest/Assets/Crest/Crest/Scripts/LodData/RegisterClipSurfaceInput.cs
```diff
@@ -214,11 +257,11 @@ private void LateUpdate()
                 return;
             }
 
-            if (_currentMaterial != _material)
```
huwb, line 217: i believe this was the only thing that used _material so I took the liberty to remove it and just get at the material directly in the below

daleeidd:
Did a test. Looking good. I didn't notice many problems

Just adding some feedback notes:
- If I press W to switch to the translate handles the painting should stop
- Painting buttons for ShapeFFT share space with baking buttons. Some whitespace could help. Or I could add a button attribute so buttons can be positioned anywhere in the UI?
- Having multiple painting components on a single GameObject will start and stop painting together (could leave this)

For the cursor have you tried looking at the _Terrain Tools_ package as it deals with the same problem?

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/AnimWavesGenerate.shader
```diff
@@ -0,0 +1,170 @@
+﻿// Crest Ocean System
```
daleeidd, line 1: Does this replace AnimWavesGerstner shaders?

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/ClipPainted.shader
```diff
@@ -0,0 +1,88 @@
+﻿// Crest Ocean System
```
daleeidd, line 1: Just noting that these painted shaders be one shader? Blend Mode and Color Mask can be scripted. Or an includes file. Not sure if there is more sim specific code is needed in the future. Low priority anyhow.

daleeidd (commented):
Did a review to get an idea but couldn't see any real problems.

crest/Assets/Crest/Crest/Scripts/LodData/RegisterClipSurfaceInput.cs
```diff
@@ -47,7 +51,7 @@ public enum Primitive
 
         [Tooltip("Where the source of the clipping will come from.")]
         [SerializeField]
-        internal Mode _mode = Mode.Primitive;
+        internal Mode _mode = Mode.Painted;
```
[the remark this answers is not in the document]
huwb: yeah good point reverted now

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/AnimWavesGenerate.shader
huwb: it does not.. at the moment..

AnimWavesGerstner is used by Gerstner stuff. Gerstner stuff does not support painted input. it could be made to but I'm not sure if it makes sense, at least from current knowledge. there might be a good use case for it though.

AnimWavesGerstnerGeometry is unfortunately used by both FFT and Gerstner for the spline case (or for user-provided geometry, in theory at least). This is a case of bad naming. We could rename to AnimWavesGenerateFromGeometry potentially?

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/ClipPainted.shader
huwb: good shout. there are some other bits of difference, for example AnimWavesHeightPainted writes to the y channel of the data. some output vectors and some float. this feels solvable.. but needs more work. will leave this comment open.

huwb:
> If I press W to switch to the translate handles the painting should stop

Oh yeah it left the cursor object hanging. Fixed. Thanks

> Painting buttons for ShapeFFT share space with baking buttons. Some whitespace could help. Or I could add a button attribute so buttons can be positioned anywhere in the UI?

Thanks for your offer to help.. I think I could use your assistance on this .. the whole UI is a bit of a challenge and a bunch of the order comes from inheritance. Collapsible groups would be ideal. Also the Mode dropdown can actually remove vast swathes of UI - if its Spline then show spline stuff, if Painted show paint stuff, etc.  However i realise this could be a whole sidequest.. if so just having the buttons moved up to the Painting Settings would work as MVP.

> Having multiple painting components on a single GameObject will start and stop painting together (could leave this)

Well found. It may have to stay. All input types share a single editor tool which i think is good, otherwise it would crowd up the unity tool list. And simply clicking the tool in the tool list immediately activates painting. The downside of this is everyone who is selected things they're being painted when the tool is active as you've observed. I'll think about solutions, but may ship like this.

> Mode property missing on most inputs

Was wondering if this would be called out. I can't get my head around if it's an issue or not. Do you think they're needed?

> No undo/redo support

Will look into this

> For the cursor have you tried looking at the Terrain Tools package as it deals with the same problem?

I have not but thats a good call, will add to my list

huwb:
> No undo/redo support

I was half way through explaining how this would be a nightmare, but then i decided to just brute force it and it seems to work really well for me.. not noticeable performance issues. LMK if it works for you!

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/AnimWavesGenerate.shader
daleeidd: > AnimWavesGerstner is used by Gerstner stuff. Gerstner stuff does not support painted input. it could be made to but I'm not sure if it makes sense, at least from current knowledge. there might be a good use case for it though.

It looks like they can use the same shader as painted is keyword driven. Maybe the changes could be added to Gerstner shader and then we can do a rename in separate PR? That would also improve the diff. LMKWYT

> AnimWavesGerstnerGeometry is unfortunately used by both FFT and Gerstner for the spline case (or for user-provided geometry, in theory at least). This is a case of bad naming. We could rename to AnimWavesGenerateFromGeometry potentially?

Sounds good. I've added it to Trello.

daleeidd:
> > Painting buttons for ShapeFFT share space with baking buttons. Some whitespace could help. Or I could add a button attribute so buttons can be positioned anywhere in the UI?
> 
> Thanks for your offer to help.. I think I could use your assistance on this .. the whole UI is a bit of a challenge and a bunch of the order comes from inheritance. Collapsible groups would be ideal. Also the Mode dropdown can actually remove vast swathes of UI - if its Spline then show spline stuff, if Painted show paint stuff, etc. However i realise this could be a whole sidequest.. if so just having the buttons moved up to the Painting Settings would work as MVP.

No worries. I'll work on something.

> > Mode property missing on most inputs
> 
> Was wondering if this would be called out. I can't get my head around if it's an issue or not. Do you think they're needed?

Nah. I wasn't sure what the plan was. Right now I cannot use a mesh renderer for foam because it defaults to painting with no recourse. Bug?

> I was half way through explaining how this would be a nightmare, but then i decided to just brute force it and it seems to work really well for me.. not noticeable performance issues. LMK if it works for you!

Works for me too. Excellent.

huwb:
> Right now I cannot use a mesh renderer for foam because it defaults to painting with no recourse. Bug?

Ah ok. i thought this was working. Yeah i reckon thats a bug. Adding the Mode dropdown to ShapeFFT made it much easier to use, so the others deserve it to.

I just remembered that i did try to add Mode to a Register component and it would need a bunch of code duplicated in each data type. i started it and it turned into a mess so i stopped. perhaps adding the mode to the Register base class but then letting each data type specify which modes they support might be a way forward. hopefully this works without too much change

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/AnimWavesGenerate.shader
huwb: > Maybe the changes could be added to Gerstner shader and then we can do a rename in separate PR?

I didn't understand, could you speak more to this?

daleeidd:
> perhaps adding the mode to the Register base class but then letting each data type specify which modes they support might be a way forward. hopefully this works without too much change

That sounds ideal. I could look into this too? This will need custom UI code and wondering if can make a property drawer that handles it nicely. Already have an idea in mind.

huwb:
Yeah please do if your up for it. If you want to commit to this branch feel free as well.

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/AnimWavesGenerate.shader
daleeidd: > I didn't understand, could you speak more to this?

Sure. Add painting changes to AnimWavesGerstner.shader instead of creating a new file and then rename it. It should keep git history then. Separate PR might be nice as need to rename AnimWavesGerstner.shader and AnimWavesGerstnerGeometry.shader.

huwb: ohh got it. alright will do

crest/Assets/Crest/Crest/Scripts/LodData/RegisterClipSurfaceInput.cs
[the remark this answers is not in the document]
huwb: effectively getting unreverted now - `_mode` now defined in base class so not easy to assign this default, unless you can think of a way?

crest/Assets/Crest/Crest/Scripts/LodData/RegisterClipSurfaceInput.cs
```diff
@@ -43,25 +41,23 @@ public enum Primitive
...
-        [Tooltip("Where the source of the clipping will come from.")]
-        [SerializeField]
-        internal Mode _mode = Mode.Primitive;
+        [Header("--- PRIMITIVE MODE ---")]
 
         [Tooltip("The primitive to render (signed distance) into the simulation.")]
         [SerializeField, Predicated("_mode", inverted: true, Mode.Primitive), DecoratedField]
         Primitive _primitive = Primitive.Cube;
 
+        // TODO is this really specific to PRIMITIVE mode?
```
huwb, line 50: could you look at this and the below? think i was too enthusiastic with the predication

crest/Assets/Crest/Crest/Scripts/LodData/RegisterClipSurfaceInput.cs
```diff
@@ -303,32 +339,35 @@ protected new void OnDrawGizmosSelected()
                 return;
             }
 
-            // Show gizmo for quad which encompasses the shape.
```
huwb, line 306: mostly whitespace

crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
```diff
@@ -307,7 +382,9 @@ public abstract partial class RegisterLodDataInputWithSplineSupport<LodDataType,
         where LodDataType : LodDataMgr
         where SplinePointCustomData : MonoBehaviour, ISplinePointCustomData
     {
-        [Header("Spline settings")]
```
huwb, line 310: TODO below could use your advice

crest/Assets/Crest/Crest/Scripts/LodData/RegisterClipSurfaceInput.cs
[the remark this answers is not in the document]
daleeidd: Thanks. In the Reset method.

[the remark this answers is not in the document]
huwb: oh yeah good point. will do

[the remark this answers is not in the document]
huwb: done

crest/Assets/Crest/Crest/Scripts/LodData/RegisterClipSurfaceInput.cs
huwb: ah cool added these notes and removed todos

crest/Assets/Crest/Crest/Scripts/LodData/RegisterHeightInput.cs
```diff
@@ -37,10 +39,58 @@ public partial class RegisterHeightInput : RegisterLodDataInputWithSplineSupport
 
         protected override bool FollowHorizontalMotion => true;
 
-        [Header("Height Input Settings")]
```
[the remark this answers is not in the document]
daleeidd: Predicated breaking the foldout is a bug. I know how to fix it though. I can also update Predicated to completely hide fields instead of just disabling.

As for the Header not disappearing, I will have to look at that too.

A question will be whether from a code perspective we'd want to group fields into classes or not. If not I can look at doing a foldout attribute. But maybe it makes sense to do so.

Also there was a comment about lambda not working with Predicated? I cannot find it anymore, but it does support calling methods. `Predicated(Type type, string methodName)`

I think reworking the UI will be easier in separate PR if that's okay. Keep it simple for now. I'm not in favour of the current header style. Just normal headings should be ok. But I will leave up to you whether to keep it.

crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
daleeidd: Replied to this in the other comment.

crest/Assets/Crest/Crest/Scripts/LodData/RegisterHeightInput.cs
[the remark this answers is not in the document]
huwb: thanks. yeah UI rework is fine in separate PR, and can be post 2022 depending on how things go. could you add a trello for it and stick it in somewhere?

i've used the fix you pushed. theres the slight issue that some of the spline settings want to be predicated on multiple things (both 'override=true' and 'mode=spline'). it works ok the way i've done it, i think. rather than supporting predicating on multiple fields, i wonder if having an additional type of predication that completely hides the param rather than disabling is better, ideally the UI would always only show relevant stuff. but that can be a discussion for later.

huwb:
Oh wait this needs migration I guess. And it's not really straightforward either. D'oh. Guess I'll add to the list. Unless we just accept it as breaking change?

huwb:
Or validation that guesses if mode setting is wrong based on components could work well, to at least recover easily. Might do that.

huwb:
Ok I have added validation with a fix button that suggests switching to `CustomGeometryAndShader` if a `Renderer` component is found, or `Spline` if `Spline` component found.

This had a cascading effect where it simplified validation and resulted in a cleanout of some of the old validation which tried to be too smart in terms of detecting what kind of input the user wanted.

Also as a bonus i've hidden the paint preview if not in `Paint` mode which helps things be coherent as well.

crest/Assets/Crest/Crest/Scripts/LodData/RegisterHeightInput.cs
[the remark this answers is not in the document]
daleeidd: Sure. Done.

I agree. I can get around to that as part of the UI rework.

crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
```diff
@@ -447,17 +541,29 @@ public abstract partial class RegisterLodDataInputBase : IValidated
...
+                {
+                    showMessage
+                    (
+                        $"The shader <i>{_paintInputMaterial.shader.name}</i> for material <i>{_paintInputMaterial.name}</i> has multiple passes which might not work as expected as only the first pass is executed. " +
+                        "See documentation for more information on what multi-pass shaders work or",
+                        "use a shader with a single pass.",
+                        ValidatedHelper.MessageType.Warning, this
+                    );
+                }
             }
```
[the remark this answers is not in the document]
huwb: Oh weird yeah. Must have taken a wrong turn refactoring. Thanks. Will check if ValidateRenderer handles any material validation and fix accordingly.

daleeidd (changes requested):
Another round. Also, the shadow ocean input needs the enum filter added.

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/AnimWavesGenerate.shader.shader
```diff
@@ -0,0 +1,170 @@
+﻿// Crest Ocean System
```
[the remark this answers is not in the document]
huwb: fixed

crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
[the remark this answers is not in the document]
huwb: i fixed that and did a bit more consolidation of CustomGeometryAndShader validation

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/AnimWavesGenerate.shader.shader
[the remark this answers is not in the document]
daleeidd: Thanks. Is it possible to have git recognise the rename to prevent some merge issues? If it is too much of a headache.

crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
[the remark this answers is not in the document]
daleeidd: Nice one.

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/AnimWavesGenerate.shader.shader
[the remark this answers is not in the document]
huwb: I'm not sure I'm afraid. It showed as a rename in my local client. But if there's something you know of that I need to do to make it click lmk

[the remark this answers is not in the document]
huwb: Perhaps a squash will sort it out. It will probably at least give it another chance to recognize the rename.

daleeidd:
> Oh wait this needs migration I guess. And it's not really straightforward either. D'oh. Guess I'll add to the list. Unless we just accept it as breaking change?

Would the migration be the _mode_ property? I'm okay with 2022 being a breaking change since it will be one anyway. There are a couple of our example scenes that need fixing up as painted is now the default.

> Or validation that guesses if mode setting is wrong based on components could work well, to at least recover easily. Might do that. That could be useful for future users, not just code for legacy data.

I think we should by default (in the _Reset_ method) try and use the same approach as before. ie: If spline is present, then spline is default, if geometry is present, then geometry is default and then finally painted (in clip surface override if the default is already set to painted or spline, set to primitive). LMKWYT

huwb:
Ok I can look at those two items.

If you could confirm that the docs is on the right track then I will try and get it all wrapped tomorrow.

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/AnimWavesGenerate.shader.shader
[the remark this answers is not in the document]
daleeidd: No worries. Will see what happens.

docs/user/authoring-input-modes.rst
```diff
@@ -0,0 +1,64 @@
...
+Once a Spline is created, this is used to drive the input. A common use of splines is to set the water level to follow a riverbed using the RegisterHeightInput component. A spline may also be used to add waves or flow velocity, if this gives the required level of fidelity (if not then painting may be a good option, described in the previous section). Another typical use case of splines is to add waves aligned to shorelines.
+
+Relevant data components will automatically be added to spline points. For example if the spline is used with a RegisterFlowInput component, the Spline Point Data Flow component will be added to spline points which can then be used to configure the flow speed.
+
+
+# Custom Geometry and Shader
+
+The most advanced type of input involves providing geometry with a shader attached that directly generates the desired data. This geometry will be rendered from a orthographic top down perspective to "print" the data onto the water.
+
+To illustrate a use case .. TODO which use case? Sea floor depth was not good example. Whirlpool was typical advanced input example. Something else?
```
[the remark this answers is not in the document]
daleeidd: Could choose an example from the Examples scene. Or could just use Whirlpool. I am going to backtrack on deleting scenes this time round.

huwb:
Regarding migration of the 'mode' option - I am not allowed to call anything like GetComponent during serialisation (which makes sense). Do you know if this is a blocker or if there is a way around this?

One idea would be to set a flag and then catch it in awake or so, but this then runs into the different scenarios of editor vs build vs standalone and infects the runtime code with bits and i'm hesitant to go that route, and would probably rather use the existing validation fix button to fix up existing cases... but LMKWYT.

crest/Assets/Crest/Crest-Examples/LakesAndRivers/Scenes/Internal/LakesAndRiversSceneCore.prefab
```diff
@@ -1968,13 +1968,22 @@ MonoBehaviour:
   m_Script: {fileID: 11500000, guid: 4202964ceac6ce0449c238ad8b873e0a, type: 3}
   m_Name: 
   m_EditorClassIdentifier: 
+  _mode: 1
```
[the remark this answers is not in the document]
huwb: Fixed manually using the validation Fix button.. for now.. see my note elsewhere about difficulty with migration

huwb:
Added Reset() functionality which i was not sure about before doing it but it seems to work quite well

> The buttons for the Spline which add inputs no longer function as they should. Will have to set the mode to spline after adding them the input.

Elegantly fixed by the Reset() function I think

daleeidd:
> Regarding migration of the 'mode' option - I am not allowed to call anything like GetComponent during serialisation (which makes sense). Do you know if this is a blocker or if there is a way around this?
> 
> One idea would be to set a flag and then catch it in awake or so, but this then runs into the different scenarios of editor vs build vs standalone and infects the runtime code with bits and i'm hesitant to go that route, and would probably rather use the existing validation fix button to fix up existing cases... but LMKWYT.
> 
> Logic like this gives me the willies
> 
> ```
>         void Awake()
>         {
>             // For builds and when Shape* component is enabled in play mode.
>             Upgrade();
>         }
> 
>         void Reset()
>         {
>             // For when the reset button is used.
>             Upgrade();
>         }
> ```

Oh right. I agree that it will be good to avoid that. I guess we'll have to skip migration and go with what you have.

docs/user/authoring-input-modes.rst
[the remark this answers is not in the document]
huwb: Thanks, went for the particle example

huwb:
I believe my next step is to fixup all the scenes

huwb:
Maybe I'll try adding a "migrate for 2022" button to OR inspector. Not super elegant but simple and shouldn't take long to do

huwb:
> Maybe I'll try adding a "migrate for 2022" button to OR inspector. Not super elegant but simple and shouldn't take long to do

Ok i think this helps. I can now fix up the Examples scene with a single click. (oddly the boat clipping example sometimes doesnt work until i select the clipping input gameobject in the hierarchy, but i'm guessing thats unrelated).

I found sometimes the Mode was set to invalid values on RegisterClipInput, because this component had an old property also called `_mode`, so i made the controversial decision to rename the new mode param to `_inputMode`. Then I did it for FFT too to make the UI consistent and docs easier to write!

crest/Assets/Crest/Crest-Examples/LakesAndRivers/Scenes/Internal/LakesAndRiversSceneCore.prefab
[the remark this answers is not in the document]
huwb: Update - also added manual migration feature, see latest comments

daleeidd:
> Ok i think this helps. I can now fix up the Examples scene with a single click. (oddly the boat clipping example sometimes doesnt work until i select the clipping input gameobject in the hierarchy, but i'm guessing thats unrelated).

Maybe. Strange though.

> I found sometimes the Mode was set to invalid values on RegisterClipInput, because this component had an old property also called `_mode`, so i made the controversial decision to rename the new mode param to `_inputMode`. Then I did it for FFT too to make the UI consistent and docs easier to write!

No worries. The alternative would have been to set the enum values to match the old definition:
```csharp
public enum Mode
{
    Painted = 2,
    Spline = 3,
    CustomGeometryAndShader = 0,
    Primitive = 1,
}
```

docs/user/authoring-input-modes.rst
```diff
@@ -0,0 +1,69 @@
...
+
+
+# Global Mode
+
+The input simply applies everywhere.
+
+
+# Painted Mode
+
+The painting modality allows the input to be directly hand authored into the world. This is flexible and powerful. Keep in mind however that this data may need modification or re-authoring if the environment changes. It would be advisable to add quick input until the environment is locked down.
```
daleeidd, line 26: "add quick input until" a little confusing.

docs/user/authoring-input-modes.rst
```diff
@@ -0,0 +1,69 @@
...
+
+Selecting Painted from the Mode dropdown will enable the Paint Mode Settings UI in the Inspector and make the painting tool available which can be enabled by either clicking the Crest Input Painting icon in the tool list, or by clicking Start Painting in the inspector.
+
+Once painting is enabled, a cursor is displayed on the water surface at the mouse position in the Scene view. We recommend reviewing the settings described below before painting in final data.
+
+By left click dragging the mouse, input can be added. For directional input types such as flow or waves, the direction of drag will specify the direction added to the data. On the other hand inputs like foam will simply add foam under the cursor. To remove data, hold shift while left click dragging.
```
daleeidd, line 32: "By left click dragging the mouse" a little awkward. "left click and dragging" is easier to read.

docs/user/authoring-input-modes.rst
```diff
@@ -0,0 +1,69 @@
...
+* **Global** - waves simply added everywhere in the world
+* **Painted** - waves are painted down by hand in particular areas
+* **Spline** - waves are added along a user defined spline
+
+These are selected using the **Input Mode** dropdown on the ShapeFFT component, which display the valid modes for each component.
+
+The following sections describe each mode.
+
+
+# Global Mode
```
daleeidd, line 19: See other files for correct heading syntax. And remember there is a live build you can view to see formatting: https://crest.readthedocs.io/en/feature-painted-waves/user/authoring-input-modes.html

docs/user/authoring-input-modes.rst
```diff
@@ -0,0 +1,69 @@
...
+* **Brush Strength** - scales intensity of the brush interaction.
+
+Some planning is advisable before committing significant time to creating the input. The data is not currently resampled when the size/position/resolution is changed, so for example the data needs to be re-authored if the resolution is changed.
+
+As a final note, painted data will be saved into the scene with the component. Consider minimising data resolution and painting multiple small areas rather than one huge area whenever it makes sense. The storage required by saving this data may be evident by a sudden increase in the size of the scene file. While most painted input is likely to compress well into a small package size, it may still be worth creating prefabs from painted input GameObjects so that the data is stored outside the main level file.
+
+
+# Spline Mode
+
+Selecting Spline from the Mode dropdown will enable Spline Mode Settings UI in the Inspector. This mode requires a Spline component to be present with at least 2 spline points added. Help boxes in the Inspector serve to guide/automate this setup.
```
daleeidd, line 48: A pass on adding italics which is probably already planned. Also I think "2" would be better as "two".

docs/user/authoring-input-modes.rst
```diff
@@ -0,0 +1,69 @@
...
+This geometry will be rendered from a orthographic top down perspective to "print" the data onto the water.
+
+The *Particle Renderer* example in the Examples scene shows a particle system being projected onto the water surface.
+
+
+# Primitive
+
+Use a primitive shape for the input such as a mathematical cube or sphere.
+
+At time of writing used for surface clipping - the volume defined by the primitive shape will be cut from the water surface.
```
daleeidd, line 69: "At time of writing used for surface clipping" Little awkward/confusing.

docs/about/history.rst
```diff
@@ -18,6 +18,9 @@ Breaking
 .. bullet_list::
 
    -  Set minimum Unity version to 2021.3.3.
+   -  Added new *Input Mode* setting to FFT component and to all *RegisterInput* components to support more diverse authoring workflows.
+      This option may need to be configured on existing components.
+      A helper migration feature to automatically fix up existing inputs is available via the *Run 2022 Migration on Ocean Inputs* button in the *OceanRenderer* inspector.
```
daleeidd, line 23: "helper migration" just migration will suffice here I think.

crest/Assets/Crest/Crest/Scripts/Helpers/Attributes/MultiPropertyAttribute.cs
```diff
@@ -247,7 +247,7 @@ internal override void OnGUI(Rect position, SerializedProperty property, GUICont
 
     /// <summary>
     /// Marks which enum options this subclass wants to use. Companion to FilteredAttribute.
-    /// Usage: [FilterEnum("_mode", FilteredAttribute.Mode.Include, (int)Mode.One, (int)Mode.Two)]
+    /// Usage: [FilterEnum("_inputMode", FilteredAttribute.Mode.Include, (int)Mode.One, (int)Mode.Two)]
```
daleeidd, line 250: This change could be dropped.

crest/Assets/Crest/Crest/Scripts/Helpers/Editor/EditorHelpers.cs
```diff
@@ -68,6 +68,77 @@ public static SceneView GetSceneViewFromSceneCamera(Camera camera)
...
+            System.Action<RegisterLodDataInputBase, RegisterLodDataInputBase.InputMode> setInputMode = (input, newMode) =>
+            {
+                Debug.Log($"Crest: Changing Input Mode of {input.GetType().Name} component on GameObject {input.gameObject.name} from {input._inputMode.ToString()} to {newMode}. Click this message to highlight this GameObject.", input);
+                input._inputMode = newMode;
+                EditorUtility.SetDirty(input);
+            };
+
+            foreach (var input in GameObject.FindObjectsOfType<RegisterLodDataInputBase>(true))
+            {
+                // Painted is a new mode. Forcibly upgrade it because it is also the default aso components are likely to have it.
```
daleeidd, line 116: "aso"

daleeidd (changes requested):
Another round. Pretty minor. I have to do some testing and maybe add more items if I find it.

crest/Assets/Crest/Crest/Scripts/LodData/RegisterClipSurfaceInput.cs
```diff
@@ -179,7 +214,11 @@ public override void Draw(LodDataMgr lodData, CommandBuffer buf, float weight, i
             buf.SetGlobalFloat(LodDataMgr.sp_LD_SliceIndex, lodIdx);
             buf.SetGlobalVector(sp_DisplacementAtInputPosition, Vector3.zero);
 
-            if (_mode == Mode.Primitive)
+            if (_inputMode == InputMode.Painted)
+            {
+                buf.DrawProcedural(Matrix4x4.identity, _paintInputMaterial, 0, MeshTopology.Triangles, 3);
```
daleeidd, line 219: Loading the main scene this is throwing exception because the material doesn't exist.

daleeidd:
> Added Reset() functionality which i was not sure about before doing it but it seems to work quite well
> 
> > The buttons for the Spline which add inputs no longer function as they should. Will have to set the mode to spline after adding them the input.
> 
> Elegantly fixed by the Reset() function I think

I think I made a mistake with this. Reset isn't called in play mode. Currently there is a problem with the ODC and this approach. On entering play mode the input is recreated and Reset is not called. Keeping Reset still makes sense because of the reset button.

> Logic like this gives me the willies
> 
> ```
>         void Awake()
>         {
>             // For builds and when Shape* component is enabled in play mode.
>             Upgrade();
>         }
> 
>         void Reset()
>         {
>             // For when the reset button is used.
>             Upgrade();
>         }
> ```

This approach is correct for this situation. In play mode only Awake will run. In edit mode both will run on instantiate. When reset button is pressed, Reset is only called. Code in Upgrade will either have to stomp previous state or not do anything additive to avoid duplication. I'll document this pattern where FilterEnumAttribute definition is.

huwb:
What's the play mode case here? I was not intending anything to happen in play mode.. but I might be missing an important requirement!

docs/user/authoring-input-modes.rst
huwb: Thanks will finalize docs with this and other notes

daleeidd:
The ODC creates a RegisterSeaFloorDepthInput via script. It has the correct default in edit mode, but not in play mode as Reset does not run.

huwb:
Oh totally missed that somehow, thanks for finding that. I actually totally neglected the RegisterSeaFloorDepthInput so did a couple of fixes.

I haven't been thinking about this as a migration issue but rather of a setup issue. I propose what's committed - configure the register component once attached. I did a similar fix for the whirlpool.

What's not great is that DefaultMode doesn't get applied when attaching a component at runtime which is unintuitive/broke. I need precisely the event when the component is attached which I don't think i can get? I don't currently consider this to be a legacy data/migration issue so I don't believe upgrading patterns apply. But i might have missed the point.

I thought perhaps i could use a *ISerializationCallbackReceiver* callback to apply our default before deserialisation but that specific case is not covered with a callback it appears.

The only option i could think of is to add an Unset entry to the Mode enum, and the component simply does nothing if Unset, and it at least gives the user a pointer. Thinking more, what if we simply configure the enum filter to display "Unset" for invalid values? Right now if i add a RegisterSeaFloorDepthInput in play mode, the mode dropdown will be empty, should we make it say Unset:

image

daleeidd:
> I haven't been thinking about this as a migration issue but rather of a setup issue. I propose what's committed - configure the register component once attached. I did a similar fix for the whirlpool.
> 
> What's not great is that DefaultMode doesn't get applied when attaching a component at runtime which is unintuitive/broke. I need precisely the event when the component is attached which I don't think i can get? I don't currently consider this to be a legacy data/migration issue so I don't believe upgrading patterns apply. But i might have missed the point.

Same. I am saying this is the correct pattern for setting the default. I was just using that code as an example. Apologies for the confusion.:

```csharp
void Awake() => SetDefault();
void Reset() => SetDefault();
void SetDefault() => _mode = DefaultMode;
```

Both Awake and Reset together should catch all cases for editor and build.

> I need precisely the event when the component is attached which I don't think i can get?

I thought that was Awake.

daleeidd:
> The only option i could think of is to add an Unset entry to the Mode enum, and the component simply does nothing if Unset, and it at least gives the user a pointer.
> 
> Alternatively, would another possibility be to simply configure the enum filter to display "Unset" for invalid values? Right now if i add a RegisterSeaFloorDepthInput in play mode, the mode dropdown will be empty, should we make it say Unset:
> 
> image

It makes sense to do that either way I think in a case a bad value does get set.

daleeidd:
> > > I need precisely the event when the component is attached which I don't think i can get?
> 
> > I thought that was Awake.
> 
> Ah no Awake always gets called, so I have no way of knowing whether the component was just added or preconfigured.

Oh yes... we'll clobber it. I don't think there is a runtime Reset equivalent. Only solution I can think of is making a Unset/None a value in the enum and the default for the field declaration and filter it out in the UI. But that does have scripting implications if we are worried about that :\

huwb:
Decided to try adding Unset and actually it makes the code simpler/clearer so its a win if you agree

I didn't attempt to do anything fancy with the mode dropdown. I suppose if Unset is not currently selected, it could be filtered from the list? I don't hate having Unset there as its less magic to just show it. I dont think anyone will click it thinking it will do something useful. However if you feel it should be removed and it sounds possible lmk and ill look into it.

I tested adding all register components, in edit and play mode, and fixed up a few things there. Also tested "Run Validation" and manual migration features.

crest/Assets/Crest/Crest/Scripts/LodData/RegisterClipSurfaceInput.cs
huwb: fixed, thanks

crest/Assets/Crest/Crest/Scripts/Helpers/Editor/EditorHelpers.cs
huwb: removed

crest/Assets/Crest/Crest/Scripts/Helpers/Attributes/MultiPropertyAttribute.cs
huwb: ah oops done

docs/user/authoring-input-modes.rst
huwb: agreed fixed

docs/about/history.rst
huwb: good point fixed

docs/user/authoring-input-modes.rst
huwb: yeah agreed, actually this sentence was unnecessary, removed

docs/user/authoring-input-modes.rst
huwb: ah yes thanks

daleeidd:
I added a few commits which fixed a couple of things like the clip input not working in the main scene.

> Decided to try adding Unset and actually it makes the code simpler/clearer so its a win if you agree

Nice. Sounds good.

> I didn't attempt to do anything fancy with the mode dropdown. I suppose if Unset is not currently selected, it could be filtered from the list? I don't hate having Unset there as its less magic to just show it. I dont think anyone will click it thinking it will do something useful. However maybe its clutter.. if you feel it should be removed and it sounds possible lmk and ill look into it.

I have this working. I can add it after it is merged or throw it on top so you can squash all the commits.

> I tested adding all register components, in edit and play mode, and fixed up a few things there. Also tested "Run Validation" and manual migration features.
> 
> To summarise:
> 
> * If component added in edit mode, `Reset()` called and default mode set as baseline, then auto detection may suggest based on attached components.

> * If component added in play mode/runtime, input mode will be unset which will emit errors to the log and to the inspector, with a fix button. Expectation is that code adding components will configure them.

With the Unset mode you added it is now trivial to handle it and also serves as a migration. See the last two commits. You can drop them if you think it is best to do so.

> * I didnt bother with FFT. If it is added in Edit mode it will detect spline/custom geo. At runtime it just defaults to Global. This is easy enough to understand and doesnt cause any issue i believe.

Sounds reasonable. We could also do that same above if you think it is worthwhile.

huwb:
Thanks for the clip fix, apologies for breaking that

> With the Unset mode you added it is now trivial to handle it and also serves as a migration

I'm guessing it needs a setdirty call to be solid as migration, which will make it less nice. But I'll follow your lead on this, happy to go with that.

Re unset yeah feel free to add it in to this, thanks

I wouldn't do this for fft but let me know if you think otherwise

huwb:
I suppose there is simply no way to deal with a default that needs to change based on the derived type, particularly if that default depends on other attached components. Think we've fallen into a quite exotic pitfall/limitation! The logical thought would then be to not have this default, or make the default always valid, and i think this is covered by my sub-pr which introduced worse complexities due to having additional state.

Just understood why FFT - are these the reasons?

* Legacy data - existing components will all be in Global mode and won't detect if they are attached to a spline.  Mitigating factors - spline stuff was still in preview, breaking changes are less-bad with 2022 upgrade, migration tool automatically fixes it with a click.
* Runtime init - adding a component at runtime will end up in Global wave mode. But calling code can (and probably should as general good practice) initialise the mode. This is standard for any code creating any component.

I think i'd vote for leaving FFT - adding an 'invalid' state and then code to handle it for the above reasons is not warranted i feel. What do you think?

Ok while i have reservations and feel we've fallen out of simple coding patterns, i cant think of a better way so yeah i will add the setdirty call and let me know if you still feel it should go to FFT.

daleeidd:
> Just understood why FFT - are these the reasons?
> 
> * Legacy data - existing components will all be in Global mode and won't detect if they are attached to a spline.  Mitigating factors - spline stuff was still in preview, breaking changes are less-bad with 2022 upgrade, migration tool automatically fixes it with a click.
> * Runtime init - adding a component at runtime will end up in Global wave mode. But calling code can (and probably should as general good practice) initialise the mode. This is standard for any code creating any component.
> 
> I think i'd vote for leaving FFT - adding an 'invalid' state and then code to handle it for the above reasons is not warranted i feel. What do you think?

Really good points. Adding invalid options like error states to enums does happen but also smelly. I'll vote go ahead without it too.

> Oh and the other point - please do add removal of Unset if it helps the UX.

I'll add it outside of PR.

daleeidd (changes requested):
Just one issue.

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
```diff
@@ -108,7 +152,15 @@ public partial class ShapeFFT : MonoBehaviour, LodDataMgrAnimWaves.IShapeUpdatab
...
-        Mesh _meshForDrawingWaves;
+        public IPaintedData PaintedData => _paintData;
+        public Shader PaintedInputShader => null;
+
+        Mesh MeshForGeneration => _inputMode == Mode.Spline ? _meshForDrawingWavesSpline : (_inputMode == Mode.CustomGeometryAndShader ? _meshForDrawingWavesCustom : null);
+        bool UsingGeometryToGenerate => _inputMode == Mode.Spline || _inputMode == Mode.CustomGeometryAndShader;
+
+        // Will either be spline mesh, or custom geo assigned by user
+        Mesh _meshForDrawingWavesSpline;
+        Mesh _meshForDrawingWavesCustom;
```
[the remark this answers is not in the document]
huwb: You couldnt get custom geo to work with FFT? Hm i have not tried that in forever. I'll look around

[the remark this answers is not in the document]
huwb: A tempting fallback will be to remove this. I highly doubt anyone will use it over painting/spline. I guess legacy data is a consideration - not sure i heard of anyone using it.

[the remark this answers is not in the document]
huwb: Ok funny enough, basic validation was basically missing for FFT. Added that first.

Also made it actually initialise *_meshForDrawingWavesCustom*

With the benefit of hindsight i think FFT could have been implemented as *RegisterFFTInput* and be deeply connected into all the stuff in RegisterLodInput. It borrows parts of it slightly awkwardly. But this would entail a migration nightmare i guess - and a rewrite.

I think this works as is but lmk what you think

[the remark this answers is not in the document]
daleeidd: Oh... wait... I just remembered that I don't think FFTs supported renderers as it was all in on splines. Apologies.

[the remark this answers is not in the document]
huwb: I believe removing it is the best option, and I've done so now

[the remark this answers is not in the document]
huwb: Also I retested on master and it doesnt work there either. Not sure why i left that code hanging around. I have therefore not bothered to add RN.

huwb:
BIRP scenes migrated

crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
```diff
@@ -95,30 +130,113 @@ public static OceanInput GetRegistrar(Type lodDataMgrType)
...
+            // run for components created at runtime ensuring they enter a valid state,
+            // and also helps data migration in edit mode (provided component is enabled).
+            if (_inputMode == InputMode.Unset)
             {
+                AutoDetectMode(out _inputMode);
+
+                // Make sure data change is registered if this runs in edit mode
 #if UNITY_EDITOR
-                if (Application.isPlaying && verifyShader)
+                EditorUtility.SetDirty(this);
```
daleeidd, line 171: Calling _SetDirty_ here has a few issues sadly.

This will need to be wrapped to avoid always dirtying the main scene when ODC adds the SFD input:

```csharp
if (!Application.isPlaying && !hideFlags.HasFlag(HideFlags.DontSaveInEditor))
{
    EditorUtility.SetDirty(this);
}
```

Also, it isn't reliable which is why I think you had issues with migrations. I found that _Undo.RecordObject_ can be more reliable but neither are reliable on scene load.

There is no nice workaround unfortunately. _SetDirty_ would need to be called in _OnEnable_ at the earliest. Or in OnValidate. But both would need some flag to prevent _Undo.RecordObject_ from messing with undo/redo history and only trigger on migration.

Since this code is permanent, we could skip dirtying the scene? Although I called it a free migration it more saves us having to do a migration on our own data. But really it is just a normal code path.

daleeidd (changes requested):
Just a couple of things and the main scene needs to be migrated.

crest/Assets/Crest/Crest-Examples/Examples/Scenes/Internal/ExamplesScene.prefab
```diff
@@ -85,7 +85,7 @@ GameObject:
     type: 3}
```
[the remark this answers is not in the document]
huwb: Cool will do

crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
huwb: My usual thing is that any serialised data visible to the user should either be what is on disk, or dirty. Not doing this opens up edge cases where its auto picking Mesh but then i later add a Spline and then get confused why suddenly its picking Spline instead, stuff like that. But thats probably a 0.00001% edge case.

Hm disappointing to hear its unreliable, i feel like its been solid for me but i'm not sure i've called it from awake in the past.

In light of this i agree on foregoing it as the edge cases are tiny as noted above.

huwb: Removed now

crest/Assets/Crest/Crest-Examples/Examples/Scenes/Internal/ExamplesScene.prefab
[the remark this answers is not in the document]
huwb: Done

huwb:
Alright i created an archive branch for the full history `feature/painted-waves-archive` and squished these changes down on top of `master`. Let me know if you have any other feedback items. Thanks!

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/ClipPainted.shader
huwb: i don't think i'll get to this in the short term so i'm going to document in the future work notes in the PR description. it needs investigation whether we can get the required variation across shaders cleanly/elegantly, as noted in my previous comment.

crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
daleeidd: > My usual thing is that any serialised data visible to the user should either be what is on disk, or dirty. Not doing this opens up edge cases where its auto picking Mesh but then i later add a Spline and then get confused why suddenly its picking Spline in my standalone builds when i thought it was set to Mesh, stuff like that. But thats probably a 0.00001% edge case.

Good point and agreed. In that scenario the added spline will dirty the scene for us.

> Hm disappointing to hear its unreliable, i feel like its been solid for me but i'm not sure i've called it from awake in the past. Or i just haven't noticed it failing..

The ODC may have been masking the issue when it creates the SFD input (or anything else that creates an ocean input from script) as this will trigger the SetDirty later.

daleeidd (changes requested):
Just needs the build errors fixed. Before doing so, grab master again as I added a build fix for another PR which I didn't catch earlier.

huwb:
D'oh thanks

crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
```diff
@@ -95,32 +130,110 @@ public static OceanInput GetRegistrar(Type lodDataMgrType)
...
                 }
-#endif
+            }
 
-                _material = _renderer.sharedMaterial;
+            if (_inputMode == InputMode.Painted)
+            {
+                if (this is IPaintable paintable)
+                {
+                    PaintableEditor.DrawPaintAreaGizmo(paintable, GizmoColor);
```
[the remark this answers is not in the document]
huwb: fixed thanks

crest/Assets/Crest/Crest/Scripts/LodData/Shared/UserPaintedDataPreview.cs
```diff
@@ -0,0 +1,42 @@
+// Crest Ocean System
```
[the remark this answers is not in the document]
huwb: fixed

huwb:
> main scene needs to be migrated

Just realised the main scene changes were not baked in. Damn dirty flag. Done now.

daleeidd (approved):
Awesome. All looks good to me. Great work mate.

huwb:
Merged, TYVM!

crest/Assets/Crest/Crest/Scripts/LodData/RegisterLodDataInput.cs
huwb: this was haunting me a bit so i went back to revist and you're right, setdirty is doing nothing here! i guess it doesnt work from "runtime" functions like `Awake()`. i've never had an issue calling it form editors etc. in this case i'm moving on, but i think this should be an exception to the rule - in general data should either match disk or be dirty. thanks again for raising this thread.

crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/AnimWavesGenerate.shader.shader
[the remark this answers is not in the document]
daleeidd: Forgot about this. Looks like it didn't happen after squashing. Oh well.

PR #1031, opened 2022-05-27
Remove ocean surface alpha

daleeidd:
Remove ocean surface alpha in favour of Albedo feature or Unity decal projector for HDRP.


huwb:
Wow GitHub mobile site won't let me review. Happily approved!

huwb:
Oh is there a material or two that can be removed too? Just checking

daleeidd:
Thanks! Nah they were removed with #1029 as they were tied with the main scene.

PR #1032, opened 2022-05-28
Local wave alpha blending

daleeidd:
Adds alpha blending for local waves (geometry or splines). It adds an _Additive_ option which is on by default. It can only be disabled for local waves as no point for global ones.

Also fixes FFTs writing into alpha channel which was adding a bunch of foam:

FFT Waves alpha channel rendered to the surface

Lastly, it removes variance from gerstner waves.

We don't use ShapeGerstner anywhere so example scenes shouldn't need tweaking.

I've added a test scene under Assets/Demo folder.

I'll look at reorganising the UI in another PR as it is getting cluttered.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
@@ -71,6 +71,14 @@ namespace Crest
         [Tooltip("Multiplier for these waves to scale up/down."), Range(0f, 1f)]
         public float _weight = 1f;
 
+        [Predicated("_inputMode", inverted: false, Mode.Global), DecoratedField]
+        [Tooltip("How the waves are blended into the wave buffer. Use <i>AlphaBlend</i> to override waves.")]
+        public Helpers.BlendPreset _blendMode = Helpers.BlendPreset.AdditiveBlend;
+
+        [Predicated("_inputMode", inverted: true, Mode.Spline), DecoratedField]
+        [Tooltip("Order this input will render. Queue is <i>Queue + SiblingIndex</i>")]
+        public int _queue = 0;
+
         [Tooltip("How much these waves respect the shallow water attenuation setting in the Animated Waves Settings. Set to 0 to ignore shallow water."), SerializeField, Range(0f, 1f)]
         public float _respectShallowWaterAttenuation = 1f;
 
@@ -104,6 +112,9 @@ namespace Crest
         {
             _paintData.CenterPosition3 = transform.position;
             _paintData.UpdateMaterial(mat, CPUTexture2DHelpers.ColorConstructFnTwoChannel);
+#if UNITY_EDITOR
+            Helpers.SetBlendFromPreset(mat, _blendMode);
+#endif
         }
 
         public Vector2 WorldSize => _paintData.WorldSize;
@@ -336,6 +347,18 @@ namespace Crest
                 UpdatePaintInputMaterial(mat);
             }
 
+#if UNITY_EDITOR
+            if (_inputMode == Mode.Spline && _matGenerateWavesGeometry != null)
+            {
+                Helpers.SetBlendFromPreset(_matGenerateWavesGeometry, _blendMode);
+            }
+
+            if (_inputMode == Mode.Global && _matGenerateWavesGlobal != null)
+            {
+                Helpers.SetBlendFromPreset(_matGenerateWavesGlobal, Helpers.BlendPreset.AdditiveBlend);
+            }
+#endif
+
             // If using geo, the primary wave dir is used by the input shader to rotate the waves relative
             // to the geo rotation. If not, the wind direction is already used in the FFT gen.
             var waveDir = (_inputMode == Mode.Spline || _inputMode == Mode.Painted) ? PrimaryWaveDirection : Vector2.right;
@@ -400,6 +423,8 @@ namespace Crest
                 {
                     _matGenerateWavesGlobal = new Material(Shader.Find("Hidden/Crest/Inputs/Animated Waves/Generate Waves"));
                 }
+
+                Helpers.SetBlendFromPreset(_matGenerateWavesGlobal, Helpers.BlendPreset.AdditiveBlend);
             }
             else if (_inputMode == Mode.Painted)
             {
@@ -411,6 +436,8 @@ namespace Crest
                 // This should probably warn or error on multiple input types (GetComponents<IUserAuthoredInput>().length > 1) in
                 // validation
                 _paintData.PrepareMaterial(_matGenerateWavesPainted, CPUTexture2DHelpers.ColorConstructFnTwoChannel);
+
+                Helpers.SetBlendFromPreset(_matGenerateWavesPainted, _blendMode);
             }
             else if (_inputMode == Mode.Spline)
             {
@@ -430,6 +457,8 @@ namespace Crest
                     {
                         _matGenerateWavesGeometry = new Material(Shader.Find("Crest/Inputs/Animated Waves/Gerstner Geometry"));
                     }
+
+                    Helpers.SetBlendFromPreset(_matGenerateWavesGeometry, _blendMode);
                 }
             }
 
@@ -438,7 +467,7 @@ namespace Crest
             {
                 // Queue determines draw order of this input. Global waves should be rendered first. They are additive
                 // so not order dependent.
-                var queue = 0;
+                var queue = _inputMode == Mode.Global ? int.MinValue : _queue;
                 var subQueue = transform.GetSiblingIndex();
 
                 // Submit draws to create the FFT waves
--- a/crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
@@ -54,6 +54,14 @@ namespace Crest
         [Tooltip("Multiplier for these waves to scale up/down."), Range(0f, 1f)]
         public float _weight = 1f;
 
+        [Predicated(typeof(ShapeGerstner), "IsLocalWaves"), DecoratedField]
+        [Tooltip("How the waves are blended into the wave buffer. Use <i>AlphaBlend</i> to override waves.")]
+        public Helpers.BlendPreset _blendMode = Helpers.BlendPreset.AdditiveBlend;
+
+        [Predicated(typeof(ShapeGerstner), "IsLocalWaves"), DecoratedField]
+        [Tooltip("Order this input will render.")]
+        public int _queue = 0;
+
         [Tooltip("How much these waves respect the shallow water attenuation setting in the Animated Waves Settings. Set to 0 to ignore shallow water."), SerializeField, Range(0f, 1f)]
         public float _respectShallowWaterAttenuation = 1f;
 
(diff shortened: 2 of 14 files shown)


daleeidd:
I forgot about feathering... will have to look at that at some point. LMKWYT anyhow

huwb:
Ahh I totally forgot we were kicking out the variance. I think we agreed to do this.. but I seem to have completely forgotten the line of thinking. Were we thinking to not do the variance thing anymore? Or was it that we were going to upgrade the variance to proper analytical variance which would be in separate data?

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
```diff
@@ -437,6 +457,12 @@ public void OnSplinePointDrawGizmosSelected(SplinePoint point)
         {
             DrawMesh();
         }
+
+        // Called by Predicated attribute. Signature must not be changed.
+        bool IsLocalWaves(Component component)
```
huwb, line 462: I'm proposing to add a Mode setting in painting pr which can be used here

daleeidd:
I believe it was the latter.

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
daleeidd: Sounds good. I'll wait until that merges.

huwb: I'd say whichever comes first, not sure how close painting is so perhaps could be done later if this merges first

huwb:
Yeah that works. I got a branch which calculates variance - I think. But not for 2022 :)

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
daleeidd: Done.

daleeidd:
Ok... I think this is done.

huwb:
Thanks will try this later

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
```diff
@@ -432,6 +447,10 @@ void InitBatches()
                     {
                         _matGenerateWavesGeometry = new Material(Shader.Find("Crest/Inputs/Animated Waves/Gerstner Geometry"));
                     }
+
+                    _matGenerateWavesGeometry.SetInt(Helpers.ShaderIDs.s_BlendSrcMode, (int)(_additive ? BlendMode.One : BlendMode.SrcAlpha));
+                    _matGenerateWavesGeometry.SetInt(Helpers.ShaderIDs.s_BlendDstMode, (int)(_additive ? BlendMode.One : BlendMode.OneMinusSrcAlpha));
+                    _matGenerateWavesGeometry.SetKeyword("_ADDITIVE", _additive);
```
huwb, line 453: Should keywords be preceded by CREST

huwb:
Cool i think i got the idea and i think its good

Could/should this be an enum *Blend Mode* where options are *Additive", *Alpha Blend*? There are other potentially usable blend modes *OneMinusDestAlpha* (or maybe *Multiply but i'm not sure)/ i dont think we should add additional blend modes beyond the initial two right now, but just checking if we should generalising this option and making it more explicit potentially. if it's going to be a pin then it is not critical.

Would it be easy to support this blend options for Painted? I want to override the waves just around the shoreline and everywhere else i want to use global waves. I believe it would work well if it would use the alpha value that i've painted (the alpha value is derived from the length of the vector in the direcition map). That way i can gently paint in my overrides and have it gracefully fall back where i haven't painted. I tried doing this with additive in the island scene and its an awkward balancing act - one must try to control the weights for both sets of waves to get them to blend nicely which is not a pleasant experience.

huwb (changes requested):
Potential changes called out in comments

huwb:
Ok i did some hacks to make sure i'm not talking rubbish regarding how painted input is blended and to build confidence. I found something that I think works well as a default, if possible.

The painted wave input is alpha blended on top of any global waves etc, with the blend weight coming from proximity to the sides of the domain. In the beginning of the video the painted data is all 0, which shows the blend clearly.

Then when i paint waves I can control both the direction of the waves and the amplitude (so i could leave a little quiet area downwind of the island, and could also have small amplitude waves in the 'bay').

https://fr2ygz2xuzx8urqegqxn7uysnqr-my.sharepoint.com/:v:/g/personal/huw_bowles_waveharmonic_com/EfeUORRHquVHq0w1D3mubyoB3KimEBVjwje8C9gTGoIRfA?e=cG2ZQc

This requires two things, which may or may not be relevant to this PR:

* Default blend mode for painted waves input set to alpha blend, but option for additive. Additive is nice to add little reflection waves to the environment and potentially other things. A blend mode dropdown could make sense here. Perhaps a second dropdown in the painted section is easiest.
* Painted needs to come after Global in order to blend in correct order. And multiple Painted waves need to blend consistently. I have pushed a proposal to `feature/local-wave-alpha-blending-painted-blend` which sorts Global first, then Painted based on transform sibling index which may be a solution. 

Pushed to a sub-branch: `feature/local-wave-alpha-blending-painted-blend`

This may have converged on something that can come after this PR which would work too.

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
daleeidd: I've removed it as not needed.

daleeidd:
Didn't see your recent posts. I came up with a nicer solution than my previous one (no keywords) and I have incorporated your commits which fixed painted alpha blending. Thanks!

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
```diff
@@ -432,20 +456,32 @@ void InitBatches()
...
+
+                    Helpers.SetBlendFromPreset(_matGenerateWavesGeometry, _blendMode);
                 }
             }
 
             var usingGeometryToGenerate = _inputMode == Mode.Spline;
             if (!usingGeometryToGenerate || MeshForGeneration)
             {
+                // Queue determines draw order of this input. GetSiblingIndex gives intuitive ordering.
+                var queue = _queue + transform.GetSiblingIndex();
```
daleeidd, line 468: I think this should work well. LMKWYT

daleeidd:
I forgot Gerstner. I will do that tomorrow if you want to wait for that. Otherwise LMKWYT.

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
huwb: I think there are potential difficulties. I could set the queue to -1 to make it earlier, but if it's the 5th sibling it will end up in queue 4, even if other siblings are not wave inputs. And this could change of I reparent the GO.

It's a bit of a flawed approach generally but the idea of having it 10+index is the 10 would push it to the end of the"normal" range, then adding index should not reorganise within that range. Even that is a bit hopeful and has fail cases too. 

Oh maybe I have an idea, will propose via code

huwb: Ok its trivial so just writing here:

`var queue = 1000 * _queue + transform.GetSiblingIndex();`

If the sibling index is less than 1000, this will successfully sort on *both* `_queue` and sibling index.

The remaining question mark is if i have other animated waves inputs, like register height input which should also take a factor of 1000 to sort consistently against ShapeFFT

```
        protected virtual bool GetQueue(out int queue)
        {
            var rend = GetComponent<Renderer>();
            if (rend && rend.sharedMaterial != null)
            {
                queue = 1000 * rend.sharedMaterial.renderQueue;
                return true;
            }
            queue = int.MinValue;
            return false;
        }
```

I sense that maybe this is getting thorny in terms of understanding and maintenance. Maybe changing CrestSortedList.Add to take a subindex forces the issue to all the call sites, and allows calling code to give a sibling index or other subindex or simply provide 0 if they dont care (additive blended things won't):

```
        public void Add(TKey key, TKey subKey, TValue value)
        {
            _backingList.Add(new KeyValuePair<TKey, TValue>(key * 1000 + subKey, value));
            _needsSorting = true;
        }
```

I haven't followed this through but could try it out if you agree to see if it works.

crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
```diff
@@ -323,6 +330,12 @@ public void CrestUpdate(CommandBuffer buf)
             // Seems like shader errors cause this to unbind if I don't set it every frame. Could be an editor only issue.
             _matGenerateWaves.SetTexture(sp_WaveBuffer, _waveBuffers);
 
+            if (_meshForDrawingWaves != null && _matGenerateWavesGeometry != null)
```
huwb, line 333: Does this need to be done every frame? If so its not the end of the world, just checking. (Would wrapping in `#if UNITY_EDITOR` potentially work?)

crest/Assets/Crest/Crest/Shaders/Resources/FFT/FFTCompute.compute
```diff
@@ -167,6 +168,7 @@ void ComputeFFT(uint3 id : SV_DispatchThreadID)
 #else
 	const float sign = ((id.x + id.y) % 2) == 1 ? -1.0 : 1.0;
 	const float3 res = float3(sign * resultX.x, sign * resultH.x, sign * resultZ.x);
-	_Output[id] = float3(sign * resultX.x, sign * resultH.x, sign * resultZ.x);
+	// Write zero to W to clear garbage.
```
huwb, line 171: eek this is quite a bug. nice one

docs/about/history.rst
```diff
@@ -23,6 +23,7 @@ Breaking
    -  Added new *Input Mode* setting to FFT component and to all *RegisterInput* components to support more diverse authoring workflows.
       This option may need to be configured on existing components.
       A migration tool to fix up existing inputs is available via the *Run 2022 Migration on Ocean Inputs* button in the *OceanRenderer* inspector.
+   -  Wave variance for ShapeGerstner has been removed which will reduce overall foam strength.
```
huwb, line 26: might want to justify why this was taken away

docs/about/history.rst
```diff
@@ -40,6 +41,11 @@ Changed
 
    -  Warn users edits in prefab mode will not be reflected in scene view until prefab is saved.
 
+Changed
+^^^^^^^
+.. bullet_list::
+
+   -  Added ability to blend waves (effectively an override) instead of only having additive waves. Disable "Additive" on the *ShapeFFT* or *ShapeGerstner* when using with a *MeshRenderer* or *Spline* to use. Useful for preventing rivers from receiving ocean waves.
```
huwb, line 48: wording out of date now that you have the enum, and i'd probably say *Added ability to alpha blend waves* as additive is a type of blend. might just be a graphics-tainted mind reading it that way though :)

huwb (changes requested):
Generally looks great. Req changes due to queue questions which need an iteration I think

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
daleeidd: Good idea. I don't think it makes sense to change the sorted list. This is specific to inputs at the moment so could implement logic in base class? Queue would be a good feature across the board. A separate PR is warranted. Happy to do it tomorrow if you're busy today.

daleeidd: I think a static method on register input base class could work.

crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
daleeidd: Only on setup I think. Done.

crest/Assets/Crest/Crest/Shaders/Resources/FFT/FFTCompute.compute
daleeidd: Indeed. Thanks.

docs/about/history.rst
daleeidd: Indeed. Done.

docs/about/history.rst
daleeidd: Good point. Done.

daleeidd:
I think that's everything. Thanks for adding #1052.

huwb (approved):
Nice one, looking forward to using this in Island!

daleeidd:
Thanks

huwb:
Seems to work well so far over in island 👍

PR #1033, opened 2022-05-29
Refactor HelpURL to be dynamic

daleeidd:
In 2021 Unity removed the sealed modifier to the HelpURL class which allows us to extend it. This is what Unity does for UPM documentation links.

Downstream I can added SRP support, but I might not need to if removing RP branching from documentation.

I also fixed some links and added missing ones.


huwb (changes requested):
Nice work!

PR #1034, opened 2022-05-29
Remove RenderWireFrame

daleeidd:
No longer needed as wireframe can be rendered through the scene view which Crest supports.


huwb (approved):
Thanks for this

PR #1035, opened 2022-05-30
Apply proper version, remove ugly old version hacks

huwb:
The spectrum data was a struggle because `_version` did not previous exist, so we could not rely on it for migration.

We therefore had to put in some heavy handed stuff for the migration.

This finally bumps the version to 1 and removes the "auto upgrade" path as mentioned in the TODO. I also removed some version 0 runtime code as version 0 is very old now.

This leaves the normal v0-v1 data migration, i was tempted to remove that too but thought since its not affecting the runtime and its using the correct, up to date migration pattern I might leave it in, for illustration purposes more than anything else. Happy to remove it though.


daleeidd (approved):
Looks good.

daleeidd:
> This leaves the normal v0-v1 data migration, i was tempted to remove that too but thought since its not affecting the runtime and its using the correct, up to date migration pattern I might leave it in, for illustration purposes more than anything else. Happy to remove it though.

Just confirming this is calling _Upgrade_ from _OnInspectorGUI_ that you want to leave in?

huwb:
Oh I'm not sure now that you mention it. It seems like we don't have an established path for SO's. There's a Data Upgrade page but it only covers using ISerializationCallbackReceiver with Monobehaviour's, I think..

> It might be able to be used with ScriptableObject (DE: Had no luck with OceanSpectrum).

daleeidd:
I investigated again and it appears to work for SOs just fine. Maybe it wasn't working in 2020.

daleeidd (approved):
Looks good.

daleeidd:
> Annoyingly it does not set the dirty flag it seems. It threw an exception if i called `SetDirty()`. That seems like a unity bug. Guess it can work like so.

By design unfortunately :\ but luckily migration code is pretty lightweight.

PR #1036, opened 2022-05-30
Factor out feature validation so it can be used from SWS

huwb:
I accidentally had flow disabled when messing around with shoreline waves, so I added validation to the SWS for ocean Flow feature enabled on OR and on material.

Rather than duplicate that stuff, I factored out the validation code to a central location, which is in this PR.

This means I can call `OceanRenderer.ValidateFeatureEnabled` from the SWS validated editor.

I only did this for Flow as its the only one SWS. However i could roll out the minor change to `LodDataMgrFlow` to the other lod datas if you think it makes sense.


daleeidd (approved):
Looks good. Nice one.

daleeidd:
> I only did this for Flow as its the only one SWS. However i could roll out the minor change to `LodDataMgrFlow` to the other lod datas if you think it makes sense.

Up to you. I already had a refactor planned with #1028 (see _Separate Ocean Material Validation_ heading) which covers this but if you prefer your approach then can do that instead.

huwb:
Ah no yours is better. But I'll merge this if that's ok and key you take over later.

PR #1037, opened 2022-05-31
Add enum filter attribute

daleeidd:
Allows for subclasses to filter enum options.

```c#
class BaseInput
{
  [Filtered]
  Mode _mode;
}

[FilterEnum("_mode", FilteredAttribute.Mode.Include, (int)Mode.Primitive, (int)Mode.Geometry)]
class Input {}
```

There is a demo commit with implementation on the inputs for testing if you wish.

It will have to be implemented in painted branch to avoid a merge conflict.


huwb:
Smart! Yeah this seems good

huwb (approved):

daleeidd:
I've renamed it to Filtered and FilterEnum which is more in line with other attributes. I've also added validation. If that's okay I will merge (accidentally squashed it already :\).

huwb:
Go for it, thanks

PR #1038, opened 2022-06-02
Use Shader Graph for example materials

daleeidd:
- Add SG package
- Add Lit SG
- Add Masked SG
- Use Lit SG in examples
- Use Masked SG in examples


daleeidd:
> Only potential query is Masked.hlsl is a generic name, could it be more descriptive perhaps. A comment at the top of the file might also help. I'm assuming it's used for general shaders, not specifically for water?

Not specific to water. I'll add a comment. I can rename it to CrestExamplesNodeMasked (and the function too) or something? 

> Anyway approving as I'm heading out and don't want to block you, but if you prefer I can rereview later this evening.

No rush as I won't be merging until tomorrow.

huwb:
Yeah that name sounds good

daleeidd:
All done. Thanks. I also removed the Masked surface shader since the SG replaces it.

PR #1041, opened 2022-06-03
Shoreline wave simulation

huwb:
Draft PR - needs simulation tweaks to make it work in Island scene. Works ok/pretty well in the main.unity on this branch.

Creating this PR to track as it is hopefully converging now.

All data will be reverted (probably).

The spectrum changes will be reverted (probably).

Some old and dubious validation that checked that an FFT or Gerstner is present is removed, as some use cases may not want waves, and this sim creates waves - it should in theory be able to produce rivers too in the future and if so FFT may be unnecessary for a river scene. LMK if this makes sense.

Must be placed at sea level and not use Sea Level Offset. I experimented with supporting it but it's tricky. The origin of the sim is placed relative to global sea level which doesnt work when sea level can vary. Not sure yet how to proceed but probably will add option to allow sim to be placed at any height, so that user could place it where they need it (perhaps at lowest point of river). Leaving for future work.

No VFX - think that should probably be a separate PR in SRP

TODO

* Fix water behaviour underneath Island that is affecting the shoreline waves it seems
* Revert data changes, and spectrum too?


daleeidd:
Looks really good so far. What was the reason it cannot run in edit mode?

huwb:
Ah i'm not really sure, i will take a look at that

PR #1043, opened 2022-06-05
Use full-screen triangle for shadow simulation

daleeidd:
Unity's Blit has a lot of limitations like not supporting the RP tag. I have moved to using our own helper method which uses FST. This change will allow us to combine all RP subshaders into one file using the RP tag.


huwb (approved):
Looks good to me

daleeidd:
Thanks!

PR #1044, opened 2022-06-06
Move "default clipping state" to settings

daleeidd:
Moved _Default Clipping State_ from _Ocean Renderer_ into _Sim Settings Clip Surface_ asset. I don't think a migration is worth the effort as we'd have to create a settings file if they don't have one which probably won't play nice in the migration pattern we favour.

### DIFF
--- a/docs/about/history.rst
+++ b/docs/about/history.rst
@@ -18,6 +18,8 @@ Breaking
 .. bullet_list::
 
    -  Set minimum Unity version to 2021.3.3.
+   -  Moved *Default Clipping State* from *Ocean Renderer* into *Sim Settings Clip Surface* asset.
+      There is no migration available for this setting so if *Everything Clipped* is needed then this will need to be configured in the *Sim Settings Clip Surface* asset.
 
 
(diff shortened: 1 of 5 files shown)


docs/about/history.rst
```diff
@@ -18,6 +18,8 @@ Breaking
 .. bullet_list::
 
    -  Set minimum Unity version to 2021.3.3.
+   -  Moved *Default Clipping State* from *Ocean Renderer* into *Sim Settings Clip Surface* asset.
+      There is no migration available.
```
huwb, line 22: This could probably instruct the user what to do.. maybe something along the lines of this?

"There is no migration available for this setting so if the *Everything Clipped* setting is needed this will need to be configured in the settings data object."

huwb (approved):
Minor suggestion but approving

docs/about/history.rst
daleeidd: Good idea. Done. Thanks.

daleeidd:
Thanks!

PR #1045, opened 2022-06-07
Experiment with new approach

huwb:
I figured a lot of the complication was coming from fighting with serialisation, and decided to untangle things by separating out serialised/user facing state with internal actual state. I replaced 'Unset' with 'Autodetect' which is a valid default for all components. Then internally, if the user setting is Autodetect, the component detects which state to go into.

This creates a valid default for the serialised data - a deserialised component with default data should work happily and we should not have to worry about migration for old data. I feel this is a compelling idea and with just a few changes i could remove the code that fights with serialisation.

However it throws up a whole new set of problems with keeping the user facing state and internal state synchronised. I did some changes and it might be working in many cases.. but i dont think all.. and there's too many cases for me to reason about all at once, so im not sure its worth pursuing further. Or maybe i just need to find the right pattern.

Another problem is its not clear what state the component as has picked, although if we could display the dropdown option with the current selected mode that could sort it, like `Autodetect (Spline)`.

I think we should stick with what we have but posting here for posterity.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Helpers/Editor/EditorHelpers.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Helpers/Editor/EditorHelpers.cs
@@ -109,32 +109,32 @@ namespace Crest.EditorHelpers
                 EditorUtility.SetDirty(input);
             };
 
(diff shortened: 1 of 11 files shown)


daleeidd:
Not a bad idea. A little hard to follow. There are a few schemes we could use but I think they are all more complicated.

PR #1049, opened 2022-06-09
Fix prefab mode breaking things

daleeidd:
Crest doesn't support prefab mode but some components still run which causes problems. This makes disabling prefab mode more robust and expands the roster of components to all problematic components.

The prefab mode experience is that the OR from the scene will continue to run and be visible in prefab mode, but prefab mode changes will not be reflected in the scene view until it is saved. This is due to the OR in the scene view not knowing about the changes made in prefab mode as they are separate instances. This is the same for the other problematic components: UR, ShapeFFT and ShapeGerstner.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
+++ b/crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
@@ -9,11 +9,6 @@ using Crest.Internal;
 #if UNITY_EDITOR
 using UnityEngine.Rendering;
 using UnityEditor;
-#if UNITY_2021_2_OR_NEWER
-using UnityEditor.SceneManagement;
-#else
-using UnityEditor.Experimental.SceneManagement;
-#endif
 #endif
 
 #if !UNITY_2021_3_OR_NEWER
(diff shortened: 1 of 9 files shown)


huwb:
Quick note before reviewing - i've realised this limitation seems to be missing from our docs - worth adding to Known Issues and FAQ?

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
```diff
@@ -1959,11 +1958,19 @@ void OnDisable()
 
         public override void OnInspectorGUI()
         {
+            var target = this.target as OceanRenderer;
+
+            if (target._isPrefabStageInstance)
+            {
+                EditorGUILayout.Space();
+                EditorGUILayout.HelpBox("Crest does not support prefab mode. Changes made in prefab mode will not be reflected in the scene view until saved.", MessageType.Warning);
```
huwb, line 1966: could you maybe expand "until saved" slightly. does this mean the user has to `File/Save` and it will update the water? perhaps saying directly "save this prefab to update the water" or something?

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
```diff
@@ -527,21 +558,28 @@ void BakeHelpBox(ShapeFFT target)
...
         public override void OnInspectorGUI()
         {
-            base.OnInspectorGUI();
+            var target = this.target as ShapeFFT;
 
-            var fft = target as ShapeFFT;
+            if (target._isPrefabStageInstance)
+            {
+                EditorGUILayout.Space();
+                EditorGUILayout.HelpBox("Crest does not support prefab mode. Changes made in prefab mode will not be reflected in the scene view until saved.", MessageType.Warning);
```
huwb, line 566: similar comment - directive "save the prefab to see changes" woudl get my vote

crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
```diff
@@ -106,6 +106,10 @@ protected static OceanWaveSpectrum DefaultSpectrum
             }
         }
 
+#if UNITY_EDITOR
+        internal bool _isPrefabStageInstance = false;
```
huwb, line 110: i assume it was not possible to have centralised state for this, like share the one on the OR or a static flag or something? sorry if you already mentioned it, i'm in "review mode" of github now and i dont want to risk clicking away :)

huwb (approved):
my comments were aesthetic so ill approve but feel free to rerequest review if its more than that.

crest/Assets/Crest/Crest/Scripts/OceanRenderer.cs
daleeidd: Correct. Sounds good. Done. I also refactored and placed constant string in common area.

crest/Assets/Crest/Crest/Scripts/Shapes/FFT/ShapeFFT.cs
daleeidd: Done.

crest/Assets/Crest/Crest/Scripts/Shapes/ShapeGerstner.cs
daleeidd: Nah. Unfortunately each one needs to track it themselves.

daleeidd:
> Quick note before reviewing - i've realised this limitation around prefab mode seems to be missing from our docs - worth adding to Known Issues and FAQ?

Indeed. Added to both. Also to _Getting Start_ page.

daleeidd:
Thanks!

PR #1051, opened 2022-06-10
Tweaks to painting input

huwb:
Few minor tweaks after authoring Island waves.

Most significant is the wave direction change. When painting waves, take the direction the user is painting for the waves, don't incorporate the global wind heading offset. Adding the offset makes the behaviour very confusing and I don't envision a use case for this in practice.


daleeidd (approved):
Good idea. Looks good.

daleeidd:
> Most significant is the wave direction change. When painting waves, take the direction the user is painting for the waves, don't incorporate the global wind heading offset. Adding the offset makes the behaviour very confusing and I don't envision a use case for this in practice.

I can see a use case for the wind direction completely overriding the painted one. For when a user wants to paint a patch of waves all in one direction which will be awkward. But that can come later if people ask for it.

huwb:
Thanks

> I can see a use case for the wind direction completely overriding the painted one

Do you mean besides just setting the FFT to Global? If so then there should be blending options. I think i touched on this in the call. I think your alpha blending of anim waves hsould open up opportunities to author different situations.

PR #1052, opened 2022-06-12
Helpers for register/deregister with input for subqueue index

huwb:
After identifying potential issues with sorting, we decided picking sibling index as a sub-queue-index might work well.

So sorting of inputs effectively happens on two keys - primarily on queue index, and then secondarily on subqueue index, for which the sibling index can be used. So relative order of siblings will dictate order they apply, similar to Unity's UI system.

This took a bit of untangling to bring over from #1032. The queue value is not initialised for FFT/Gerstner, leaving that for #1032.

This simplifies/cleans up the code as a bonus.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Interaction/SphereWaterInteraction.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Interaction/SphereWaterInteraction.cs
@@ -238,14 +238,12 @@ namespace Crest
 
         void OnEnable()
         {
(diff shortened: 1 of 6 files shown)


daleeidd (approved):
Looks great. Nice one. Just needs release notes.

huwb:
Oh yeah i guess we could call out the sub-sort support, good shout. Done now. Will merge to keep things moving but feel free to modify or request changes after.

PR #1053, opened 2022-06-13
Add hide option to Predicated and clean up UI

daleeidd:
Essentially cleans up UI for several components by hiding irrelevant sections per mode. Also includes a few fixes.

- Add hide option to Predicated
- Apply hide option to several components
- Add some extra headings
- Fix SFD input mode filter

I added the optional hide option as the last parameter which saved fixing up a bunch of stuff. Also _Predicated_ now keeps the disabled state so we can have multiple _Predicated_ attributes playing nicely together. Cannot do anything too fancy though.

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/Helpers/Attributes/MultiPropertyAttribute.cs
+++ b/crest/Assets/Crest/Crest/Scripts/Helpers/Attributes/MultiPropertyAttribute.cs
@@ -263,4 +263,24 @@ namespace Crest
             _property = property;
         }
     }
+
+    /// <summary>
+    /// Drop-in replacement for Header. Use when hiding property with Predicated.
+    /// </summary>
+    public class HeadingAttribute : DecoratorAttribute
+    {
+        readonly string _heading;
+
+        public HeadingAttribute(string heading)
+        {
+            _heading = heading;
+        }
+
+        internal override void Decorate(Rect position, SerializedProperty property, GUIContent label, DecoratedDrawer drawer)
+        {
+            // Register margin with IMGUI so subsequent spacing is correct.
+            EditorGUILayout.GetControlRect(false, EditorGUIUtility.singleLineHeight * 0.5f);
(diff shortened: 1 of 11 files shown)


huwb:
This is great! Massive difference.

I clicked through the scenes and everything seems fine.

Only one note regarding FFT - the Collision Data Baking section + buttons are only relevant to Global. Potentially theres a deeper issue here that it needs to robustly turn off/warn/etc if user is using wrong input mode. I think latter is def future work, but hiding this stuff would be nice to at least hint to the user that the two dont work together. This would clean up all the Bake buttons and solve the jumble of paint+bake buttons when Painted is selected.

crest/Assets/Crest/Crest/Scripts/Helpers/Attributes/MultiPropertyAttribute.cs
```diff
@@ -263,4 +263,21 @@ public FilterEnumAttribute(string property, FilteredAttribute.Mode mode, params
             _property = property;
         }
     }
+
+    public class HeadingAttribute : DecoratorAttribute
```
huwb, line 267: could you add bit of documentation for this class. is Heading to enable it to be hideable, or to fix spacing, or both? wonder if the name could be descriptive of what it does, but thats not a critical point, Heading works.

huwb (approved):
Minor documentation request, happy to leave it with you now. The FFT feedback would be nice to hit (see other comment), but its an extension and this pr could merge without, so approving.

crest/Assets/Crest/Crest/Scripts/Helpers/Attributes/MultiPropertyAttribute.cs
daleeidd: Done. Just for hiding. Otherwise it is drop-in replacement for Header.

daleeidd:
I missed that one. Good idea. I have done ShapeFFT suggestion. I moved _Culling_ section further up so it is out of the way. LMKWYT.

huwb:
Works well!

daleeidd:
Thanks

PR #1054, opened 2022-06-14
Hide filtered enum unset value

daleeidd:
Hides the "unset" value when not needed for filtered enums.


huwb (approved):
Yeah happy to go with this if it doesnt create any weird situations with the authoring. I'll mess around with it more as we approach release, happy for this to merge.

daleeidd:
No weird situations from my testing. Thanks!

PR #1055, opened 2022-06-14
Move OR debug options into foldable

daleeidd:
Moves OR debug options into foldable and changes them all to be unchecked by default:


daleeidd:
Screenshot is old but I renamed "No Follow Viewport" to "Disable Follow Viewport".

huwb (approved):
Nice looks good

daleeidd:
Thanks

Issue #1056, opened 2022-07-29
Entering and exiting play mode makes ocean tiles pickable again


huwb:
**Describe the bug**

If you disable picking of the ocean gameobject to stop the ocean being pickable, this propagates the setting to the tiles and it works. However entering/exiting play mode or any other rebuild of the ocean hierarchy loses the unpickability off the tiles.

Perhaps when the ocean hierarchy is constructed, the non-pickability shoudl be detected on the parent gameobject and propagated downwards.

**Versions**
Repro'd with BIRP, 2021.3.3f1

**Platform**
Editor / win64

huwb:
Not a good start - simply querying pickability of the ocean using `SceneVisibilityManager.instance.IsPickingDisabled(ocean.gameObject)` seems to clear the pickability flag on the ocean when exiting playmode. 

It's as if the pickability state is left hanging after exiting play mode, and then querying it refreshes it to default (pickable).

huwb:
Added a PR which may sort it. Will let Dale chime in when he's back

daleeidd:
For anyone dropping by here, locking the water layer is another option: https://blog.theknightsofunity.com/hide-lock-layer-scene-window/

The only downside is if one has liberally used the Water layer elsewhere. Generally we only add the Water layer to the ocean tiles.

PR #1058, opened 2022-07-31
Make ocean tiles non-pickable by default

huwb:
Issue #1056

Makes ocean tiles non-pickable by default, adds debug option for it.

This will stomp the users preference for pickability. However pickability is almost never desired I guess. And there's a debug option to disable it if need be. LMK what you think.

I initially just intended to respect the pickability state of the ocean gameobject and propagate it to the tiles but it was a non-starter it seems - simply querying pickability of the ocean using `SceneVisibilityManager.instance.IsPickingDisabled(ocean.gameObject)` seems to clear the pickability flag on the ocean when exiting playmode. It's as if the pickability state is left hanging after exiting play mode, and then querying it refreshes it to default (pickable).

### DIFF
--- a/crest/Assets/Crest/Crest/Scripts/OceanBuilder.cs
+++ b/crest/Assets/Crest/Crest/Scripts/OceanBuilder.cs
@@ -6,6 +6,7 @@
 
 using System.Collections;
 using System.Collections.Generic;
+using UnityEditor;
 using UnityEngine;
 
 namespace Crest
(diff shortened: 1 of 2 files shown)


NewColours:
A side effect that might not be obvious: When the sea tiles are pickable, it breaks snapping objects together in the editor. It took a long time for me to figure out why my floor tiles up high would not snap to each other, but this turned out to be the cause (as from a sufficient distance moving the target passed over dozens of sea tile snap points).

Please merge this 🤞

daleeidd:
@NewColours you could lock the Water layer instead? That appears to hold up even on editor restarts. Let me know how you go.

Reference: https://blog.theknightsofunity.com/hide-lock-layer-scene-window/

daleeidd:
Closing in favour of #1066

PR #1059, opened 2022-09-13
burst queries and calculating velocities

ecuzzillo:
And, move many things to NativeCollections for that purpose. 

In our game, in areas with several water objects, this change made Query() calls disappear from the profiler, where before they had taken 2-5ms/frame. 

This only bursts the things that showed up for our particular usage, but the changes here likely enable bursting many more things for other use cases.


daleeidd:
Thank you for the PR. In the end decided to push this out to mainstream so had to change the base to the master branch (required copying work over). Closing this in favour of #1143

PR #1060, opened 2022-09-22
Fix painting alpha blending clobbering waves

daleeidd:
Attempted to solve the alpha blending clobbering all waves within the quad even with zero data.

Seems to work okay. If someone wanted to paint flat water they just have to paint very slowly.


crest/Assets/Crest/Crest/Shaders/OceanInputs/Resources/AnimWavesGenerate.shader
```diff
@@ -121,6 +121,8 @@ Shader "Hidden/Crest/Inputs/Animated Waves/Generate Waves"
 					{
 						float2 axis = _PaintedData.Sample(LODData_linear_clamp_sampler, paintUV).xy;
 						float axisLen2 = dot(axis, axis);
+						wt *= saturate(axisLen2);
```
huwb, line 124: this is the squared length rather than the length

if alpha blend is enabled, this weight is effectively applied twice. once below here

							disp.xz = disp.x * axis + disp.z * float2(-axis.y, axis.x);
							disp.y *= sqrt(axisLen2);

and then once while alpha blending.

this can be changed by normalizing `axis` before use so that it has length 1, as in the code snippet i sent. or maybe applying it twice is fine. since its using the squared length, its probably using weight to the power of 3 currently when alpha blend is enabled, which might be working fine and intentional.

huwb:
i had a quick play with it and it seems to work well for me, for Additive and Alpha Blend.

i tried Premultiplied and it made a quite intense result and i didnt really understand. The tooltip for the blend option probably needs to be updated to explain to a laymen the different options and when to use each.

PR #1090, opened 2022-10-26
Add wave sampling shader for renderer input

daleeidd:
Adds support for mesh renderers to ShapeWaves. And adds a shader which samples the spectrum based on the painted data approach where RG channels are the XZ wave direction and together are the magnitude (0-1 maps to −1-1). Object rotation and _Wave Direction Heading Angle_ are also applied to the direction.

There is a demo scene under the Demo folder for testing. It samples waves using an arrow as a mask.


daleeidd:
@holdingjason This should be working correctly now. 0-1 maps to -1-1 and 0.5 being 0 (ie no waves). Getting exactly 0.5 in a texture is quite difficult so it gives a small margin.

PR #1101, opened 2023-01-13
Do not execute in background

daleeidd:
In certain cases memory would slowly increase indefinitely. It would be cleared when editor became active again but would like to avoid leaks altogether.

This could be a Unity bug but more effective to patch than report. It also has the benefit of saving energy.


mstevenson:
Thanks for this, I noticed that the Editor would crash after sitting idle in the background for a few hours with an ocean visible, and I suspect that this was the cause.

PR #1143, opened 2024-02-02
Burstify queries

magenta404:
Takes work from and replaces #1059

Based off master instead of experimental as few things changed to adopt this mainstream. Requires an opt-in via scripting symbol "CREST_BURST_QUERY" to avoid a breaking change.


crest/Assets/Crest/Crest/Scripts/Collision/SamplingHelpers.cs
```diff
@@ -10,12 +12,32 @@ namespace Crest
...
+#if CREST_BURST_QUERY
+        Vector3 _queryPos;
+        Vector3 _queryResult;
+        Vector3 _queryResultNormal;
+        Vector3 _queryResultVel;
+
+        // These are only ever used on the main thread, and they're filled in when they're used, so we only need one
+        // global copy of them. if we had one for every sample height helper, we would spend forever creating and
+        // destroying nativearrays. consider revisiting this when 22.2 is the minspec, because then nativearray becomes
+        // fully unmanaged. Also, obviously revisit if their usage is ever jobified.
```
daleeidd, line 29: Not sure about this as when would we ever be creating/destroying this helper class frequently? Happy to go with it though.

crest/Assets/Crest/Crest/Scripts/Collision/CollProvider.cs
```diff
@@ -33,7 +38,12 @@ public interface ICollProvider
         /// <param name="o_resultDisps">Displacement vectors for water surface points that will displace to the XZ coordinates of the query points. Water heights are given by sea level plus the y component of the displacement.</param>
         /// <param name="o_resultNorms">Water normals at the query positions. Pass null if this information is not required.</param>
         /// <param name="o_resultVels">Water surface velocities at the query positions. Pass null if this information is not required.</param>
+#if CREST_BURST_QUERY
+        /// <param name="useNormals"></param>
+        int Query(int i_ownerHash, float i_minSpatialLength, ref NativeArray<Vector3> i_queryPoints, ref NativeArray<Vector3> o_resultDisps, ref NativeArray<Vector3> o_resultNorms, ref NativeArray<Vector3> o_resultVels, bool useNormals);
```
daleeidd, line 43: "useNormals" should be dropped or also added to the other method to keep the API consistent. I would vote with dropping unless it really saves on performance.

daleeidd: That being said, checking null or Length to determine intent of the user throughout the call stack makes things very confusing down the line.

daleeidd: Went with dropping it and improved checks throughout. Also added a helper to make passing "null" easier.

crest/Assets/Crest/Crest/Scripts/Collision/SamplingHelpers.cs
daleeidd: The solution in this comment does save users from disposing this class. May leave comment as-is.

skaughtx0r:
Do these changes make it so the Query for water heights can be done in a Job?

daleeidd:
Only Burst compiled. I have updated the branch to 4.20.1 if you would like to see if there is an improvement and give feedback.

